PR #14, opened 2018-09-11
[WIP] Make detectors and printers registrable

disconnect3d:
Thx to this PR Slither users will be able to plug-in their own detectors and printers.


disconnect3d:
@montyly From the things we discussed recently, this still lacks:
- `choose_detectors` and `choose_registers` implementations: so it launches **every detectors/printers you provide to it**
- handling the envvar to handle plugin (I believe this should be in `choose_*` funcs)

That being said I am putting this as a WIP.

slither/__main__.py
```diff
@@ -198,6 +200,8 @@ def parse_args(detector_classes, printer_classes):
                             dest="detectors_to_run",
                             const=detector_cls.ARGUMENT)
 
+    # Second loop so that the --exclude are shown after all the detectors
```
disconnect3d: Ha, good point. Didn't think about it :P.

PR #16, opened 2018-09-11
Add automatic plugins loading

disconnect3d:
From `SLITHER_PLUGINS_MODULES` envvar or `slither_plugin` module.

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -8,9 +8,8 @@ import os
 import sys
 import traceback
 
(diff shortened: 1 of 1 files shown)


disconnect3d:
Closing the PR as we will try with plugins arch: https://docs.pylonsproject.org/projects/pylons-webframework/en/latest/advanced_pylons/entry_points_and_plugins.html

This may be reopened if we decide to use it.

PR #20, opened 2018-09-25
SlithIR

montyly:
WIP of the Slither intermediate representation


slither/analyses/write/are_variables_written.py
```diff
@@ -0,0 +1,56 @@
+"""
+    Detect if all the given variables are written in all the paths of the function
+"""
+from slither.core.cfg.node import NodeType
+from slither.core.declarations.solidity_variables import SolidityFunction
+from slither.slithir.operations.index import Index
+from slither.slithir.operations.member import Member
+from slither.slithir.operations.lvalue import OperationWithLValue
```
disconnect3d: We can have in `operations/__init__.py`:
```python
from .index import Index
from .member import Member
from .lvalue import OperationWithLValue
```

which would let us do:
```python
from slither.slithir.operations import Index, Member, OperationWithLValue
```

montyly: I started moving towards this structure in https://github.com/trailofbits/slither/pull/20/commits/c4b55e128cde4dbea1eba92b895b6284e9465acd

PR #28, opened 2018-10-17
Printer human readable

montyly:
WIP of a human-readable summary

### DIFF
--- a/slither/core/declarations/contract.py
+++ b/slither/core/declarations/contract.py
@@ -245,7 +245,7 @@ class Contract(ChildSlither, SourceMapping):
         Returns:
             Structure
         """
-        return next((st for st in self.structure if st.name == structure_name), None)
+        return next((st for st in self.structures if st.name == structure_name), None)
 
     def get_structure_from_canonical_name(self, structure_name):
         """
@@ -255,7 +255,7 @@ class Contract(ChildSlither, SourceMapping):
         Returns:
             Structure
         """
-        return next((st for st in self.structure if st.canonical_name == structure_name), None)
+        return next((st for st in self.structures if st.canonical_name == structure_name), None)
 
     def get_event_from_name(self, event_name):
         """
@@ -275,7 +275,7 @@ class Contract(ChildSlither, SourceMapping):
         Returns:
             Enum
         """
-        return next((e for e in self.enum if e.name == event_name), None)
+        return next((e for e in self.enums if e.name == enum_name), None)
 
     def get_enum_from_canonical_name(self, enum_name):
         """
@@ -285,7 +285,7 @@ class Contract(ChildSlither, SourceMapping):
         Returns:
             Enum
         """
-        return next((e for e in self.enum if e.canonical_name == event_name), None)
+        return next((e for e in self.enums if e.canonical_name == enum_name), None)
 
     def get_summary(self):
         """ Return the function summary
--- a/slither/core/declarations/contract.py
+++ b/slither/core/declarations/contract.py
@@ -287,6 +287,18 @@ class Contract(ChildSlither, SourceMapping):
         """
         return next((e for e in self.enums if e.canonical_name == enum_name), None)
 
+    def is_erc20(self):
+        """
+            Check if the contract is a erc20 token
+            Note: it does not check for correct return values
+        Returns:
+            bool
+        """
+        full_names = [f.full_name for f in self.functions]
+        return 'transfer(address,uint256)' in full_names and\
+               'transferFrom(address,address,uint256)' in full_names and\
+               'approve(address,uint256)' in full_names
+
     def get_summary(self):
         """ Return the function summary
 
--- a/slither/printers/summary/printer_human_summary.py
+++ b/slither/printers/summary/printer_human_summary.py
@@ -1,0 +1,107 @@
+"""
+    Module printing summary of the contract
+"""
+import logging
+
+from slither.printers.abstractPrinter import AbstractPrinter
+from slither.utils.colors import blue, green, magenta, red, yellow
+
+from slither.utils.code_complexity import compute_cyclomatic_complexity
+
+from slither.detectors.detectors import Detectors
+
+class PrinterHumanSummary(AbstractPrinter):
+
+    ARGUMENT = 'print-human-summary'
+    HELP = 'Print an human readable summary of the contracts'
+
+    @staticmethod
+    def get_summary_erc20(contract):
+        txt = ''
+        functions_name = [f.name for f in contract.functions]
+        state_variables = [v.name for v in contract.state_variables]
+
+        if 'paused' in functions_name or 'pause' in functions_name:
+            txt += "\t\t Can be paused? : {}\n".format(yellow('Yes'))
+        else:
+            txt += "\t\t Can be paused? : {}\n".format(green('No'))
+
+        if 'mint'  in functions_name:
+            if not 'mintingFinished' in state_variables:
+                txt += "\t\t Minting restriction? : {}\n".format(red('None'))
+            else:
+                txt += "\t\t Minting restriction? : {}\n".format(yellow('Yes'))
+        else:
+            txt += "\t\t Minting restriction? : {}\n".format(green('No Minting'))
+
+        if 'increaseApproval' in functions_name:
+            txt += "\t\t ERC20 race condition mitigation: {}\n".format(green('Yes'))
+        else:
+            txt += "\t\t ERC20 race condition mitigation: {}\n".format(red('No'))
+
+        return txt
+
+    def get_detectors_result(self):
+        detectors = Detectors()
+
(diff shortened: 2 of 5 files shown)


slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+
+from slither.detectors.detectors import Detectors
+from slither.printers.abstractPrinter import AbstractPrinter
+from slither.utils.code_complexity import compute_cyclomatic_complexity
+from slither.utils.colors import green, red, yellow
+
+
+class PrinterHumanSummary(AbstractPrinter):
+    ARGUMENT = 'print-human-summary'
+    HELP = 'Print an human readable summary of the contracts'
```
disconnect3d:
```suggestion
    HELP = 'Print a human readable summary of the contracts'
```

slither/utils/code_complexity.py
```diff
@@ -0,0 +1,75 @@
+# Funciton computing the code complexity
+
+def compute_number_edges(function):
+    """
+        Compute the number of edges of the CFG
```
disconnect3d:
```suggestion
    Compute the number of edges of the CFG
```

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+            if compute_cyclomatic_complexity(f) > 7:
+                is_complex = True
+
+        result = red('Yes') if is_complex else green('No')
+
+        return "\tComplex code? {}\n".format(result)
+
+    def output(self, _filename):
+        """
+            _filename is not used
```
disconnect3d:
```suggestion
        _filename is not used
```

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+            Heuristic, the code is complex if:
+                - One function has a cyclomatic complexity > 7
+        Args:
+            contract
+        """
+        is_complex = False
+
+        for f in contract.functions:
+            if compute_cyclomatic_complexity(f) > 7:
+                is_complex = True
```
disconnect3d: We could skip checking other functions if one is complex

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+        txt = "Number of low issues: {}\n".format(green(len(issues_low)))
+        txt += "Number of medium issues: {}\n".format(yellow(len(issues_medium)))
+        txt += "Number of high issues: {}\n".format(red(len(issues_high)))
+
+        return txt
+
+    @staticmethod
+    def is_complex_code(contract):
+        """
+            Check if the code is complex
```
disconnect3d: One less indent level? Same of other funcs above

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+    ARGUMENT = 'print-human-summary'
+    HELP = 'Print an human readable summary of the contracts'
+
+    @staticmethod
+    def get_summary_erc20(contract):
+        txt = ''
+        functions_name = [f.name for f in contract.functions]
+        state_variables = [v.name for v in contract.state_variables]
+
+        if 'paused' in functions_name or 'pause' in functions_name:
```
disconnect3d:
```suggestion
        if 'pause' in functions_name:
```

This check is enough as `pause` is in `paused` :)

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+
+class PrinterHumanSummary(AbstractPrinter):
+    ARGUMENT = 'print-human-summary'
+    HELP = 'Print an human readable summary of the contracts'
+
+    @staticmethod
+    def get_summary_erc20(contract):
+        txt = ''
+        functions_name = [f.name for f in contract.functions]
+        state_variables = [v.name for v in contract.state_variables]
```
disconnect3d:
```suggestion
        state_variables = [v.name.lower() for v in contract.state_variables]
```

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class PrinterHumanSummary(AbstractPrinter):
+    ARGUMENT = 'print-human-summary'
+    HELP = 'Print an human readable summary of the contracts'
+
+    @staticmethod
+    def get_summary_erc20(contract):
+        txt = ''
+        functions_name = [f.name for f in contract.functions]
```
disconnect3d:
```suggestion
        functions_name = [f.name.lower() for f in contract.functions]
```

disconnect3d: I am not 100% sure about it but we should consider either converting things to lowercase or matching some of the strings as case insensitive.

I assume that for example `paused` function is not a standard one (e.g. for erc20 token) and one can name it as `isPaused` or `getPaused` (lol).

slither/printers/summary/printer_human_summary.py
disconnect3d: See comment above (for `.lower()` in state variables).

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+                txt += "\t\t Minting restriction? : {}\n".format(yellow('Yes'))
+        else:
+            txt += "\t\t Minting restriction? : {}\n".format(green('No Minting'))
+
+        if 'increaseApproval' in functions_name:
+            txt += "\t\t ERC20 race condition mitigation: {}\n".format(green('Yes'))
+        else:
+            txt += "\t\t ERC20 race condition mitigation: {}\n".format(red('No'))
+
+        return txt
```
disconnect3d: We could also have like:
```python
def colored(condition, true='Yes', false='No'):
    return green(true) if condition else red(false)

paused = colored('pause' in functions_name)

if 'mint' in functions_name:
    mint = colored('mintingFinished' in state_variables, false='None')
else:
    mint = green('No Minting')

race_cond = colored('increaseApproval' in functions_name)

return (
    f'\t\t Can be paused? {paused}\n'
    f'\t\t Minting restriction? {mint}\n'
    f'\t\t ERC20 race condition mitigation: {race_cond}\n'
)
```

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+        if 'mint' in functions_name:
+            if not 'mintingFinished' in state_variables:
+                txt += "\t\t Minting restriction? : {}\n".format(red('None'))
+            else:
+                txt += "\t\t Minting restriction? : {}\n".format(yellow('Yes'))
+        else:
+            txt += "\t\t Minting restriction? : {}\n".format(green('No Minting'))
+
+        if 'increaseApproval' in functions_name:
+            txt += "\t\t ERC20 race condition mitigation: {}\n".format(green('Yes'))
```
disconnect3d: Consider being more explicit here as:
* the check itself doesn't necessarily mean that there is race condition mitigation - at least I assume that it depends on how the `increaseApproval` is implemented, right?
* someone needs to look into the code to understand what does it mean
* maybe we should change `Yes` and `No` here into: `has increaseApproval function`, `misses increaseApproval function`?

slither/printers/summary/printer_human_summary.py
disconnect3d: Also, maybe we should use some more general names that `green` and `red` for colors, so it is would be possible to configure colors in one place?

we could have like `ok(msg)` and `err(msg)` or `yes` and `no` or sth like that.

slither/core/declarations/contract.py
```diff
@@ -286,6 +286,18 @@ def get_enum_from_canonical_name(self, enum_name):
         """
         return next((e for e in self.enums if e.canonical_name == enum_name), None)
 
+    def is_erc20(self):
+        """
+            Check if the contract is a erc20 token
```
benstew:
```suggestion
            Check if the contract is an erc20 token
```

slither/printers/summary/printer_human_summary.py
```diff
@@ -0,0 +1,102 @@
...
+        txt += "Number of high issues: {}\n".format(red(len(issues_high)))
+
+        return txt
+
+    @staticmethod
+    def is_complex_code(contract):
+        """
+            Check if the code is complex
+            Heuristic, the code is complex if:
+                - One function has a cyclomatic complexity > 7
```
benstew: Friendly reminder that benchmark testing should still be completed to determine if this is an appropriate default threshold

slither/utils/code_complexity.py
```diff
@@ -0,0 +1,75 @@
+# Funciton computing the code complexity
```
benstew:
```suggestion
# Function computing the code complexity
```

PR #36, opened 2018-10-18
detectors: Add low level calls detector.

cryptomental:
Closes https://github.com/trailofbits/slither/issues/27


scripts/travis_test.sh
```diff
@@ -63,6 +63,10 @@ if [ $? -ne 3 ]; then
     exit 1
 fi
 
+slither tests/reentrancy.sol --disable-solc-warnings --detect-low-level-calls
```
montyly: Could you add a separate testcase? It's better for maintainability to have at least one specific testcase per detector

PR #42, opened 2018-10-19
Adds printer for inheritance and updates inheritance-graph printer

### LINKED ISSUE #23: Printer: Inheritance printing on stdout
Hey,

It would be cool to be able to print inheritance to stdout (to quickly analyze things), examples:
* see what inherits from BaseContract: `BaseContract childrens: A, B, C, D, E, ...` (it should also list the contracts that derives from A, B, C, D, E)
* see all base classes for given contract: `Children inherits from: BaseContract1, BaseContract2, BaseBaseContract, SuperBaseBaseBaseContract`

pvgupta24:
Resolves #23 
@disconnect3d @montyly  Please take a look :)


slither/printers/inheritance/inheritance.py
```diff
@@ -1,123 +1,45 @@
...
-            filename += ".dot"
-        info = 'Inheritance Graph: ' + filename
+        info = 'Inheritance\n'
+        info += blue('Child_Contract -> ') + green('Base_Contracts')
+        for contract in self.contracts:
+            info += blue(f'\n+ {contract.name} -> ')
+            if contract.inheritance:
+                info += green(", ".join(map(str, contract.inheritance)))
+            else:
+                info += magenta("Root_Contract")
```
pvgupta24: Contracts without any children, and without any parent are also logged as of now. Is this fine?

montyly: I think it's ok to print them, but it would be less confusing to just print an empty list here

montyly:
Thanks for your contribution, the PR looks good!
Could you add an example of use here https://github.com/trailofbits/slither/tree/master/examples/printers ?

pvgupta24:
@montyly made the requested changes and added example for the same. PTAL :)

PR #52, opened 2018-10-21
Add exceptions to the naming convention rules

montyly:
Add of some exceptions to the Solidity recommended naming convention:

- Allow constant variables `name`/`symbol`/`decimals` to be lowercase (ERC20)
- Allow '_' at the beginning of the mixed_case match for private variables and unused parameters

If nobody disagrees with these exceptions, I will merge the PR


redshark1802:
First of all, I'm against making exceptions for certain constant types to be lower case. I know instantly that a variable is a constant when I see CONSTANT_NAME. The whole benefit of a naming convention is to be consistent and increase readability, though the solidity docs mention "it should be" instead of "must be" (although the docs are not written RFC-style).

Prefixing privates with '_' is a mixed topic across languages and I have no strong argument for either style but this would diverge from the official documentation.

montyly:
Thanks for your feedback!

I do agree that we should avoid exceptions, though if something is reported really frequently and don't need change, I think it becomes more annoying for the user than useful.

Typically several ERC20 tokens implement the `name`/`symbol`/`decimals` as public constant variables. To be consistent with the naming convention, they should rename these variables to uppercase, and create public functions to return the values. This would increase the Solidity code size, without a real benefit from my point of view.

redshark1802:
@montyly You are right, I somehow missed that you mentioned ERC20.

slither/detectors/naming_convention/naming_convention.py
```diff
@@ -21,6 +25,12 @@ def is_cap_words(name):
     def is_mixed_case(name):
         return re.search('^[a-z]([A-Za-z0-9]+)?_?$', name) is not None
 
+    @staticmethod
+    def is_mixed_case_with_underscore(name):
+        # Allow _ at the beginning to represent private variable
+        # or unused parameters
+        return re.search('^[a-z_]([A-Za-z0-9]+)?_?$', name) is not None
```
disconnect3d:
```suggestion
        return re.search('^[a-z_][A-Za-z0-9]*_?$', name) is not None
```
Also we should compile the regex `SOME_REGEX = re.compile(...)` and use it here.

slither/detectors/naming_convention/naming_convention.py
```diff
@@ -80,8 +90,11 @@ def detect(self):
                                     'sourceMapping': func.source_mapping})
 
                 for argument in func.parameters:
-
-                    if self.is_mixed_case(argument.name) is False:
+                    if argument in func.variables_read_or_written:
+                        incorrect_naming = self.is_mixed_case(argument.name) is False
+                    else:
+                        incorrect_naming = self.is_mixed_case_with_underscore(argument.name) is False
+                    if incorrect_naming:
```
disconnect3d: the `incorrect_naming` and `is False` is tricky/not so easy to understand.

Maybe we should revert it and make it a `correct_name`? This way we won't have `is False` twice and we will just do at the end `if not correct_name:`

slither/detectors/naming_convention/naming_convention.py
```diff
@@ -119,7 +136,11 @@ def detect(self):
                                         'constant': var.name,
                                         'sourceMapping': var.source_mapping})
                 else:
-                    if self.is_mixed_case(var.name) is False:
+                    if var.visibility == 'private':
+                        incorrect_naming = self.is_mixed_case_with_underscore(var.name) is False
+                    else:
+                        incorrect_naming = self.is_mixed_case(var.name) is False
+                    if incorrect_naming:
```
disconnect3d: Same as above

tests/naming_convention.sol
```diff
@@ -52,6 +52,14 @@ contract Test {
 }
 
 contract T {
+    uint private _myPrivateVar;
+    uint _myPublicVar;
+
+
+    function test(uint _unused, uint _used){
+        _used;}
```
disconnect3d: Is that valid in Solidity? oO'

Issue #29, opened 2018-10-17
Detector: Detect public functions that could be declared as external


montyly:
If a public function is never called, it should be marked as external, so that it can be optimized by solc
- Severity: Informational

Hints:Use [slithIR](https://github.com/trailofbits/slither/tree/56079226efc78a83438c55b91ce8dabbb0b7813c) and check:
- the function is never the destination of an [InternalCall](https://github.com/trailofbits/slither/blob/56079226efc78a83438c55b91ce8dabbb0b7813c/slither/slithir/operations/internal_call.py)
- there is no [InternalDynamicCall](https://github.com/trailofbits/slither/blob/56079226efc78a83438c55b91ce8dabbb0b7813c/slither/slithir/operations/internal_dynamic_call.py) (or ensure that the function is never the destination of an `InternalDynamicCall`)
- Iterate only over the derived contracts ([slither.contracts_derived](https://github.com/trailofbits/slither/blob/56079226efc78a83438c55b91ce8dabbb0b7813c/slither/core/slither_core.py#L31)) to be sure that no inherited contract calls the function

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it as part of the Ethereum Community Fund via ECF Web 3.0 Infrastructure Fund fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/trailofbits/slither/29/1518).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/trailofbits/slither/29/1518).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $48,205.92 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

samparsky:
@montyly I am interested in this issue, would like to work on it

montyly:
Great!
For information, I pushed `slithir` into master, and started its documentation here: https://github.com/trailofbits/slither/wiki/SlithIR

[...]

PR #53, opened 2018-10-22
Detect public functions to be declared as external

samparsky:
Fixes issue #29 

@montyly Can you help review


montyly:
Thanks for your contribution!

- `INTERNAL_DYNAMIC_CALL` `.function` returns a `Variable`, and not a `Function`; this is because it's a dynamic call which may point to many different functions (and the set of potential destinations is not yet computed). Finding all the destinations is out-of-scope of this detector, I think we can simply discard the detector if we find any `INTERNAL_DYNAMIC_CALL` in the contract

- If the function is called through a `HighLevelCall`, it should not be considerd as non-potential external function, a `HighLevelCall` is an external call to another contract.

- Ather some thinking, I think `def detect_public(self, contract):` could just return `contract.all_functions_called`, which returns all the functions reachable by the contract

slither/detectors/functions/external_function.py
```diff
@@ -0,0 +1,79 @@
...
+        """ Returns a list of InternallCall, InternalDynamicCall, SolidityCall, HighLevelCall
+            calls made in a function
+
+        Returns:
+            (list): List of all InternallCall, InternalDynamicCall, SolidityCall, HighLevelCall
+        """
+        result = []
+        for node in func.nodes:
+            for ir in node.irs:
+                if isinstance(ir, ( InternalCall, InternalDynamicCall, HighLevelCall, SolidityCall )):
```
disconnect3d:
```suggestion
                if isinstance(ir, (InternalCall, InternalDynamicCall, HighLevelCall, SolidityCall)):
```

samparsky:
@monthly Can you clarify further what you mean by this?
>  I think we can simply discard the detector if we find any INTERNAL_DYNAMIC_CALL in the contract

montyly:
Without an additional analysis, you can't know the set of potential functions executed through an `InternalDynamicCall` as its computed at runtime (through its possible to compute it statically, slither doesn't do it yet).

Consider the following example:
```
contract C{
    
    function () returns(uint) ptr;
    
    
    function set_test1() external{
        ptr = test1;
    }
    
    function set_test2() external{
        ptr = test2;
    }
    
    function test1() public returns(uint){
        return 1;
    }
    
    function test2() public returns(uint){
        return 2;
    }
    
    function test3() public returns(uint){
        return 3;
    }
    
    function exec() external returns(uint){
        return ptr();
    }
    
}
```

Here you have three functions that are `public`, and are candidates to become `external`. But only `test3()` can, because `ptr` is never assigned to `test3`

With the IR we can compute the set of function that can be associated to `ptr`, though I think it's simpler to just not analyze a contract if it has a `InternalDynamicCall` for this detector

samparsky:
@montyly I am having issues compiling the example contract you gave above. The slithIR  gives the error below

```
 raise VariableNotFound('Variable not found: {}'.format(var_name))
slither.solc_parsing.expressions.expression_parsing.VariableNotFound: Variable not found: ptr()
```

samparsky:
@montyly Its working okay after doing a rebase on the master branch

samparsky:
@montyly Kindly review,
I have added that a contract should be ignored if it contains an InternalDynamicCall

PR #54, opened 2018-10-22
[WIP] Support of compact AST

montyly:
WIP to support solc's compact AST (https://github.com/trailofbits/slither/issues/22).

Support is not optimal yet. For example, we do not use any new features of the compact AST.

Once we decide that our support for the compact AST is stable, we can remove support for the original AST to simplify the parsing. 

Completing support for the compact AST will allow Truffle integration (https://github.com/trailofbits/slither/issues/18).


slither/detectors/naming_convention/naming_convention.py
```diff
@@ -81,7 +81,11 @@ def detect(self):
 
                 for argument in func.parameters:
 
-                    if self.is_mixed_case(argument.name) is False:
+                    argument_name = argument.name
+                    if argument_name and argument_name.startswith('_'):
+                        argument_name = argument_name[1::]
```
disconnect3d:
```suggestion
                        argument_name = argument_name[1:]
```

disconnect3d:
```suggestion
                        argument_name = argument_name[1:]
```

or eventually: `argument_name = argument_name.lstrip('_')` but that would remove all underscores from the begining.

slither/detectors/naming_convention/naming_convention.py
```diff
@@ -119,7 +123,12 @@ def detect(self):
                                         'constant': var.name,
                                         'sourceMapping': var.source_mapping})
                 else:
-                    if self.is_mixed_case(var.name) is False:
+
+                    var_name = var.name
+                    if var_name and var_name.startswith('_'):
+                        var_name = var_name[1::]
```
disconnect3d:
```suggestion
                        var_name = var_name[1:]
```

benstew:
Looking good @montyly! Really excited to support the compact AST. While doing a little research, I was surprised by the lack of information/documentation regarding JSON ASTs that was available. Going with the compact AST essentially just got rid of the nested attributes object to include the properties on the node itself. But other than that change and a slightly cleaner output, is there anything else special that we should highlight in the documentation and elsewhere? Or was this primarily driven by the `--ast-json` deprecation rumblings?

montyly:
Yes, you are right, the compact AST got rid of the nested attributes and have, overall, a cleaner structure. 
It provides also more information to recover the relations between elements, for example, `Identifier` has a `referencedDeclaration` field, which allows knowing if two  `Identifier` point to the same object. 
Right now, slither don't use any additional info from the compact format that is not in the original AST, and recover everything by itself. If slither was supporting only the compact format, we could simplify the parsing and the recovery of information.

There were two main goals for supporting the compact ast: the future deprecation of the original ast, as you mentioned, and a better support for Truffle, which outputs by default the compact AST in the `build/contracts` directory.

I am not aware of an official documentation for the AST, or the compact AST, though I would love one.

benstew:
Thanks @montyly. Yes, I'm seeing the addition of a few new fields in the compact AST that could be leveraged in Slither once we support. But I think Truffle is the big win here.

Also, thanks for calling out the change `referenceDeclaration`.  I was looking around and not a ton of documentation available about the JSON AST. There is so little documentation that others feel our pain and are calling out additional documentation + test coverage [here](https://github.com/ethereum/solidity/issues/4758)

The 0.5.0 solidity [changelog](https://github.com/ethereum/solidity/blob/a434896458656249113c7e269d8fde64b3df5d69/Changelog.md) provided some insight but still not sure when that will be released. There are two changes to the JSON AST that I wanted to highlight:

1. JSON AST: Remove constant and payable fields (the information is encoded in the stateMutability field)
2. JSON AST: Replace the isConstructor field by a new kind field, which can be constructor, fallback or function.

Let me know if you think I should open up a new issue to tackle those.  Seems like we could support easily with a few changes to [function.py](https://github.com/trailofbits/slither/blob/master/slither/solc_parsing/declarations/function.py#L39 ) following the `is_compact_ast`property decorator pattern you are using in this PR.

montyly:
Yes you are right! Solidity 0.5.0 will bring minor modifications to the AST, such as the ones you highlighted. We also need to integrate grammar changes, such as the field `payable` for addresses.

I am waiting to see the final version of 0.5.0 before applying any modification.

PR #55, opened 2018-10-22
WIP: cli improvements

### LINKED ISSUE #50: Modify command line arguments
The command line arguments are setup differently than I would expect for Slither. I wonder what people think about changing them? I would have expected some of these changes:

* `--detect` which takes a comma-separated list indicating the checks to run (default: all)
* `--print` which takes a comma-separated list of printers to output (default: ???)
* `--output` which accepts a single parameter of either `stdout` or `json` (default: stdout)
* remove exclusion rules entirely
* calling slither with no arguments should print the help
* reorganize the help screen into groups  of related parameters for easier reading

redshark1802:
- help is printed when no arguments are supplied
- rework detector arguments: detectors are now a comma-separated list, defaults to all
- rework printer arguments: printers are now a comma-separated list, defaults to contract-summary
- add version command

fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,32 +36,39 @@ If Slither is run on a directory, it will run on every `.sol` file of the direct
 
 ### Printers
 
-* `--printer-summary`: Print a summary of the contracts
-* `--printer-quick-summary`: Print a quick summary of the contracts
-* `--printer-inheritance`: Print the inheritance relations
-* `--printer-inheritance-graph`: Print the inheritance graph in a file
-* `--printer-vars-and-auth`: Print the variables written and the check on `msg.sender` of each function
+By default, the `contract-summary` printer is used. Use --printers comma-separated list of printers, 
+or `none` to disable the default printer. 
 
-## Checks available
+Num | Printer | Description
+--- | --- | ---
+1 | `contract-summary` | a summary of the contract
+2 | `function-summary` | the summary of the functions
+3 | `inheritance` | the inheritance relation between contracts
+4 | `inheritance-graph` | the inheritance graph
+5 | `slithir` | the slithIR
+6 | `vars-and-auth` | the state variables written and the authorization of the functions
 
-By default, all the checks are run. Use --detect-_name-of-check_ to run one check at a time.
+## Detectors
 
-Num | Check | What it Detects | Impact | Confidence
+By default, all the detectors are run. Use --detectors comma-separated list of detectors to run.
+
+Num | Detector | What it Detects | Impact | Confidence
 --- | --- | --- | --- | ---
-1 | `suicidal` | Suicidal functions | High | High
-2 | `uninitialized-state` | Uninitialized state variables | High | High
-3 | `uninitialized-storage` | Uninitialized storage variables | High | High
-4 | `arbitrary-send` | Functions that send ether to an arbitrary destination | High | Medium
-5 | `reentrancy` | Reentrancy vulnerabilities | High | Medium
-6 | `locked-ether` | Contracts that lock ether | Medium | High
-7 | `tx-origin` | Dangerous usage of `tx.origin` | Medium | Medium
-8 | `assembly` | Assembly usage | Informational | High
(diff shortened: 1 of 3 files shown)


slither/__main__.py
```diff
@@ -228,9 +225,7 @@ def main_impl(all_detector_classes, all_printer_classes):
 
 
 def parse_args(detector_classes, printer_classes):
-    parser = argparse.ArgumentParser(description='Slither',
-                                     usage="slither.py contract.sol [flag]",
-                                     formatter_class=lambda prog: argparse.HelpFormatter(prog, max_help_position=35))
+    parser = argparse.ArgumentParser()
```
benstew: Nitpick - I realize this is a WIP but the current output doesn't include the `Slither` description/title and we are duplicating arguments. 

screen shot 2018-10-22 at 5 05 20 pm

Seems like an easy fix. We could utilize the same parameters from before and pass them into our parser. Our output would be more consistent with what we had before and it removes that duplication:
```python
    parser = argparse.ArgumentParser(description='Slither',
                                     usage="slither.py contract.sol [flag]")
```
screen shot 2018-10-22 at 5 29 45 pm

This approach tends to stay consistent with other CLIs. Our messaging isn't ideal in the case that someone passes in a bad arguement (eg `abc12345`) but we still surface a useful message. Don't think we need to kill ourselves anyway for optimizing to that use case.

benstew:
@redshark1802 - Looking good. Just provided some brief feedback to keep the output consistent with how it was previously.

slither/__main__.py
redshark1802: thx for the feedback, fixed

slither/__main__.py
```diff
@@ -280,14 +279,21 @@ def parse_args(detector_classes, printer_classes):
...
-        detector_arg = '--detect-{}'.format(detector_cls.ARGUMENT)
-        detector_help = '{}'.format(detector_cls.HELP)
-        parser.add_argument(detector_arg,
-                            help=detector_help,
-                            action="append_const",
-                            dest="detectors_to_run",
-                            const=detector_cls.ARGUMENT)
+    parser.add_argument('--detectors',
+                        help='Comma-separated list of detectors, defaults to all, '
+                             'available detectors: {}'.format(', '.join([d.ARGUMENT for d in detector_classes])),
```
disconnect3d:
```suggestion
                             'available detectors: {}'.format(', '.join(d.ARGUMENT for d in detector_classes)),
```

slither/__main__.py
```diff
@@ -280,14 +279,21 @@ def parse_args(detector_classes, printer_classes):
...
+                        action='store',
+                        dest='detectors_to_run',
+                        default='all')
+
+    parser.add_argument('--printers',
+                        help='Comma-separated list fo contract information printers, '
+                             'defaults to contract-summary and can be disabled by using \'none\', '
+                             'available printers: {}'.format(', '.join([d.ARGUMENT for d in printer_classes])),
```
disconnect3d:
```suggestion
                             'available printers: {}'.format(', '.join(d.ARGUMENT for d in printer_classes)),
```

README.md
```diff
@@ -36,17 +36,21 @@ If Slither is run on a directory, it will run on every `.sol` file of the direct
...
-* `--printer-quick-summary`: Print a quick summary of the contracts
-* `--printer-inheritance`: Print the inheritance relations
-* `--printer-inheritance-graph`: Print the inheritance graph in a file
-* `--printer-vars-and-auth`: Print the variables written and the check on `msg.sender` of each function
+By default, the `contract-summary` printer is used. Use --printers comma-separated list of printers, 
+or `none` to disable the default printer. 
 
-## Checks available
+* `contract-summary`: Print a summary of the contracts
+* `quick-summary`: Print a quick summary of the contracts
```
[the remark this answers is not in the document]
redshark1802: The printer `quick-summary` doesn't exist, I removed it from the list.

PR #56, opened 2018-10-22
implementation of call graph printer

### LINKED ISSUE #44: Printer: Callgraph printer
A callgraph printer would show for each function, what are the contracts/functions called.

For example, on
```
contract ContractA{
    function my_func_a(){}
}

contract ContractB{
    
    ContractA a;
    
    constructor(){
        a = new ContractA();
    }
    
    function my_func_b(){
        a.my_func_a();
        my_second_func_b();
    }
    
    function my_second_func_b(){
    }
}
```
It should show that `ContractB.my_func_b()` calls `ContractA.my_func_a` and `ContractB.my_second_func_b()`

The printer will output a dot, similar to the inheritance printer. 

It's probably more readable to group the functions of a contract into a dot `subgraph`

evgeniuz:
Fixes #44


evgeniuz:
So far implemented internal calls printing, plan to proceed by optimizing internal call rendering a bit and then adding external calls (between classes).

I have couple of questions:

1. Should this printer display calls to builtin solidity functions (i. e. solidity functions are grouped in their own subgraph with calls pointing to them, right now it's implemented this way, but it's easy to hide calls to builtins)?
2. Should it display functions that have no calls (neither internal nor external), i. e. just display node without any arrows?

RehxFnek:
Call Graph, in Application Discovery Browser point of view, from the Projects list in the focal piece of the Explore Projects see, select the undertaking that contains the asset for which you need to produce a Program Call Graph. Use [Brother Printer Support](https://www.printererrorrepair.com/brother-printer-support-phone-number) for print your call graph.

evgeniuz:
Implemented according to requirements, printer supports internal/external calls (grouped per contract in subgraphs) as well as calls to builtins (which are grouped in their own subgraph).

montyly:
Thanks for your contribution! The PR looks good, I will review it in detail tomorrow.

I think it makes sense to display the calls to builtin solidity functions

montyly:
So after some testing, I realized two issues:
- Library calls were not linked
- A call to a public state variable created an incorrect node

This is mostly because `function.external_calls` returns a list of CallExpression, which are complex to parse.

I simplify the API in https://github.com/trailofbits/slither/pull/58 (branch `dev-refactor-nodes-info`) 
The `function.external_calls` was removed and replaced by:

- `solidity_calls`: which returns a list of `SolidityFunction`
- `high_level_calls`, which returns list of `(Contract, Function|Variable)`. `Variable` is used if for calls to a public state variables. `high_level_calls` includes library calls.
- `low_level_calls`, which you can ignore I think

Merging or rebase from `dev-refactor-nodes-info` should fix the two issues, but necessitates to change the `_external_function_node` function

evgeniuz:
Agree, this approach is much better than parsion `Expression` in printer. Will rebase from `dev-refactor-nodes-info`) and refactor.

montyly:
Great, I merged https://github.com/trailofbits/slither/pull/58 is merged in `dev`, that should simplify the rebase

evgeniuz:
Done, also changed example to have library/variable access.

Issue #65, opened 2018-10-30
vars-and-auth printer output bug


rmi7:
When using the `vars-and-auth` printer there is a bug in the output.

## Problem Description
The entire table is written out on each iteration of going through the functions of a contract, instead of printing out once after the for each function loop is done.

## Example output currently

```
INFO:Printers:
Contract MyContract
+----------+-------------------------+---------------------------------------------+
| Function | State variables written |           Conditions on msg.sender          |
+----------+-------------------------+---------------------------------------------+
|  setCEO  |       ['ceoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
+----------+-------------------------+---------------------------------------------+
INFO:Printers:
Contract MyContract
+----------+-------------------------+---------------------------------------------+
| Function | State variables written |           Conditions on msg.sender          |
+----------+-------------------------+---------------------------------------------+
|  setCEO  |       ['ceoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCMO  |       ['cmoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
+----------+-------------------------+---------------------------------------------+
INFO:Printers:
Contract MyContract
+----------+-------------------------+---------------------------------------------+
| Function | State variables written |           Conditions on msg.sender          |
+----------+-------------------------+---------------------------------------------+
|  setCEO  |       ['ceoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCMO  |       ['cmoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCSO  |       ['csoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
+----------+-------------------------+---------------------------------------------+
INFO:Printers:
Contract MyContract
+----------+-------------------------+---------------------------------------------+
| Function | State variables written |           Conditions on msg.sender          |
+----------+-------------------------+---------------------------------------------+
|  setCEO  |       ['ceoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCMO  |       ['cmoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCSO  |       ['csoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
|  setCFO  |       ['cfoAddress']    | ['require(bool)(msg.sender == ceoAddress)'] |
+----------+-------------------------+---------------------------------------------+
```

## Expected Output

just the last one..

[...]

PR #66, opened 2018-10-30
fix vars-and-auth print bug

rmi7:
fixes issue #65


montyly:
Thank you for the fix!

Issue #63, opened 2018-10-30
Use json output for travis test


montyly:
Right now the travis tests only check for the number of results returned:

https://github.com/trailofbits/slither/blob/39f560bf973acbc31c908991f94d4cb57a7e4828/scripts/travis_test.sh#L5-L16

It would make more sense to test the json output.

- [ ] Generate pre-defined json for each testcase (used as ground true)
- [ ] Update scripts/travis_test.sh to export the result in a json file (`--json`)
- [ ] For each testcase, compare the generated json with the corresponding pre-defined json
- [ ] If if an issue is found, print the difference
- [ ] Clean the generated json file(s) (but not the ground true)

rmi7:
fixed in PR #67

PR #67, opened 2018-10-31
Use json output for travis test

rmi7:
implements #63

- [x] Generate pre-defined json for each testcase (used as ground truth)
  - see `scripts/tests_generate_expected_json.sh`
- [x] Update scripts/travis_test.sh to export the result in a json file (--json)
- [x] For each testcase, compare the generated json with the corresponding pre-defined json
  - using `diff`
- [x] If if an issue is found, print the difference
  -  when there is a difference, printing out test info (file/detector) + `diff` output
- [x] Clean the generated json file(s) (but not the ground true)
  - all "temp" files are removed

extra
- used `python -m json.tool` to generate pretty printed json output, otherwise `diff` output will be the entire file (since it's 1 line), not just the part that "differs"


montyly:
Thanks for your contribution. The PR looks good!

I like the use of `json.tool`

For some reason, some pre-generated json files are incorrect (they are empty), for example travis shows:
```
\nfailed test of file: tests/backdoor.sol, detector: backdoor\n
1c1,11
< []
---
> [
>     {
>         "vuln": "backdoor"
>         "contract": "C"
>         "sourceMapping": {
>             "start": 42
>             "length": 74
>             "filename": "tests/backdoor.sol"
>         }
>     }
> ]\n
```

Maybe we should also move the pre-generated json to a sub directory? so that we don't confound the json that are the slither results, and the json that are the contracts precompiled to ast (like [tests/old_solc.sol.json](https://github.com/trailofbits/slither/blob/master/tests/old_solc.sol.json))

rmi7:
ye I saw that too in travis lat night, investigating that now.

where do you suggest to place the pre-gen json files, `tests/expected_json/`?

rmi7:
I'm getting unexpected results locally, 

running the original `scripts/travis_test.sh` locally in `master` returns: 

```
INFO:Slither:tests/<contract-name>.sol analyzed (0 contracts), 0 result(s) found
```

for the following tests:

```
test_slither tests/backdoor.sol "backdoor" 1
test_slither tests/backdoor.sol "suicidal" 1
test_slither tests/pragma.0.4.24.sol "pragma" 1
test_slither tests/unused_state.sol "unused-state" 1
test_slither tests/locked_ether.sol "locked-ether" 1
```

could you verify this is also happening for you or that it's just me

montyly:
You can name the directory `expected_json`.

I have the same execution of the travis script than travis-ci.com locally.
What is your commit? Can you try:
```
$ slither tests/backdoor.sol 
INFO:Detectors: Backdoor function found in C.i_am_a_backdoor
INFO:Detectors: Suicidal function in tests/backdoor.sol Contract: C, Function: i_am_a_backdoor
INFO:Detectors: Function 'i_am_a_backdoor' is not in mixedCase, Contract: 'C' 
INFO:Detectors: Public function in tests/backdoor.sol Contract: C, Function: i_am_a_backdoor should be declared external
```
(tested on 014bce3a8552eb8da6deced86a103674f73e9302)

rmi7:
after more testing it appears `python -m json.tool` sorts keys, and the output order might differ on each run.
Looking into replacing it with with something else to pretty print json, but limited to python and bash I guess.

rmi7:
locally, all json outputs match when running hte test. 

could it be that the order in which detectors output results is not deterministic?

rmi7:
after quite the debugging python and travis adventure I think it's working now.

my findings:
- it appears that the detectors output json order is not deterministic (I can sort json keys using `json.dump()`, but not based on values), so I made it deterministic by using `sorted()` inside each detectors `detect(self)` function to sort based on sepcific values (`contract.name`, etc.)
- had to use custom python script to pretty print sorted json output, `python -m json.tool` does not seem to generate key-sorted output on travis 🤔
- had to explicitly set `separators(',',': ')` in the `json.dumps()` call to make it work on travis 🤔

montyly:
It looks good!

`json.tool` has the `--sort-keys` option, would that solve the problem without having to sort the json output in the detectors?

rmi7:
no it would not, for example with this output

```json
[
    {
        "contract": "Test",
        "filename": "tests/uninitialized.sol",
        "functions": [
            "use"
        ],
        "sourceMapping": [
            {
                "filename": "tests/uninitialized.sol",
                "length": 34,
                "start": 189
            },
            {
                "filename": "tests/uninitialized.sol",
                "length": 143,
                "start": 356
            }
        ],
        "variable": "balances",
        "vuln": "UninitializedStateVars"
    },
    {
        "contract": "Test2",
        "filename": "tests/uninitialized.sol",
        "functions": [
            "use"
        ],
        "sourceMapping": [
            {
                "filename": "tests/uninitialized.sol",
                "length": 15,
                "start": 695
            },
            {
                "filename": "tests/uninitialized.sol",
                "length": 117,
                "start": 875
            }
        ],
        "variable": "st",
        "vuln": "UninitializedStateVars"
    },
    {
        "contract": "Test2",
        "filename": "tests/uninitialized.sol",
        "functions": [
            "init"
        ],
        "sourceMapping": [
            {
                "filename": "tests/uninitialized.sol",
                "length": 6,
                "start": 748
            },
            {
                "filename": "tests/uninitialized.sol",
                "length": 52,
                "start": 817
            }
        ],
        "variable": "v",
        "vuln": "UninitializedStateVars"
    },
    {
        "contract": "Uninitialized",
        "filename": "tests/uninitialized.sol",
        "functions": [
            "transfer"
        ],
        "sourceMapping": [
            {
                "filename": "tests/uninitialized.sol",
                "length": 19,
                "start": 55
            },
            {
                "filename": "tests/uninitialized.sol",
                "length": 82,
                "start": 81
            }
        ],
        "variable": "destination",
        "vuln": "UninitializedStateVars"
    }
]
```

On travis I would get results where the order of json objects woul differ, even though I can sort on keys, what I need is to be able to sort on the value of specific keys, so that for example I always first get `Test`, then `Test2`, then `Uninitialized`, if there are multiple json objects for 1 contract (`Test2` output in above example) I need a way to get those json objects in a deterministic order, that's why after sorting on the value of `"contract"` I also sort on `"variable"` (depending on detector this could be somethign else like function name), 

so sorting on key only gets me deterministic property name order inside each json object, but I also need deterministic order of json objects, and for that I need to sort on the value of specific properties

See this travis build output for what I mean: https://travis-ci.com/trailofbits/slither/builds/89895099

montyly:
oh I see! Ordering elements are indeed an issue.

Alternatively, we could write a small python script that compare the two json, and [recursively order the elements](https://stackoverflow.com/a/25851972), so that lists and dicts are ordered?
I can help to write it.

My concern is that it sounds more maintainable to not assume any sorting from the detectors.

rmi7:
that sounds like a good idea, I think I am capable to write that myself

you are right, it would be cleaner not to have to sort every detectors output (en thus edit each detectors python file)

rmi7:
seems ot be working (had this bug locally where python2 would execute the script, even though I aliased python to python3) 

created a script which will take in a json file path (expecting an array at top level), and recursively sorts all object's keys, and the objects in a list based on the values of their keys (check the python script file, it contains comments), it should work on any json file (with an array at the top level)

if you see some python code that looks weird, blame it on me, I've never written any python before this PR. so improvements/tips are welcome.

could you also manually rerun the travis a couple of times to see that it works.

rmi7:
after some more commits to improve comments/code it still seems to work on travis, so probably you don't need to rerun travis

montyly:
Great work @rmi7! 

I merged your PR in dev, so that the json takes advantage of the line numbers information, and some small fixes (like `order_list` does not need to filter on None value anymore)

rmi7:
ye that (`null` values in output json) should be fixed, great 👍

PR #74, opened 2018-11-07
Normalize detectors json output function properties

rmi7:
## Current situation

json output of the different detectors return different property name for the name of the function in which something was detected: `func`, `function`, `function_name`, and `funcs`, `functions`.

## Wanted Situation

they all use the same property name: `function` and `functions`.


montyly:
Related issue: https://github.com/trailofbits/slither/issues/33

montyly:
Thank for your contribution @rmi7, this convention makes more sense!

Issue #76, opened 2018-11-09
Shorten --exclude-detectors to just --exclude


dguido:

montyly:
Done in https://github.com/trailofbits/slither/commit/82d4730f125c1101c923e4121c1f0f0c74027f67

PR #68, opened 2018-11-02
[WIP] Solidity detector for unused return calls

6ug:
[DO NOT REVIEW]
Fixes https://github.com/trailofbits/slither/issues/45

TODO -
* `total_values` should probably only add if the IR is a `HighLevelCall`

PR #78, opened 2018-11-12
Detector: Unused return value of a call

### LINKED ISSUE #45: Detector: Unused return value of a call
If the return value of a function is never used, it's likely to be bug.

- severity low/medium

SlithIR can be used to track if the value returned by a call is used:
- If the function does not return a value, the  `lvalue` of the operation is `None`, otherwise its a variable.
- We can check if the `lvalue` is in the `read` list of another operation

Example: 

```
contract Target{
    function f() returns(uint);
}

contract Function{

    function test(Target t){
        t.f();
    }

}
```

benstew:
Fixes #45, #76 

Not trying to step on any toes but it looks like #68 is inactive so I took a stab at putting together a detector for unused return values of a call.

Local testing of detector:
screen shot 2018-11-12 at 10 44 01 am

Local testing of updated `--exclude` syntax:
screen shot 2018-11-12 at 9 58 55 am

Todo:
- [x] Compare against dev branch (compatible with travis testing integration)
- [ ] Clean up commit history after switching base branch
- [ ] Create expected_json file for travis test coverage


slither/detectors/operations/unused_return_values.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        results = []
+        for c in self.slither.contracts_derived:
+            unused_return_values = self.detect_unused_return_values(c)
+
+            if unused_return_values:
+                funcs_with_unused_return_value_by_contract = defaultdict(list)
+
+                for unused_return_value in unused_return_values:
+                    funcs_with_unused_return_value_by_contract[unused_return_value.contract.name].append(unused_return_value.function.name)
```
benstew: Naming is a little verbose. Wanted to add clarity. Happy to shorten up.

slither/detectors/operations/unused_return_values.py
```diff
@@ -0,0 +1,71 @@
...
+from slither.slithir.operations.high_level_call import HighLevelCall
+
+class UnusedReturnValues(AbstractDetector):
+    """
+    If the return value of a function is never used, it's likely to be bug
+    """
+
+    ARGUMENT = 'unused-return'
+    HELP = 'Unused return values'
+    IMPACT = DetectorClassification.LOW
```
benstew: Open to any feedback here on the impact/confidence. Went with a low/medium but was on the fence.

PR #81, opened 2018-11-20
[WIP] Add dataflow/abstract interpretation framework

montyly:
- [x] Add dataflow framework to facilitate the creation of dataflow analysis/abstract interpretation
- [x] Rewrite the reentrancy detector to use the framework
- [ ] Investigate if other public detectors could take advantage of the framework
- [ ] Document the framework
- [ ] Add user-level script examples

### DIFF
--- a/slither/analyses/dataflow/generic.py
+++ b/slither/analyses/dataflow/generic.py
@@ -1,0 +1,50 @@
+from abc import ABC, abstractmethod
+
+class Dataflow(ABC):
+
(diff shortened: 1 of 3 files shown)


montyly:
I am closing this PR.

While having a general dataflow/abstract interpretation framework would be nice to have, we did not have a significant need for it. 

This PR is out of sync, if we want such a generic framework in the future it will be better to start it from scratch.

PR #86, opened 2018-12-04
Use --user to show how to install slither from git

ggrieco-tob:


montyly:
Good catch!

For users using git, maybe we should detail how to use virtualenv/virtualenvwrapper rather than local installation?

disconnect3d:
We don't have `--user` in `pip` installation too. I think we could just write a line like `Please note that it is best to install it under a separate Python virtual environment (so the dependencies versions don't conflict with other installed packages).` instead of choosing to use `--user`.

montyly:
I added developer install instructions in the wiki: https://github.com/trailofbits/slither/wiki/Developer-installation

I am closing this PR and I will add the wiki to the README

PR #93, opened 2018-12-08
Fixed Docker - slither not compatible with solc > 0.5.x atm, so …

mihairaulea:
…downgrade solc to 0.4.25


montyly:
Good catch, thank you!

We are going to update the docker image on dockerhub soon.

PR #132, opened 2019-01-14
[WIP] Expose immediate base contracts and base constructors

### LINKED ISSUE #131: Expose immediate base contracts for each contract
Currently slither API provides `contract.inheritance` which refers to linearized base contracts (a flattened list of the inheritance hierarchy). This requests support for the immediate base contracts inherited from for any given contract. 

For example:
```
contract A {
    // ...
}

contract B is A {
    // ...
}

contract C is A {
    // ...
}

contract D is B, C {
    // ...
}
```

I would like to be able to determine that the base contracts for `D` are `B, C` (in order, even if reversed), where as linearized base contracts would yield `C, B, A`

Xenomega:
This pull request aims to resolve #131 . It aims to provide the immediate base contracts of a given contract in the correct order of declaration via `contract.immediate_inheritance`. This contrasts to the existing provided `contract.inheritance` which provides a flattened, linearized base contract list (all ancestors in the inheritance chain, immediate and not).

### DIFF
--- a/scripts/json_diff.py
+++ b/scripts/json_diff.py
@@ -7,10 +7,10 @@ if len(sys.argv) !=3:
     print('Usage: python json_diff.py 1.json 2.json')
     exit(-1)
 
-with open(sys.argv[1]) as f:
+with open(sys.argv[1], encoding='utf8') as f:
     d1 = json.load(f)
 
-with open(sys.argv[2]) as f:
+with open(sys.argv[2], encoding='utf8') as f:
     d2 = json.load(f)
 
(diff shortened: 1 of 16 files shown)


Xenomega:
This pull request has now been refactored. It additionally adds support for both AST formats. It also now exposes base constructor calls on both contract definitions and constructor definitions.

NOTE: Please verify legacy-ast heuristics are sufficient, as legacy-ast is known to present unreliable nodes at times. All tests to this point have yielded expected results.

PR #148, opened 2019-01-21
[WIP] Add feature to review upgradable contracts

montyly:
Add printers and external tools to review a contract upgradability based on `delegatecall`

New printers:
- `function-id`: Print the heccack256 signature of the functions
- `variables-order`: Print the storage order of the state variables

Two utils:
- `compare_function_ids.py`: check if the proxy and the implementation have a function id collision
- `compare_variables_order.py`: check if the state variables order is respected

It's a WIP, further tests must be done to ensure that `compare_variables_order.py` catch all the corner cases


utils/upgradability/compare_variables_order.py
```diff
@@ -0,0 +1,56 @@
+'''
+    This utility looks for functions collisions between a proxy and the implementation
```
bmperrea: nit: stale comment?

PR #161, opened 2019-02-04
Fix UserDefinedTypeName type parsing from names

### LINKED ISSUE #137: Error when parsing type which refers to an aliased import
The following contracts will cause slither to error:

Main.sol:
```
import {Dependency as x} from "./Dependency.sol";

contract Main {
    x y;
}
```

Dependency.sol:
```
contract Dependency {
    uint y;
}
```

Error:
```
ERROR:TypeParsing:Type not found x
```

### LINKED ISSUE #160: Misintrepretation of types on collision of UserDefinedTypeName
The following contract will be misinterpreted by Slither when using legacy AST:
```
pragma solidity ^0.4.0;

contract A {

    struct CollidingName {
        uint x;
    }

    mapping(uint => CollidingName) public cycles; // refers to struct
}

contract CollidingName {

}
```

The mapping will assume it is from `uint` to `contract CollidingName` when it should really be from `uint` to `struct CollidingName`.

The issue is that type is being parsed from the `name` attribute instead of the `type` attribute, which provides less information about the type. In the above case, `name` = `CollidingName` while `type` = `struct A.CollidingName`. (In type_parsing.py, line 163).

Xenomega:
This pull request resolves #160 and resolves #137 . It changes parsing on legacy-ast to use the `type` attribute instead of the `name` attribute, which provides more information, and was likely to be intended to be used from the start.

All relevant changes are in `type_parsing.py`. Please refer to the examples in both mentioned issues for examples of breaking behavior that has been fixed. 

A pass was done on over a thousand contracts to verify integrity of parsing. Further tests are always welcome, as this is a critical change.

### DIFF
--- a/slither/solc_parsing/solidity_types/type_parsing.py
+++ b/slither/solc_parsing/solidity_types/type_parsing.py
@@ -160,7 +160,7 @@ def parse_type(t, caller_context):
     elif t[key] == 'UserDefinedTypeName':
         if is_compact_ast:
             return _find_from_type_name(t['typeDescriptions']['typeString'], contract, contracts, structures, enums)
-        return _find_from_type_name(t['attributes'][key], contract, contracts, structures, enums)
+        return _find_from_type_name(t['attributes']['type'], contract, contracts, structures, enums)
 
     elif t[key] == 'ArrayTypeName':
         length = None
--- a/slither/solc_parsing/solidity_types/type_parsing.py
+++ b/slither/solc_parsing/solidity_types/type_parsing.py
@@ -160,7 +160,10 @@ def parse_type(t, caller_context):
     elif t[key] == 'UserDefinedTypeName':
         if is_compact_ast:
             return _find_from_type_name(t['typeDescriptions']['typeString'], contract, contracts, structures, enums)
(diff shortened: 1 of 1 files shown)


montyly:
The change breaks the support for old solc compilations

For example on https://etherscan.io/address/0xe554056146fad6f2a7b5d5dbbb6f6763d58926c5#code

```
slither contract.sol  --solc solc-0.4.11 --legacy-ast
```
```
line 163, in parse_type
    return _find_from_type_name(t['attributes']['type'], contract, contracts, structures, enums)
KeyError: 'type'
```

slither/solc_parsing/solidity_types/type_parsing.py
```diff
@@ -160,7 +160,7 @@ def parse_type(t, caller_context):
     elif t[key] == 'UserDefinedTypeName':
         if is_compact_ast:
             return _find_from_type_name(t['typeDescriptions']['typeString'], contract, contracts, structures, enums)
-        return _find_from_type_name(t['attributes'][key], contract, contracts, structures, enums)
+        return _find_from_type_name(t['attributes']['type'], contract, contracts, structures, enums)
```
montyly: Maybe we can check if one of the key exists, and switch to the correct one?

PR #166, opened 2019-02-06
Fix shadowing-function detector and enhance inheritance-graph

### LINKED ISSUE #165: shadowing-function detector does not work as expected
The `shadowing-function` detector (used internally for the `inheritance-graph` printer) does not return correct results. It relies on checking that an inheriting contract has a function definition with the same `full_name`, but does this by using the `functions` property of a contract. The `functions` property returns functions from contracts inherited from as well. This causes this detector to return every function for any inherited contract as if it is being shadowed (which the printer later filters out, making it seem as if this detector is behaving as intended).

Example:
```
contract A {
    function notShadowed() public {

    }
}

contract B is A {

}
```
This contract will produce the following results from the detector:
```
[{'vuln': 'ShadowingFunctionContract', 'filename': '.\\tests\\c3_shadowing.sol', 'contractShadower': 'B', 'contract': 'A', 'functions': ['notShadowed()']}]
```

Obviously `notShadowed` is not shadowed and should not be returned as a result. 

Furthermore, this could be extended to include modifiers so that the printer can highlight shadowed modifiers as well.

Xenomega:
This pull request aims to resolve #165 . It fixes the `shadowing-function` detector and extends/

**Changes:**
-Fixed an issue where the `shadowing-function` detector would return all inherited functions (even if not shadowed).
-Added an `internal` directory under `detectors` to house detectors which are not used for vulnerabilities, but as helpers throughout the program (such as for the `inheritance-graph` printer).
-Updated `inheritance-graph` printer to highlight all function collisions (shadowing + shadowed functions, not just shadowing).
-Updated `inheritance-graph` printer to consider modifier collisions (not just functions).
-Updated `inheritance-graph` printer to highlight collisions via C3 linearization (and added a relevant `internal` detector to help with detection of all c3 linearization function collisions).

Note: Unimplemented functions and their later implementations will not count as collisions currently (and will not be highlighted). This can easily be changed if desired.

### DIFF
--- a/slither/detectors/internal/c3_shadowing_internal.py
+++ b/slither/detectors/internal/c3_shadowing_internal.py
@@ -1,0 +1,88 @@
+"""
+Module detecting potential C3 linearization bugs (for internal use by printers)
+"""
+
+from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
+
+
+class C3LinearizationShadowingInternal(AbstractDetector):
+    """
+    C3 Linearization Shadowing
+    """
+
+    ARGUMENT = 'shadowing-c3-internal'
+    HELP = 'C3 Linearization Shadowing (Internal)'
+    IMPACT = DetectorClassification.HIGH
+    CONFIDENCE = DetectorClassification.HIGH
+
+    # This detector is not meant to be called as a generic detector
+    # It's only used by inheritances printers
+    WIKI = 'undefined'
+    WIKI_TITLE = 'undefined'
+    WIKI_DESCRIPTION = 'undefined'
+    WIKI_EXPLOIT_SCENARIO = 'undefined'
+    WIKI_RECOMMENDATION = 'undefined'
+
+
+    def detect_shadowing_definitions(self, contract):
+        """
+        Detects if a contract has two or more underlying contracts it inherits from which could potentially suffer from
+        C3 linearization shadowing bugs.
+
+        :param contract: The contract to check for potential C3 linearization shadowing within.
+        :return: A list of list of tuples: (contract, function), where each inner list describes colliding functions.
+        """
+
+        # Loop through all contracts, and all underlying functions.
+        results = {}
+        for i in range(0, len(contract.immediate_inheritance) - 1):
+            inherited_contract1 = contract.immediate_inheritance[i]
+
+            for function1 in inherited_contract1.functions_and_modifiers:
+                # If this function has already be handled or is unimplemented, we skip it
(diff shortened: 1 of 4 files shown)


montyly:
Some notes:

- After some tests, I am not sure we should highlight the shadowing functions. It makes the use of the printer a bit ambigious (ex: you might think that a function is shadowed, while its the shadowing function). We could use another color, but it's not necessary good to have too many code colors. Another solution is to write the canonical name of the shadowing function next to the shadowed function's name, any thoughts?  

- We could do the same for the state variables. Right now, the derived contracts show all the state variables, while it could only show the new ones (and in red the shadowed ones, as it's probably a bug)

- My first idea of having detectors that are not 'calleable' from the cli command was a bad idea. We can probably just create a `slither.utils.inheritance` module with functions that provide the same info

Xenomega:
For simplicity in distinguishing commits, the following changes have been made up to this point:

- Internal detectors were removed and logic was moved to `slither.utils.inheritance_analysis`. This provides functions to obtain all shadowing/shadowed function pairs. 
    - `detect_direct_function_shadowing(contract)` will obtain all shadowing via direct line of inheritance for a given contract.  
    - `detect_c3_function_shadowing(contract)` will obtain all indirect shadowing through C3 linearization. 
    - `detect_function_shadowing(contracts, direct_shadowing=True, indirect_shadowing=True)` combines the results from each into pairs which more simplistically state what is shadowed, and what is shadowing. This is used to obtain clean results for the inheritance-graph printer.
 - `inheritance-graph` was modified to only highlight shadowed functions.

TODO:
- [ ] Currently `detect_function_shadowing` returns overshadowed pairs for c3 linearization, but it does not take into account if there are multiple overshadowings via c3 that neutralize eachother (overshadow back to the expected function).
- [ ] Add state variable shadowing helpers to `inheritance_analysis`, plug it in to `inheritance-graph`.

Xenomega:
For documentation:

To this point, all TODO's are complete (save for the last). Additionally:
-If a contract has multiple inheritance, edges will be numbered in the order of declaration.
-Fixed a bug where special output formatting for state variables of type "Contract" would never be reached.

Xenomega:
This PR is ready for review now. The final behavior for `inheritance-graph` is as follows:
- Edges are numbered according to order of inheritance.
- Variable shadowing will be shown in red.
- Direct function shadowing will be shown in orange.
- Additional information about indirect shadowing via c3 linearization will be shown in an additional text box at the bottom of any affected node.

Example:
graphviz

Other supporting changes/bugfixes are detailed throughout this thread.

PR #179, opened 2019-02-25
Fix Truffle invocation conflict on Windows

### LINKED ISSUE #173: Truffle compile interop fails on Windows
Slither fails to run `truffle compile` on Windows:

Command: 
```
slither .
```
Output:
```
INFO:Slither:'truffle compile' running (use --truffle-version truffle@x.x.x to use specific version)
ERROR:root:Error in .
ERROR:root:Traceback (most recent call last):
  File "c:\users\vyper\documents\github\slither\slither\__main__.py", line 471, in main_impl
    (results, number_contracts) = process_truffle(filename, args, detector_classes, printer_classes)
  File "c:\users\vyper\documents\github\slither\slither\__main__.py", line 87, in process_truffle
    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "C:\Users\Vyper\AppData\Local\Programs\Python\Python37-32\lib\subprocess.py", line 769, in __init__
    restore_signals, start_new_session)
  File "C:\Users\Vyper\AppData\Local\Programs\Python\Python37-32\lib\subprocess.py", line 1172, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

If `truffle compile` is run manually, and the `truffle compile` code is disabled in slither, analysis will succeed, so it can probably be assumed that the `subprocess.Popen` command is only blocking issue.

NOTE: `truffle` is in the Windows path.

Xenomega:
This pull request aims to resolve #173 . The `truffle compile` command invoked through slither will fail on Windows due to naming conflicts. On Windows, `.JS` is an executable extension by default (see the `PATHEXT` environment variable), which confuses windows into invoking `truffle.js` instead of `truffle.cmd`. 

Truffle's own website suggests a few fixes:
https://truffleframework.com/docs/truffle/reference/configuration#resolving-naming-conflicts-on-windows

The only fix which does not require the user modifying their environment is to explicitly invoke `truffle.cmd`. This seems simplistic and non-problematic, so I have added a switch to `__main__.py` which invokes `truffle.cmd` if on Windows, otherwise simply `truffle` is invoked.

This has fixed truffle invocation on my Windows system.

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -6,6 +6,7 @@ import inspect
 import json
 import logging
 import os
+import platform
 import subprocess
 import sys
 import traceback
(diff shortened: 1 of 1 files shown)


montyly:
As we are fixing Windows support: if a truffle version is specified in `package.json`, slither will run `npx truffle@x.x.x compile`, does that work correctly on Windows?

Issue #192, opened 2019-03-22
Return values from functions are not propagated to call sites for data dependency


rajeevgopalakrishna:
Return values of functions are not propagated to call sites for data dependency. For example, in the below contract:
```
contract PropagateThroughReturnValue {
  uint var_dependant;
  uint var_state;

  function foo() public {
    var_dependant = bar();
  }

  function bar() internal returns (uint) {
    return (var_state);
  }
}

```
`bar` returns `var_state` and therefore should make `var_dependent` depend on `var_state` because of the call site `var_dependant = bar();` in `foo`. But the `data_dependency` printer output shows that it is not:

```
Contract PropagateThroughReturnValue
+---------------+---------------+
|    Variable   |  Dependencies |
+---------------+---------------+
| var_dependant |       []      |
|   var_state   | ['var_state'] |
+---------------+---------------+
```

PR #193, opened 2019-03-22
 Fixed data dependency to propagate return values of functions to call sites

rajeevgopalakrishna:
This PR addresses https://github.com/crytic/slither/issues/192 to fix data dependency to propagate return values of functions to call sites.

With this fix, for the below contract:
```
contract PropagateThroughReturnValue {
  uint var_dependant;
  uint var_state;

  function foo() public {
    var_dependant = bar();
  }

  function bar() internal returns (uint) {
    return (var_state);
  }
}
```

the `data_dependency` printer output below shows that `var_dependent` in `foo` depends on `var_state` returned from `bar` because of the call site `var_dependant = bar();` in `foo`:

```
Contract PropagateThroughReturnValue
+---------------+---------------+
|    Variable   |  Dependencies |
+---------------+---------------+
| var_dependant | ['var_state'] |
|   var_state   | ['var_state'] |
+---------------+---------------+
```

### DIFF
--- a/slither/analyses/data_dependency/data_dependency.py
+++ b/slither/analyses/data_dependency/data_dependency.py
@@ -4,7 +4,7 @@
 from slither.core.declarations import (Contract, Enum, Function,
                                        SolidityFunction, SolidityVariable,
                                        SolidityVariableComposed, Structure)
-from slither.slithir.operations import Index, OperationWithLValue
+from slither.slithir.operations import Index, OperationWithLValue, Return, InternalCall
 from slither.slithir.variables import (Constant, LocalIRVariable,
                                        ReferenceVariable, ReferenceVariableSSA,
                                        StateIRVariable, TemporaryVariable,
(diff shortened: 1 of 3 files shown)


slither/analyses/data_dependency/data_dependency.py
```diff
@@ -161,11 +162,27 @@ def pprint_dependency(context):
...
+    for contract in slither.contracts:
+        for function in contract.all_functions_called:
+            found_return = False
+            if (not found_return):
+                for node in function.nodes:
+                    if (node.type == NodeType.RETURN):
+                        found_return = True
+                        for ir in node.irs_ssa:
+                            if isinstance(ir, Return):
+                                return_values[function] = ir.values
```
montyly: We should use a `collections.defaultdict` and the operator `+=`, otherwise function with multiple return statement will not work

rajeevgopalakrishna: Sorry, my bad. Multiple returns is an obvious bug-will fix.

Will also look into memoization of return values.

rajeevgopalakrishna:
As suggested above, moved the capturing of return values (with memoization) to properties in `slither/core/declarations/function.py` from `slither/analyses/data_dependency/data_dependency.py`. Fixed the bug where multiple returns within a function would be missed.

PR #194, opened 2019-03-29
Bug fix for line number calculations

### LINKED ISSUE #180: Source mapping for lines are not always correct
The line numbers for issues reported by slither often don't line up to the source lines indicated. I don't have any public examples but it should be relatively easy to reproduce. I noticed when reviewing a contract with a lot of `Conformance to Solidity naming conventions` issues.

Xenomega:
This issue aims to resolve #180 . By default, python's `open()` function parses CR LF (`\r\n`) line endings as LF (`\n`). During line number calculation, a byte offset is converted into line numbers by splitting the string with a `\n` character, and summing up the length of each line until the correct line byte range is found. With the removal of CR control characters, a miscalculation of line lengths would cause a later-than-intended line range to be reported.

This fixes the issue by adding a `newline=''` argument to the `open()` functions related to source mapping. The use of a blank new line will allow python's new line related parsing functions to work, while retaining CR LF line endings. Reference: https://stackoverflow.com/questions/20350305/python-read-crlf-text-file-as-is-with-crlf

NOTE: This means that the `source_code` string property for a source mapping will now retain CR characters within them. This property is not currently used for anything but line number calculations, making it non-problematic. In the future, if code intends to print/operate on these source code properties, it should be mindful that line endings can be CR LF or LF.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ setup(
     description='Slither is a Solidity static analysis framework written in Python 3.',
     url='https://github.com/trailofbits/slither',
     author='Trail of Bits',
-    version='0.6.0',
+    version='0.6.1',
     packages=find_packages(),
     python_requires='>=3.6',
     install_requires=['prettytable>=0.7.2', 'pysha3>=1.0.2'],
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -450,6 +450,11 @@ def parse_args(detector_classes, printer_classes):
                         action='store_true',
                         default=defaults_flag_in_config['legacy_ast'])
 
+    parser.add_argument('--ignore-return-value',
+                        help=argparse.SUPPRESS,
+                        action='store_true',
+                        default=False)
+
     # if the json is splitted in different files
     parser.add_argument('--splitted',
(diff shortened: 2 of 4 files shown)


disconnect3d:
For clarity, I believe this:
> During line number calculation, a byte offset is converted into line numbers by splitting the string with a \n character, and summing up the length of each line until the correct line byte range is found.

Is implemented in https://github.com/crytic/slither/blob/82d4730f125c1101c923e4121c1f0f0c74027f67/slither/core/source_mapping/source_mapping.py#L15-L33

Btw. this should probably use `.splitlines()`; also note that there is  `os.linesep` that is cross platform but I doubt it is helpful here.

Xenomega:
@disconnect3d Yes, that was the line calculation code which caused the miscount due to CR LF being parsed as LF line endings. Thanks for the suggestion too, I've updated the code to use `splitlines()`. Although it splits on some additional characters which IDEs do not consider as new lines, Solidity disallows them as `ILLEGAL` characters, which should make it safe to use. That should also fix up another edge case with sole CR characters.

PR #201, opened 2019-04-12
Add right-to-left-override character (U+202E) detection

shshzi:
As can be seen in https://youtu.be/P_Mtd5Fc_3E?t=1813


montyly:
Hi @shshzi 

Thank you for your contribution. It's a nice finding!

PR #202, opened 2019-04-18
Similarity code detection tool

gustavo-grieco:
This PR implements a new util, `slither-simil`, to detect similar Solidity functions in a large amount of smart contracts using Machine Learning. More information and documentation is available [here](https://github.com/crytic/slither/wiki/Code-Similarity-detector).


utils/similarity/cache.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
```
montyly: Consider 
```
try:
  import numpy as np
except ImportError:
  ## print "please install numpy: pip ...
```

utils/similarity/encode.py
```diff
@@ -0,0 +1,188 @@
+import os
+import sys
+
+from slither import Slither
+from slither.slithir.operations import *
```
montyly: Consider using explicit import

utils/similarity/encode.py
```diff
@@ -0,0 +1,188 @@
...
+        slither = Slither(filename, solc=solc)
+    except:
+        logger.error("Compilation failed")
+        return r
+
+    # Iterate over all the contracts
+    for contract in slither.contracts:
+
+        # Iterate over all the functions
+        for function in contract.functions:
```
montyly: Consider using `contract.functions_not_inherited` to simplify the upcoming https://github.com/crytic/slither/pull/213
(`function.contract == contract` can then be removed)

utils/similarity/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+import traceback
+import operator
+
+from .info     import info
+from .test     import test
+from .train    import train
+from .plot     import plot
+
+logging.basicConfig()
+logger = logging.getLogger("Slither-simil")
```
montyly: Consider not mixing custom logger and `print` usage. Al the print can probably be changed to `logger.info()`

utils/similarity/encode.py
```diff
@@ -0,0 +1,188 @@
...
+    elif "tuple" in _type:
+        return "tuple"
+    elif "contract" in _type:
+        return "contract"
+    elif "mapping" in _type:
+        return "mapping"
+    else:
+        return _type.replace(" ","_")
+
+def encode_ir(ir):
```
montyly: I would split this into two functions `encode_ir` and `encode_value`. 
SlithIR is flat, so all the `ir.value` , `ir.rvalue` will never return an Operation

utils/similarity/plot.py
```diff
@@ -0,0 +1,72 @@
...
+    plt = None
+    
+from fastText import load_model
+from .cache import load_cache
+
+logger = logging.getLogger("Slither-simil")
+
+def plot(args):
+
+    if decomposition is None or plt is None:
```
montyly: Consider moving this to the try /except block above

utils/similarity/cache.py
montyly: Note: import numpy is present in other files

utils/similarity/__init__.py
```diff
@@ -0,0 +1,12 @@
+# from https://stackoverflow.com/questions/563022/whats-python-good-practice-for-importing-and-offering-optional-features
+import sys
+
+try:
+    import fastText
+except ImportError:
+    fastText = None
+
+if fastText is None:
```
montyly: Consider moving this into the try /except block

montyly: It could be worth to include this try except in all the import fastText from the other modules

montyly (commented):
The code relies heavily on large 
```
try
...
except Exception
```
Would it be possible to remove and catch properly the exceptions, or they are too many of them?

utils/similarity/cache.py
ggrieco-tob: fastText is a requirement for this util, and it checked in the initialization:

https://github.com/crytic/slither/blob/dev-simil/utils/similarity/__init__.py#L4-L12

And numpy is a requirement of fastText, so adding that try/except will have no effect. I can still do it if you want to keep the consistency.

utils/similarity/encode.py
ggrieco-tob: There is a lot of things to import: https://github.com/crytic/slither/blob/dev-simil/slither/slithir/operations/__init__.py 

:sob:

utils/similarity/plot.py
ggrieco-tob: If I do that, `matplotlib` won't be an optional dependency anymore. So `slither-simil` will refuse to run without `matplotlib` regardless of the user using the plot mode or not.

utils/similarity/__main__.py
ggrieco-tob: It should be fixed now.

utils/similarity/__init__.py
ggrieco-tob: > Consider moving this into the try /except block

Done.

> It could be worth to include this try except in all the import fastText from the other modules

That shouldn't be necessary. This will execute during the initialization.

utils/similarity/encode.py
ggrieco-tob: During training this could be useful to avoid repetition in the dataset, but I'm not sure this is what we want during testing. In that case, the user cannot specify a inherited function (since it will be missing). :confused:

utils/similarity/cache.py
ggrieco-tob: I finally added this try/catch, in case of the user decides to import this module individually (which is very unlikely, but it could happen)

ggrieco-tob:
> The code relies heavily on large
> 
> ```
> try
> ...
> except Exception
> ```
> 
> Would it be possible to remove and catch properly the exceptions, or they are too many of them?

I removed that. It should be fixed now.

PR #208, opened 2019-04-25
Re-add --solc-ast

### LINKED ISSUE #207: Slither crashes due to removed --solc-ast argument
The release of `0.6.3` moved compilation from the `slither` codebase into `crytic-compile`. With this, some compilation related arguments were removed/changed (`--disable-solc-warning`, `--solc-ast`, etc).

As a result of `--solc-ast` being removed, compilation on a standard directory (non-truffle, etc) will result in the following error:
```
ERROR:root:Traceback (most recent call last):
  File "c:\users\vyper\documents\github\slither\slither\__main__.py", line 512, in main_impl
    extension = "*.sol" if not args.solc_ast else "*.json"
AttributeError: 'Namespace' object has no attribute 'solc_ast'
```

The relevant code which depended upon the `--solc-ast` argument must be modified.

montyly:
Fix #207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ env:
   - TEST_SUITE=scripts/travis_test_etherscan.sh
   - TEST_SUITE=scripts/travis_test_dapp.sh
   - TEST_SUITE=scripts/travis_test_etherlime.sh
+  - TEST_SUITE=scripts/travis_test_cli.sh
+  - TEST_SUITE=scripts/travis_test_printers.sh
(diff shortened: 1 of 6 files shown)


montyly:
In addition, this PR adds 3 new types of test for travis:
- tests_cli: tests some of the cli flags
- test_pritners: call all the printers
- test_slither_config: test the --config flag

And add:
- "all" can be selected for printers (all the printers are then run)
- 'ignore_return_value' can be set in the config file

PR #215, opened 2019-04-26
Fix FP in erc20-interface as a result of ERC721 similarities

### LINKED ISSUE #186: erc20-interface: remove FP in case of other ERC implementations
The `erc20-interface` detector will return FP if an ERC closes to ERC20 is implemented.

We should identify what ERCs are affected and add their support.

Xenomega:
This PR aims to resolve #186 . It checks if any ERC721 functions exist before assuming a contract is implementing ERC20. This is due to the fact that ERC721 has similar definitions to ERC20.

More information about the implementation of this PR can be found in the issue thread.


Xenomega:
As of commit https://github.com/crytic/slither/pull/215/commits/7238e9b0337aa002b348d2837a6a2eb93d0be74b :
- Added an `erc721-interface` detector, which is similar to `erc20-interface`, but checks the properties according to an ERC721 token. 
- `erc20-interface` now checks return types of more functions than it did before.
- `erc20-interface` used to only report poorly formed functions _declared in the immediate contract_ (not inherited). This was relaxed, as parts of the interface might be provided in other base classes which would not pass checks identifying it as an ERC20 token on its own, and these functions would not be reported about. 
  - A good example of this is seen with ERC721: If the sole ERC165 function is provided through a seperate contract which ERC721 depends on, the ERC721 detector would note the ERC721 contract as a contract to check the functions of, but not ERC165. This would cause the ERC165 function checks to be skipped in the detector. The same can happen in ERC20 if parts are provided through different contracts that wouldn't individually meet the criteria to be scanned.

PR #216, opened 2019-04-28
Dev too many digits

GillesdeB:
The goal is to add a new Detector that finds patterns of literal values in numbers that are too big for humans to safely manage, issue a warning, and advise to use the proper denominations of Ether. For example:
       uint x = 0x0000000000001;
       uint y = 1000000000000000000;
       balance += x + y + 1000000;
would be detected as too big but 100 wei or 1000 szabo would not. My limit so far is anything with more than “00000” (5 zeros) should use another denomination of Ether.

I received help from Josselin Feist and heavily use his code example.


montyly:
Thank you @GillesdeB for your contribution!

The PR looks good, we are going to run it on a large scale benchmark to see what kind of results give the heuristic

montyly:
Hi @GillesdeB, the detector is merged in `dev` and will be available on the upcoming release.

Thanks for your work!

GillesdeB:
Thank you very much @montyly for your help making this happen.
I look forward to downloading and testing the new version.

PR #222, opened 2019-05-02
Added keyword for json output to stdout

### LINKED ISSUE #221: Create argument to output json via stdout
Currently, slither's `--json` argument takes a file path to output the JSON results to. It would be nice if there was an additional command which would output the JSON over standard output (stdout), such that JSON results can be passed between programs without the need for file I/O.

This command should wrap the typical JSON output in another JSON object which specifies if the operation succeeded or not. In the event of a failure, any errors should be output under an additional field as well. This way, external tooling will have a clean and uniform way to know if the operation succeeded, or why it had not.

Xenomega:
This pull request aims to resolve #221 . It adds a switch which checks if the `--json` path is `-`. If it is, it will output the JSON to stdout.

Due to the way that slither currently handles errors, slither may print an error with a logger and call `exit(code)` at various points throughout the codebase, and never return the `main_impl`, where `--json` is called. As a result, this PR takes a naive approach of setting logging levels to `ERROR` for the various main loggers set up in `main_impl`. This way, all output except errors will be silenced, leaving only the JSON to be output cleanly. In the event of an error, it will be output as normal to stdout. This means external tooling should attempt to parse the response, if it succeeds, the user can continue. If it fails, it means the data received over stdout is actually an error.

A more comprehensive approach for the future would likely call for changing the execution flow in `main_impl`, and wrapping of loggers to instead capture the output in case of an error (or raise exceptions instead of logging errors, and bubble them back up to `main_impl` so they can be caught). This would instead offer the ability to encapsulate all output in json (as mentioned in the issue), through use of a `success` boolean field, an `exception` string field, and the underlying `results` json output. External tooling could then follow the same form of parsing throughout, regardless of exceptions or success, and more cleanly determine status.


Xenomega:
Updated to make use of newly merged exception handling scheme. Disables loggers globally and encapsulates JSON results in the `results` field of JSON of the form:
If error: `{ success: false, error: "error message here if success == false", results: [] }`
If success: `{ success: true, error: null, results: <ACTUAL RESULT ARRAY> }`

Issue #224, opened 2019-05-04
[Question] Why unused return only applicable to HighLevelCall?


yxliang01:
In my opinion, the rule `unused-return` should also be applicable to low level calls, I am wondering why it's now only limited to detect for high level calls? Since there's an explicit `if` to skip low level calls, there might be some undocumented intention?

Thanks

CC related : @montyly @benstew

montyly:
Hi @yxliang01, it's a good question, thank you for bringing this up!

After some thinking, I think we should:
1 - Add `InternalCall` and `InternalDynamicCall` to the unused return detector
2 - Create a separate detector for unused returns of low-level calls
3 - Create a separate detector for the lack of check in case of `send` usage

The reason for splitting this into multiple detectors is that it can make sense to NOT check the return value of a low-level call, or a `send`, to prevent the contract to be blocked. Typically, if a contract needs to send ethers to an untrusted destination, using `send` without checking its return value might be better than using `transfer`. 

If we have multiple detectors, the user can decide to disable the detectors related to low-level calls/send, while keeping the high-level/internal calls check.

yxliang01:
@montyly Hmm your reasoning makes sense. :) Happy to see that in near future! As such, maybe you can also process https://github.com/crytic/slither/pull/225 ?

montyly:
Thank you for your suggestions @yxliang01! 
I merged the changes in `dev`. The new detectors will be available in the upcoming 0.6.4 release (in a few days)

PR #225, opened 2019-05-04
fix(detectors/operations/unused_return_values.py): detection incomplete set of operations

yxliang01:
The unused return problem exists not only for `HighLevelCall` but also `LowLevelCall` and `Send`. This commit adds checks on the other two previously ignored operations.

This fixes https://github.com/crytic/slither/issues/224 (if this PR makes sense).


montyly:
Replaced by #230

PR #226, opened 2019-05-06
JSON output improvements

Xenomega:
This pull request provides light improvements towards detector JSON output. It allows additional fields to be specified in slither results and result elements in order to alleviate the need to create custom JSON which may not conform to general detector output spec.

Changes:
- Added JSON output methods for `struct`/`enum`/`event`.
	- Updated `shadowing-builtin` and `shadowing-local` to output as type `event` (previously output events as `function`).
- Added JSON output methods for `other`: types which are not supported by other mappings, for which you can simply supply a string `name`, and `source_mapping` for. These will have the `type` field as `other`.
	- Used to update `solc-version` and `pragma` detectors which were outputting custom json to map to a pragma statement.
- Added JSON output method for individual nodes
- Standardized node output: Typical element JSON output would have (`type`, `name`, `source_mapping`), but `add_nodes_to_json` would output (`type`, `expression`, `source_mapping`). `expression` was replaced with `name` to standardize output.
- Added optional `additional_fields` field in json formatting functions to allow detectors to specify detector-specific data per element or result.
	- Used to standardize output in `naming-convention` detector, which previously did not output a `type`, but output custom data, causing issues in external tooling.

These additional detectors which previously added custom fields now use standardized json methods:
- [x] `complex_function.py`
- [x] `reentrancy_benign.py`
- [x] `reentrancy_read_before_write.py`
- [x] `incorrect_solc.py`
- [x] `const_functions.py`
- [x] `constant_pragma.py`
- [x] `reentrancy_eth.py`


Xenomega:
In addition to the detectors above which need cleanup, the following detectors group findings into a single result, and should be broken up to report a single finding per result:

- [ ] constable-states
- [ ] empty-functions
- [ ] erc20-indexed
- [ ] incorrect-inequality
- [ ] pragma
- [ ] solc-version
- [ ] unused-state

Xenomega:
At this point all intended detectors have been refactored. Additionally, as of https://github.com/crytic/slither/pull/226/commits/19271a071016de8bff9f36ba2cbb8022734ba8ee, the `type` for node elements was changed from `expression` to `node`, and now includes data regarding the parent function under a new field called `function`.

PR #234, opened 2019-05-10
Adds contract/function context to the different program constructs in naming-convention detector.

rajeevgopalakrishna:
This PR adds contract/function context to the different program constructs in `naming-convention` detector.

Issue #218, opened 2019-04-29
Line/column number calculation regression


Xenomega:
Recent adoption of `crytic-compile` in slither have required some code paths to be changed. A regression can be observed in source mapping line/column calculations due to CR, LF, and CRLF line endings, as described in #194 .

The problematic code can be found here:
https://github.com/crytic/slither/blob/04c147f7e50223c6af458ca430befae747ccd259/slither/slither.py#L68

The same solution described in the previous PR should be taken for all source file reading operations: retain format and line endings.

This issue currently will cause incorrect line/column number calculations in files with CRLF line endings.

Issue #150, opened 2019-01-23
slither-format: automatic code improvements


montyly:
We can create a tool similar to `clang-format`, that takes a Solidity file and uses slither detectors to improve the code, in terms of readability and optimization.

We can automatically fix:
- `unused-state`
- `solc-version`
- `pragma`
- `naming-convention`
- `external-function`
- `constable-states`
- `constant-function` 

`slither-format` could produce a Solidity file, or a git patch.

samparsky:
@montly I am interested in this

rajeevgopalakrishna:
I suppose we can close this, correct @montyly ?

montyly:
yes!

PR #235, opened 2019-05-10
[WIP] Tool slither-format: automatic code improvements

rajeevgopalakrishna:
This PR (depends on #233 and #234) addresses #150 to add a new utility tool slither-format which uses slither detectors to identify code patterns of concern (w.r.t security, readability and optimisation) and automatically fix those code patterns with suggested changes.

The current list of detectors used to detect and fix include: `unused-state`, `solc-version`, `pragma`, `naming-convention`, `external-function`, `constable-states` and `constant-function`.

Detectors highlight names, context and source-mapping of code constructs which are then used by slither-format to programmatically locate those constructs in the Solidity files and then replace them with changes based on best practices. Lexical analysis for identification of such constructs is confined to the smallest possible region to avoid conflicts with similarly named constructs (with potentially different types or signatures) in other scopes, functions or contracts within the same file (because of shadowing, overloading etc.).

_Done:_ Framework and basic functionality. Unit testing.
_WIP:_ Real-world testing. More tool options. Merging with recent format changes in detector JSON output (#226).
_Known bugs:_
1. `naming-convention` formatting doesn't work on NatSpec comments, e.g. @param.
2. Variables used as indices on LHS (e.g. `_to` in `balances[_to] = 100`) don't get reported from Slither in variables that are read and hence are not formatted in `naming-convention`.
3. Bugs potentially from incorrectly reported source mappings, which might be related to https://github.com/crytic/slither/issues/218.


rajeevgopalakrishna:
This PR was created against a stale `dev` branch by mistake; hence closed and replaced by #238.

PR #236, opened 2019-05-10
Adds event.full_name for JSON output of events in naming-convention detector

rajeevgopalakrishna:
The recent PR https://github.com/crytic/slither/pull/226 with JSON output improvements makes most earlier changes in #234 irrelevant/stale. The only change missing (`event.full_name`) is in this PR.


rajeevgopalakrishna:
Removed the use of `event.full_name` in naming-convention formatting to simply use `event.name`, realising that using `full_name` (event name + parameter type signatures) doesn't matter here. So this PR #236 is not required.

PR #237, opened 2019-05-10
slither core/parsing changes required for slither-format

rajeevgopalakrishna:
This PR replaces PR #233 which was created against a stale `dev` branch by mistake. Same set of changes as PR #233 required for slither-format tool:

1. Adds _internal_calls_as_expressions to complement _external_calls_as_expressions
2. Adds source_mapping hooks for function parameters and returns
3. Sets source_mapping offsets for function calls, modifier calls and identifier uses

### DIFF
--- a/slither/core/cfg/node.py
+++ b/slither/core/cfg/node.py
@@ -155,6 +155,7 @@ class Node(SourceMapping, ChildFunction):
         self._library_calls = []
         self._low_level_calls = []
         self._external_calls_as_expressions = []
+        self._internal_calls_as_expressions = []
         self._irs = []
         self._irs_ssa = []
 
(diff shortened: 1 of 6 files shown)


montyly:
2- I think we should remove all the `get_source_*_declaration` from contract and function, in favor of `get_local_variable`, etc type of function. These functions are not used anymore and are a bit too specific. Any thoughts?

3 - for `parameters_src` and `return_src` should not be raw Solidity source mapping text. Until now all the source mapping are attached to an object, which inherits from [`SourceMapping`](https://github.com/crytic/slither/blob/master/slither/core/source_mapping/source_mapping.py). I am not sure how to handle them. Maybe we can create a source mapping object for them? An alternative is to just use the source mapping from the parameter variable / return node, but I dont know if that works for your case.

rajeevgopalakrishna:
> 2- I think we should remove all the get_source_*_declaration from contract and function, in favor of get_local_variable, etc type of function. These functions are not used anymore and are a bit too specific. Any thoughts?

Agree. Checked that they're not used elsewhere within `slither`. Changed `slither-format` to directly use `source_mapping` on variables and events. Removed these functions from `slither`.

rajeevgopalakrishna:
> 3 - for parameters_src and return_src should not be raw Solidity source mapping text. Until now all the source mapping are attached to an object, which inherits from SourceMapping. I am not sure how to handle them. Maybe we can create a source mapping object for them? An alternative is to just use the source mapping from the parameter variable / return node, but I dont know if that works for your case.

Changed `parameters_src` and `returns_src` to `SourceMapping` objects instead of the earlier raw source text.

PR #238, opened 2019-05-10
[WIP] Tool slither-format: automatic code improvements

rajeevgopalakrishna:
This PR (depends on #237 ~~and #236~~ and replaces PR #235 which was created against a stale `dev` branch by mistake) addresses #150 to add a new utility tool slither-format which uses slither detectors to identify code patterns of concern (w.r.t security, readability and optimisation) and automatically fix those code patterns with suggested changes.

The current list of detectors used to detect and fix include: `unused-state`, `solc-version`, `pragma`, `naming-convention`, `external-function`, `constable-states` and `constant-function`.

Detectors highlight names, context and source-mapping of code constructs which are then used by slither-format to programmatically locate those constructs in the Solidity files and then replace them with changes based on best practices. Lexical analysis for identification of such constructs is confined to the smallest possible region to avoid conflicts with similarly named constructs (with potentially different types or signatures) in other scopes, functions or contracts within the same file (because of shadowing, overloading etc.).

_Done:_ Framework and basic functionality. Unit testing.
_WIP:_ More real-world testing and bug-fixes. More tool options.
_Known bugs:_
1. naming-convention formatting doesn't work on NatSpec comments, e.g. @param.
2. Variables used as indices on LHS (e.g. _to in balances[_to] = 100) don't get reported from Slither in variables that are read and hence are not formatted in naming-convention.
~~3. Bugs potentially from incorrectly reported source mappings, which might be related to #218.~~


montyly:
Few recommendations:
- Consider using the cli crytic_compile options, for example:
https://github.com/crytic/slither/blob/7cadf980b83ecbb31456c3502c1a8508ce811bf9/utils/upgradeability/__main__.py#L6
https://github.com/crytic/slither/blob/7cadf980b83ecbb31456c3502c1a8508ce811bf9/utils/upgradeability/__main__.py#L31-L38
As a result,
https://github.com/crytic/slither/blob/2e930765280cba818b37acc5bc9890551cb2e5c2/utils/slither_format/__main__.py#L54
Would be become `slither = Slither(args.filename,  **vars(args)) `, and this flag can be removed:
https://github.com/crytic/slither/blob/2e930765280cba818b37acc5bc9890551cb2e5c2/utils/slither_format/__main__.py#L30

- Consider adding `slither-format` to [`setup.py`](https://github.com/crytic/slither/blob/7cadf980b83ecbb31456c3502c1a8508ce811bf9/setup.py#L19), so we can call `slither-format` directly from the command line

rajeevgopalakrishna:
> Few recommendations:
> * Consider using the cli crytic_compile options, for example:

Done.

utils/slither_format/format_constable_states.py
```diff
@@ -0,0 +1,30 @@
...
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('Slither.Format')
+set_colorization_enabled(True)
+
+class FormatConstableStates:
+
+    @staticmethod
+    def format(slither, patches, elements):
+        for element in elements:
+            FormatConstableStates.create_patch(slither, patches, element['source_mapping']['filename_absolute'], element['name'], "constant " + element['name'], element['source_mapping']['start'], element['source_mapping']['start'] + element['source_mapping']['length'])
```
montyly: Consider using multiple lines declaration (this comment is true for most of the files of the PR).
We should follow as best as we can the [pep8 max legnth](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

utils/slither_format/format_external_function.py
```diff
@@ -0,0 +1,55 @@
...
+        for element in elements:
+            Found = False
+            for contract in slither.contracts:
+                if not Found and contract.name == element['type_specific_fields']['parent']['name']:
+                    for function in contract.functions:
+                        if function.name == element['name']:
+                            # If function parameters are written to in function body then we cannot convert this function
+                            # to external because external function parameters are allocated in calldata region which is
+                            # non-modifiable. See https://solidity.readthedocs.io/en/develop/types.html#data-location
+                            if not FormatExternalFunction.function_parameters_written(function):
```
montyly: This should be removed as we have now https://github.com/crytic/slither/pull/257

utils/slither_format/format_external_function.py
```diff
@@ -0,0 +1,55 @@
...
+                                FormatExternalFunction.create_patch(slither, patches, element['source_mapping']['filename_absolute'], "public", "external", int(function.parameters_src.source_mapping['start']), int(function.returns_src.source_mapping['start']))
+                                Found = True
+                                break
+
+    @staticmethod
+    def create_patch(slither, patches, in_file, match_text, replace_text, modify_loc_start, modify_loc_end):
+        in_file_str = slither.source_code[in_file]
+        old_str_of_interest = in_file_str[modify_loc_start:modify_loc_end]
+        m = re.search(" public", old_str_of_interest)
```
montyly: Consider using `" public "`, to prevent collision with the name of the function

utils/slither_format/format_naming_convention.py
```diff
@@ -0,0 +1,600 @@
...
+from slither.core.cfg.node import Node
+from slither.slithir.operations import NewContract
+from slither.slithir.operations import Member
+from slither.utils.colors import red, yellow, set_colorization_enabled
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('Slither.Format')
+set_colorization_enabled(True)
+
+class FormatNamingConvention:
```
montyly: Consider adding code documentation around the string manipulation and the use of regex

utils/slither_format/format_naming_convention.py
```diff
@@ -0,0 +1,600 @@
...
+                            }
+                            if not patch in patches[in_file]:
+                                patches[in_file].append(patch)
+                        else:
+                            logger.error(red("Could not find modifier?!"))
+                            sys.exit(-1)
+                            
+    @staticmethod
+    def create_patch_modifier_uses(slither, patches, name, contract_name, in_file):
+        for contract in slither.contracts:
```
montyly: Consider using `slither.get_contract_from_name(contract_name)`

`get_contract_from_name` can be probably also be used in other parts of the PR

It's going to be easier to maintain. The only caveat is that it can return a `None`, but I suppose that it would be a a bug in this context?

utils/slither_format/format_naming_convention.py
```diff
@@ -0,0 +1,600 @@
...
+            if (contract.name == contract_name):
+                target_contract = contract
+        for contract in slither.contracts:
```
montyly: Consider removing the loop on all the contracts by
`for contract  in [target_contract] + target_contract.derived_contracts`

utils/slither_format/format_naming_convention.py
rajeevgopalakrishna: Ah, didn't know about this function. Done.

utils/slither_format/format_naming_convention.py
rajeevgopalakrishna: Done.

utils/slither_format/format_external_function.py
rajeevgopalakrishna: Will remove once #257 is merged on `dev` else it fails unit tests.

utils/slither_format/format_external_function.py
rajeevgopalakrishna: Made the regex stricter. The prefix collision could be with the modifier if present.

montyly:
I made significant changes, among others:
- create a module to store formatters [`slither_format.formatters`](https://github.com/crytic/slither/tree/ed239cf970eb062d5494feb632585aa1103833d7/utils/slither_format/formatters)
- use functions rather than class for the formatters (all the methods were staticmethod)
- refactor several pieces of code, including:
     -  all the formatter calls `utils.patches.create_patch` to generate the patches, so we ensure the consistency of the dictionary generated. It also simplifies the add of the new formatters, the rules are: create a `format` function, and call `create_patch` for any new patch
     - the loop in naming convention are simplified (ex: for the contracts lookup)
     - naming convention uses a namedtuple for the parameters, to improve the code readability and avoid incorrect parameters order

The PR still needs improvements:
- all the unit tests are not passing. They already were not working on f2b89aaa0b2ad9eade1d3ab2aa87a73c847457a7 for me
- I think we should favor a custom patches generation rather than calling `diff`, otherwise, it will not work for Windows users 
https://github.com/crytic/slither/blob/ed239cf970eb062d5494feb632585aa1103833d7/utils/slither_format/slither_format.py#L118
- I am not sure about the sorting of patches. I tend to prefer a simplest solution, where we generate one patch per issue. A patch fixing all the issues can be an option, but probably not the default behavior

montyly:
What was done:
- Refactor to remove unnecessary arguments. 
- The patches are generated directly to the detector json results, which will simplify the pipeline
- `difflib` is used to generate the patch, rather than `diff`
- The patches are only merged if they correspond to the same issue. For now, merging all the fixes of different finding to one patch is disabled; I am not sure it's something we want to promote, as its error-prone. We could re-enable later

Additionally, there is one change to the naming convention detector:
- allow private and internal functions to start with `_` 


Once thorough tests are performed, we can merge the PR

PR #266, opened 2019-05-29
Opt-in JSON result types + JSON fields for console log/compilation

### LINKED ISSUE #247: Typical logging output should be included in JSON
It's great that now Slither natively supports outputting JSON format output to STDOUT. However, all the logs are disabled in this mode. I think that it's better to all logs go to STDERR in this mode, so that the STDOUT still only has a clean JSON output and logs are still being printed to terminal without interfering with the STDOUT JSON output.

Thanks

Xenomega:
This pull request aims to resolve #247 . It improves upon the way JSON results are output with the `--json` argument by cleaning up the way results are constructed and including stdout/stderr output in the JSON results.

Changes:
- Introduced `slither/utils/output_capture.py`: Used to redirect/mirror stdout/stderr output descriptors for `print()` statements and `logging` calls (which store stdout/stderr upon initialization and need swapping).
- Output `stdout` and `stderr` fields under the `results` field in JSON output.
- General JSON output cleanup

This is marked as WIP until the following tasks are completed:
- [x] When outputting JSON to a file, standard output should remain outputting to the console, but should also be captured to be output in JSON.
- [x] Including stdout/stderr should have an opt-in or opt-out argument, so that it can be excluded if undesired (bloats output).
- [x] Further testing
- [x] Additional TODOs listed in below comments


Xenomega:
Commit https://github.com/crytic/slither/pull/266/commits/e85848b9988e183800adc460c53c70e0935e0eb6 adds support for an argument `--json-types`, which can be used to specify desired JSON result types when using the `--json` command. If `--json-types` is not specified, all result types will be included.

Current types include:
- `console` - Returns stdout/stderr console output under `results\stdout` and `results\stderr` in the JSON.
- `detectors` - Returns detector results/findings under `results\detectors` in the JSON.
- `detector-types` - Returns the same output as `--list-detectors` under `results\detector-types` in the JSON. `--list-detectors` was kept for compatibility, and because it can be used without performing analysis on any smart contracts.

Additional TODOs for this PR:
- Add another `--json-type` for `ast`, `bytecode`, etc.

slither/__main__.py
```diff
@@ -576,20 +586,46 @@ def main_impl(all_detector_classes, all_printer_classes):
             number_contracts = 0
             results = []
             if args.splitted and args.solc_ast:
-                (results, number_contracts) = process_files(filenames, args, detector_classes, printer_classes)
+                (slither, results, number_contracts) = process_files(filenames, args, detector_classes, printer_classes)
             else:
                 for filename in filenames:
-                    (results_tmp, number_contracts_tmp) = process(filename, args, detector_classes, printer_classes)
+                    (slither, results_tmp, number_contracts_tmp) = process(filename, args, detector_classes, printer_classes)
```
montyly: the `slither` will not be correct here, as it is a loop.
The variable seems to be only used for the `compilation` output. I am not sure about keeping this output, as it's a duplicate of `crytic-compile` standard export format, without following the same json format.

I think we should either, we drop `compilation` here, or we create `crytic_compile.export_json`, that we store into a `compilations` list. Any thought?

Xenomega: I wouldn't drop `compilation` entirely. External tooling such as VSCode might find it useful to have all of this information, and might not be able to compile all of the build configurations which `crytic-compile` can on its own. Additionally, running a separate compilation process outside of slither would eat up too much time. 

I'm fine with aggregating all the compilation information in this loop and using forwarding JSON output from `crytic-compile` into slither JSON output as you had mentioned, as long as we output the same data and don't dump to disk (as disk I/O would be slow).

Xenomega: This PR has been updated to change `compilations` into an array of all the compilations which took place (a globbed filename will many compilations + slither instances to analyze each target). Each compilation uses crytic-compile's standard export format now.

montyly:
Changes:
- rename detectors-types to list-detectors
- Add list-printers
- Change default json to only output detectors

Issue #274, opened 2019-06-11
slither-check-upgradeability raises "IndexError: list index out of range"


EricR:
When contract v1 defines more state variables than contract v2 does, slither-check-upgradeability raises an IndexError. In addition, I think the code can probably be refactored a little bit and log output made clearer.

PR #275, opened 2019-06-11
General improvements for utils/upgradeability/compare_variables_order.py

EricR:
- Fixed https://github.com/crytic/slither/issues/274
- Refactored code
- Made logging more explicit and (hopefully) clearer

### DIFF
--- a/tests/check-upgradeability/test_1.txt
+++ b/tests/check-upgradeability/test_1.txt
@@ -2,6 +2,6 @@ INFO:CheckInitialization:[92mRun initialization checks... (see https://github.c
 INFO:CheckInitialization:[93mInitializable contract not found, the contract does not follow a standard initalization schema.[0m
 INFO:CompareFunctions:[92mRun function ids checks... (see https://github.com/crytic/slither/wiki/Upgradeability-Checks#functions-ids-checks)[0m
 INFO:CompareFunctions:[92mNo function ids collision found[0m
-INFO:VariablesOrder:[92mRun variable order checks between the proxy and implementation... (see https://github.com/crytic/slither/wiki/Upgradeability-Checks#variables-order-checks)[0m
-INFO:VariablesOrder:[93mStorage variable defined in proxy Proxy: address destination[0m
+INFO:VariablesOrder:[92mRun variable ordering checks between the proxy and implementation... (see https://github.com/crytic/slither/wiki/Upgradeability-Checks#variables-order-checks)[0m
+INFO:VariablesOrder:[93mstate variable defined in proxy Proxy: address destination[0m
 INFO:VariablesOrder:[92mNo variable ordering issues found between proxy and implementation[0m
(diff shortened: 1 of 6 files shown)


montyly:
Nice work!

So the loop now in `compare_variables_order_proxy` will first print out all the variable in the proxy, in the previous version, it was printing all the variables, but with a warning when something was wrong.

Typically, we would have:
```
- var1
- var2 -> warning something
- var3
```
Now we have
```
- var1
- var2 
- var3

- var2 -> warning something
```

I think the previous order was a bit better, as we reduce the size of the output. The main issue though that the info was not clear enough.

Maybe we should keep the original logic, but refactor the output to be more clear?

EricR:
> Nice work!
> 
> So the loop now in `compare_variables_order_proxy` will first print out all the variable in the proxy, in the previous version, it was printing all the variables, but with a warning when something was wrong.
> 
> Typically, we would have:
> 
> ```
> - var1
> - var2 -> warning something
> - var3
> ```
> 
> Now we have
> 
> ```
> - var1
> - var2 
> - var3
> 
> - var2 -> warning something
> ```
> 
> I think the previous order was a bit better, as we reduce the size of the output. The main issue though that the info was not clear enough.
> 
> Maybe we should keep the original logic, but refactor the output to be more clear?

Good point. I agree the older format was better, but the warning could be clearer.

EricR:
> Nice work!
> 
> So the loop now in `compare_variables_order_proxy` will first print out all the variable in the proxy, in the previous version, it was printing all the variables, but with a warning when something was wrong.
> 
> Typically, we would have:
> 
> ```
> - var1
> - var2 -> warning something
> - var3
> ```
> 
> Now we have
> 
> ```
> - var1
> - var2 
> - var3
> 
> - var2 -> warning something
> ```
> 
> I think the previous order was a bit better, as we reduce the size of the output. The main issue though that the info was not clear enough.
> 
> Maybe we should keep the original logic, but refactor the output to be more clear?

I reworked the logic a little bit in an attempt to fix some faulty checks. For example, if there are more variables in version 1 of a contract than version 2, the check will currently report that variables are missing when they may just be out of order:

https://github.com/crytic/slither/blob/1128b4c2008e13c2a49bad37f34e261b35409f97/utils/upgradeability/compare_variables_order.py#L32-L36

Similarly, if version 2 of a contract has more variables than version 1, the check will currently report that variables are new when they may just be out of order:

https://github.com/crytic/slither/blob/1128b4c2008e13c2a49bad37f34e261b35409f97/utils/upgradeability/compare_variables_order.py#L46-L49

I agree that the repeated variables in output isn't great. What if instead we did the following?

```python
    for idx in range(len(vars_v1)):
        (v1_type, v1_name) =  vars_v1[idx]
        if idx > len(vars_v2):
            logger.info(red('Expected variable in {}: {} {}'.format(
                new_name, v1_type, v1_name
            )))
            issues_found = True
            continue

        (v2_name, v2_type) =  vars_v2[idx]
        if (v1_name != v2_name) or (v1_type != v2_type):
            logger.info(red('Expected variable in {}: {} {}. Found {} {} instead.'.format(
                new_name, v1_type, v1_name, v2_type, v2_name
            )))
            issues_found = True
        elif warn:
            logger.info(yellow('Variable in {}: {} {}'.format(new_name, v2_type, v2_name)))
```

This way we aren't warning about something missing, but rather that it isn't where it was expected to be in v2. Then we can just display a list of new variables after the ordering warnings.

montyly:
I am closing this PR. It is out of date, and will be replaced by https://github.com/crytic/slither/pull/354

PR #276, opened 2019-06-12
Detect Aragon OS

EricR:
Detection for Aragon OS and [Aragon App](https://github.com/aragon/aragonOS/blob/dev/contracts/apps/AragonApp.sol).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+# Contributing to Manticore
+First, thanks for your interest in contributing to Slither! We welcome and appreciate all contributions, including bug reports, feature suggestions, tutorials/blog posts, and code improvements.
+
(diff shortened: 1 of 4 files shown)


montyly:
What are there other aragon dapps that are commonly used? 
For example it would make sense to also detect any delegatecall proxy pattern setup (https://github.com/aragon/aragonOS/blob/dev/contracts/common/DelegateProxy.sol)

EricR:
> What are there other aragon dapps that are commonly used?
> For example it would make sense to also detect any delegatecall proxy pattern setup (https://github.com/aragon/aragonOS/blob/dev/contracts/common/DelegateProxy.sol)

Good call, I'll add detection for other common contracts.

EricR:
@montyly I added a bunch of other common contracts from AragonOS.

PR #293, opened 2019-07-03
Translate to SlithIR missing expressions

montyly:
This PR aims to translate to SlithIR the Solidity expressions that are not yet converted, including:
- Modifier calls
- State variable initialization

This PR will need extensive tests prior to being merged


montyly:
What was changed:
- Add a new type of function: constructor_variable to hold state variale initialization
- Add Function.FunctionType to determine if the function is a constructor/fallback/constructor_variables
- API change: function.is_fallback() becomes a property

Issue #286, opened 2019-06-27
Export Printer to JSON


montyly:
# Task Description

We want to add a printer output similar to our [JSON wrapping output](https://github.com/crytic/slither/blob/master/slither/__main__.py#L111-L126) for the detector 

- [ ] Classify the printer per output types (ex: text, dot generation, ...)
- [ ] Severalize the output to be sent through JSON
- [ ] Add the JSON printer output to our standard JSON

# Acceptance criteria
- [ ] Slither JSON output can contain any printer results
- [ ] An update of the [JSON documentation](https://github.com/crytic/slither/wiki/JSON-output) is prepared
- [ ] Unit tests for printers, based on the JSON output are prepared.

# Required Skills
- [ ] Experience with Python 3.6
- [ ] Experience with Solidity smart contracts
- [ ] Experience with Slither

This issue requires to discuss with me to design a proper output format and requires good programming skills.

Feel free to ask questions here, or join our [slack](https://empireslacking.herokuapp.com/) (#ethereum)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/crytic/slither/286/3215).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/crytic/slither/286/3215).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $136,073.47 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/crytic/slither/286/3217).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/crytic/slither/286/3217).
[...]

PR #295, opened 2019-07-13
Add JSON output functionality for printers

Man-Jain:
This changes the code to return the result as output if --json argument
is passed to printers.

Solves https://github.com/crytic/slither/issues/286

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -86,7 +86,7 @@ def _process(slither, detector_classes, printer_classes):
 
         results.extend(detector_results)
 
-    slither.run_printers()  # Currently printers does not return results
+    results.extend(slither.run_printers())
 
     return slither, results, analyzed_contracts_count
 
@@ -429,6 +429,12 @@ def parse_args(detector_classes, printer_classes):
     args = parser.parse_args()
     read_config_file(args)
 
+    json_types_args = args.json_types.split(',')
+    if args.printers_to_run is not None and 'printers' not in json_types_args:
+        args.json_types = json_types_args.append('printers')
+        args.json_types = json_types_args.remove('detectors')
+        args.json_types = ','.join(json_types_args)
+
     args.filter_paths = parse_filter_paths(args)
 
     # Verify our json-type output is valid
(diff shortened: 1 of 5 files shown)


Xenomega:
Hello @Man-Jain , we appreciate your contribution.

This PR is a good start, but misses the mark a bit, so we would like to request changes. I would focus on improving the following:
- JSON results should be aggregated (combined), so that multiple results (detectors, printers, console output) can be returned in one JSON file. All different result types are put under `results`. For a better example of how this works, check out https://github.com/crytic/slither/pull/266, which will be merged shortly. That PR will return different result types under the `results` JSON node, such as `console`, `detectors`, `detector-types`. It should be a good example of how expected output should generally look.
- The JSON output should not simply be a filename pointing to the data output in traditional non-JSON form. That would overlook the point of adding JSON output, since the filenames are already deterministic. You should instead be outputting the important data the printer provides solely in JSON format.

Hopefully that clears things up a bit! Again, we appreciate you working on this issue. :)

Man-Jain:
Hey Thanks for the feedback. It is really helpful. So this will mean that I'll have to put my patch on top of yours.

Xenomega:
No problem! The output for printers should not go to file in this case. Instead you should find a way to represent the data that _would've_ been output in a file if `--json` wasn't specified, in the JSON itself. 

Example: Lets say you wanted JSON output for inheritance graph, you could simply output the DOT file format encoded in the JSON itself. For the non-graph inheritance printer, you could make a JSON node that has a list of subnodes, where each subnode is a contract, and the subnodes under that point to other contracts they derive from.

That's just an example, but the format is up to you to design. It should be intuitive, concise, and provide all the information you can imagine any external tooling might desire.

Man-Jain:
As per the example you have suggested would require me to change almost all of the printers. It would mean that all of the printers have to be adapted to format the output differently in case of --json is given to it. As for the output of printers some print formatted text, some print tables and some output charts. Wouldn't this be a lot of work for one issue to cover.

Man-Jain:
The code now is able to print the output of a printer to json as it does in case of detector. The code is able to detect the type of output and formats the json in format such that external tools can utilize it. I  have also modified two printers as an example of how this will look. The json looks like [this](https://pastebin.com/feMEB1sX) when both these printers are run simultaneously. The format of the printer itself would depend upon how the printer is implemented.

Do we have a need where we would have to run printer and detectors in a single command. Because that might require going deeper into the codebase. As [this](https://github.com/crytic/slither/blob/master/slither/__main__.py#L71) function implies that only one of them can be run at a time. 

Please review and give your feedback.

Another thing I saw that when giving json-types argument `detector-types` it fails. Is it expected?

Man-Jain:
Waiting for your feedback on this commit so that I can work further.

montyly:
I am closing this PR, as it is replaced by https://github.com/crytic/slither/pull/356

Issue #316, opened 2019-08-20
Slither should not complain about naming convention for echidna properties


agroce:
By convention, property checks in echidna are named "echidna_" or "crytic_" -- so all naming convention gripes from Slither are likely to be false positives.

montyly:
Fixed with #317

PR #317, opened 2019-08-20
ignore naming convention restriction on echidna_ and crytic_ functions

agroce:
Addresses #316


slither/detectors/naming_convention/naming_convention.py
```diff
@@ -97,6 +98,8 @@ def _detect(self):
                 if not self.is_mixed_case(func.name):
                     if func.visibility in ['internal', 'private'] and self.is_mixed_case_with_underscore(func.name):
                         continue
+                    if (func.name.find("echidna_") == 0) or (func.name.find("crytic_") == 0):
```
montyly: Should it be `find` or `startswith` ? 

They are unlikely to have collision, but we probably want to enforce the properties to start with these patterns, rather than at arbitrary place?

agroce: Huh, I simply was unaware of startswith as an alternative to find == 0.  Fixing

PR #325, opened 2019-09-04
Added solc-select to installation section of README.md

burrrata:
Banged my head against the wall for hours trying to install older versions of solc to test Slither. Turns out, I could've skipped all that by just using [solc-select](https://github.com/crytic/solc-select). It's an amazing tool. It just works. I love it. Added a few sentences to the README to recommend using solc-select for older contracts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,9 @@ Num | Printer | Description
 
 ## How to install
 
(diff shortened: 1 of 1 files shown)


montyly:
Hi @burrrata. It's a good idea, thank you!

PR #335, opened 2019-09-12
Add ID to result

montyly:
Each json result has an unique Id.

The ID is for now generated as sha3-256(description), but can be improved to simplify the duplicate results triage


montyly:
replaced by https://github.com/crytic/slither/pull/367

Issue #347, opened 2019-10-15
Create json result for slither-check-upgradability


montyly:
Similar to the detectors, slither upgradability checks could return a standard json with the results

montyly:
Partially done with https://github.com/crytic/slither/pull/348. The documentation is https://github.com/crytic/slither/wiki/JSON-output#slither-check-upgradeability

What is missing ts the source mapping of the element

PR #348, opened 2019-10-17
Add JSON output for slither-check-upgradability tool

rajeevgopalakrishna:
This PR addresses #347 to add JSON output support for `slither-check-upgradability` tool as per the format specified in https://github.com/crytic/slither/wiki/JSON-output.

For now, it adds only the textual information which was being logged. Next step is to add source mapping and any more contextual information, as required.


rajeevgopalakrishna:
Done. Ready for final review and merge.

montyly:
What I did:
- Fix minor bug: the json output was always enabled
- Move `output_json` to a shared module, so we avoid duplicate code. This will be always useful for other tools (such as the erc conformance).

PR #354, opened 2019-10-26
Refactor of slither-upgradeability-check

### LINKED ISSUE #188: slither-check-upgradability: warning if constant are used
In the case of delegatecall proxy pattern, constant should be avoided, as they can create issues during upgrade.

slither-check-upgradability should warn if constant variables are present.

We should also add a note about it in the `constable-states` detector description

### LINKED ISSUE #274: slither-check-upgradeability raises "IndexError: list index out of range"

### LINKED ISSUE #331: slither-check-upgradeability: remove shadowing on fallback function
slither-check-upgradeability warns on a shadowing on the fallback function between the implementation and the proxy.
This shadowing is not an issue and should not be reported

### LINKED ISSUE #347: Create json result for slither-check-upgradability

montyly:
This PR contains several improvements on the upgradeability checks:

- Only the logic contract is needed, proxy becomes optional
- Change flag name [BREAKING CHANGE]
- Clean main to ease the add of new checks
- Clean variable ordering logic; merge the two function into one generic (fix #274)
- Refactor json output and add source mapping (fix #347). 
- Add check for constant conformance (fix #188)
- Overall output improvements
- Remove FP in case of fallback function collision (fix #331)

The json output and the constant conformance checks need to be tested. 
The documentation needs to be updated to merge this PR.

### DIFF
--- a/scripts/travis_test_upgradability.sh
+++ b/scripts/travis_test_upgradability.sh
@@ -128,6 +128,31 @@ then
     exit -1
 fi
 
+slither-check-upgradeability "$DIR_TESTS/contractV1.sol" ContractV1 --solc solc-0.5.0 --new-contract-filename "$DIR_TESTS/contract_v2_constant.sol" --new-contract-name ContractV2 > test_10.txt 2>&1
+DIFF=$(diff test_10.txt "$DIR_TESTS/test_10.txt")
+if [  "$DIFF" != "" ]
+then
+    echo "slither-check-upgradeability 10 failed"
+    cat test_10.txt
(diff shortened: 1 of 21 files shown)


montyly:
This PR is now based on the upcoming JSON refactoring: https://github.com/crytic/slither/pull/355

PR #355, opened 2019-10-26
Move json add_ functions from abstractdetectors to json_utils

montyly:
The goal of this PR is to help the standardization of all the json output.

Any tools / third-party tools will be able to follow the json format already used by the detectors. 
Example:
```python
from slither.utils import json_utils

json_elem = json_utils.generate_json_result(info)
json_utils.add_variable_to_json(my_var, json_elem)

json_utils.output_json('filename.json', None, { "my_custom_tool": json_elem}

```

Currently `generate_json_result` returns an `OrderedDict`. I am hesitating to create a specific object, and change the API to:

```python
from slither.utils import json_utils

json_elem = json_utils.generate_json_result(info)
json_elem.add_variable_to_json(my_var)

json_utils.output_json('filename.json', None, { "my_custom_tool": json_elem}
```

The main limitation is that as long as `generate_json_result` returns a `OrderedDict`, it might be easier for users to manipulate and add custom fields if needed. Also, users are going to mix between dictionary/list and the object returned by `generate_json_result`, which might be confusing if it's not a standard python object.

This PR is an API breaking change and will require to update the private detectors.

### DIFF
--- a/slither/detectors/abstract_detector.py
+++ b/slither/detectors/abstract_detector.py
@@ -1,8 +1,9 @@
 import abc
 import re
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
+
+from slither.utils import json_utils
 from slither.utils.colors import green, yellow, red
-from slither.core.source_mapping.source_mapping import SourceMapping
 from slither.formatters.exceptions import FormatImpossible
 from slither.formatters.utils.patches import apply_patch, create_diff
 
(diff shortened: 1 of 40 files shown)


montyly:
I went back and remove the breaking API change: `AbstractDetector` will keep the `self.add_X_to_json` functions. These functions will just be a wrapper around the `json.utils.add_X` functions.

I think this should be simpler to use for the devs working only on detectors.

PR #357, opened 2019-10-28
fix(detectors/attributes/incorrect_solc): report 0.5.11 as old version

### LINKED ISSUE #320: solc-version detector: Missing 0.5.11
`0.5.11` is incorrectly reported as an old version.

yxliang01:
This fixes the report that 0.5.11 is an old version.

This fixes #320


yxliang01:
Hmm @montyly So is this PR still relevant? If you are thinking to directly implement the version comparison-based method, we can close this. But, in my opinion, is the ETA for that method is uncertain, it might be still good to merge this to have 0.5.11 correctly categorized before a better mechanism is applied.

montyly:
Hi @yxliang01, thanks for pointing this out! 

I merged your PR, so we have a temporary fix until we rewrite the detector.

PR #360, opened 2019-11-01
[WIP] Adds initial support for kspec coverage tool

rajeevgopalakrishna:
This PR adds support for a new kspec coverage tool.

Usage: `slither-kspec-coverage contract.sol kspec.md`

The tool currently checks which functions in the specified `contract.sol` have corresponding [Klab](https://github.com/dapphub/klab) specifications in the provided `kspec.md`. It outputs a table illustrating which contract functions have `kspec funcs`, `kspec reached funcs` or `not covered funcs`.

`kspec funcs` indicates contract functions that have corresponding Klab specifications. `kspec reached funcs` indicates contract functions that do not have corresponding Klab specifications themselves but are called by functions that do have Klab specifications. `not covered funcs` highlights contract functions that do _not_ have corresponding Klab specifications. The last two categories of functions need attention because they do not have Klab coverage.

For example: `slither-kspec-coverage ./kspec_coverage/test/token/token.sol ./kspec_coverage/test/token/spec.md` produces:

```
+-----------------+---------------------------+-----------------------+---------------------+
| Contract Name   | kspec funcs               | kspec reached funcs   | not covered funcs   |
+=================+===========================+=======================+=====================+
| Token (100.00%) | transfer(address,uint256) | add(uint256,uint256)  |                     |
|                 |                           | sub(uint256,uint256)  |                     |
+-----------------+---------------------------+-----------------------+---------------------+
1/3 (33.33%) functions are directly covered by a kspec
2/3 (66.67%) functions are reached by a kspec
0/3 (0.00%) functions are not reached by a kspec
There is a 0.00% standard deviation among percentage of functions covered per contract
```


slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+from tabulate import tabulate
+
+def get_all_covered_kspec_functions(target):
+    # Create a set of our discovered functions which are covered
+    covered_functions = set()
+
+    BEHAVIOUR_PATTERN = re.compile('behaviour\s+(\S+)\s+of\s+(\S+)')
+    INTERFACE_PATTERN = re.compile('interface\s+([^\r\n]+)')
+
+    def refactor_type(type):
```
montyly: Is there a particular reason to have an inner function?

I would recommend creating a private function instead. Overall inner functions are slower to execute and make the code more difficult to maintain.

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+                function_arguments = function_full_name[start:end].split(',')
+                function_arguments = [refactor_type(arg.strip().split(' ')[0]) for arg in function_arguments]
+                function_full_name = function_full_name[:start] + ','.join(function_arguments) + ')'
+                covered_functions.add((contract_name, function_full_name))
+                i += 1
+        i += 1
+    return covered_functions
+
+
+def get_slither_functions(slither, include_functions=True, include_modifiers=False, include_interfaces=False,
```
montyly: I am not sure about the expected behavior of this function.

We could probably simplify it with `slither.functions` and `slither.funcitons_and_modifiers`. 

But I think the current format is not going to work as expected; every contract object creates an instance of every function inherited, so this might mislead the code coverage. I think we should use `contract.functions_declared`.

If `include_functions` is never set to False, we can remove it from the parameters

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+        if include_variable:
+            for variable in contract.variables:
+                if variable.visibility not in ['public', 'external']:
+                    continue
+                results.append((contract, variable))
+
+    return results
+
+
+def _get_functions_reached(compiled_functions, origin_func_desc, results):
```
montyly: Can you describe what is the purpose of this function?

We might be able to use `function.all_high_level_calls()` and similar to avoid the recursion

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # Print out messages regarding covered functions
+    covered_percentages = []
+    previous_contract = None
+    header_row = ['Contract Name', 'kspec funcs', 'kspec reached funcs', 'not covered funcs']
+    rows = []
+    current_kspec_direct, current_kspec_indirect, current_kspec_missing = ([], [], [])
+    count_kspec_direct, count_kspec_indirect, count_kspec_missing = (0, 0, 0)
+
+    def add_current_row():
```
montyly: Similar to the above, we should avoid inner functions

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+            current_kspec_indirect.append(compiled_func_desc[1])
+        else:
+            current_kspec_missing.append(compiled_func_desc[1])
+
+    # Add our last constructed row
+    if previous_contract is not None:
+        add_current_row()
+
+    # Create our table and print it
+    print(tabulate(rows, header_row, tablefmt='grid'))
```
montyly: Can we move to a logger instead of print?

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+    ))
+    print("{}/{} ({:.2%}) functions are reached by a kspec".format(
+        count_kspec_indirect, total_functions_displayed, count_kspec_indirect/total_functions_displayed
+    ))
+    print("{}/{} ({:.2%}) functions are not reached by a kspec".format(
+        count_kspec_missing, total_functions_displayed, count_kspec_missing/total_functions_displayed
+    ))
+
+    # Calculate the general deviation among coverage
+    average_percentage = sum(covered_percentages) / len(covered_percentages)
```
montyly: It might lead to a `/ 0`

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+    average_deviation_percentage = 0
+    for covered_percentage in covered_percentages:
+        average_deviation_percentage += pow(covered_percentage - average_percentage, 2)
+    average_deviation_percentage = sqrt(average_deviation_percentage / len(covered_percentages))
```
montyly: Similar, this might lead to a `/ 0`

montyly:
We should also add unit tests to ensure it works on all the verified Solidity contracts from https://github.com/runtimeverification/verified-smart-contracts

slither/tools/kspec_coverage/analysis.py
```diff
@@ -0,0 +1,215 @@
...
+            contract_name = match.groups()[1]
+            match = INTERFACE_PATTERN.match(lines[i + 1])
+            if match:
+                function_full_name = match.groups()[0]
+                start, end = function_full_name.index('(') + 1, function_full_name.index(')')
+                function_arguments = function_full_name[start:end].split(',')
+                function_arguments = [refactor_type(arg.strip().split(' ')[0]) for arg in function_arguments]
+                function_full_name = function_full_name[:start] + ','.join(function_arguments) + ')'
+                covered_functions.add((contract_name, function_full_name))
+                i += 1
```
Xenomega: I would refactor this code, it's weak and too order dependent/specific. I can imagine it causing bugs. It was just kept t his way because it worked without issue for the specific codebase it was designed for.

rajeevgopalakrishna:
Refactor complete. Uses `json_utils`.

montyly:
I am closing this in profit of https://github.com/crytic/slither/pull/364

PR #362, opened 2019-11-04
Make generate_json_result generating the JSON elements automatically

montyly:
API break: generate_json_result takes a list as input, instead of a str

The list can contain a string, or any elements that can be converted to a string with source mapping
Ex: `Variable`, `Contract`, `Event`, `Function`, ..
`generate_json_result` will automatically add the elements to the json.
As a result, detectors only need to add the elements to the info list, and do not need to add the elements to the json.

This simplifies the API, and help to standardize our ouputs.

`AbstractDetectors` has a `STANDARD_JSON` parameter, if set to False, `generate_json_result` will not add the elements to the json

For example:
```python
                    info = 'Backdoor function found in {}.{} ({})\n'
                    info = info.format(contract.name, f.name, f.source_mapping_str)

                    # Add the result in result
                    json = self.generate_json_result(info)
                    self.add_function_to_json(f, json)
```
becomes
```python
                    info = ['Backdoor function found in ', f, '\n']

                    # Add the result in result
                    json = self.generate_json_result(info)
```


montyly:
The entire API for outputing the results is changed:

## Detector
A detector will return a list of `slither.utils.output.Output` objects, that are generated from `abstract_detector.generate_output(info)`.

If the detector does not require specific handling, `info` is a list of `str`| slither object. For example:

https://github.com/crytic/slither/blob/f70e89bf12c1d5211050ea2b7de7bc785af42778/slither/detectors/statements/tx_origin.py#L71-L72

Slither takes care to create a string representation, with source mapping, and to generate the json with all the information

## Printer
Printer works similar to the detector, but they return only one `Output` instead of a list

## Tools
All the tools use `slither.utils.output.Output` to generate output. 

`Output` contains different `add_*` functions for specific usages. There is a generic `Output.add(obj)`, which is enough and simplifies the API.

`Output` is not directly serializable to a json. `Output.data` returns the python dictionary that can be serializable. `Output.data` should be used to store the output result in another dictionary.

Issue #178, opened 2019-02-22
Improve reporting on delegatecall proxy pattern


montyly:
A recurrent source of false positives is the use of `delegatecall` proxy pattern, which leads to reporting issues like the uninitialized variables, that are inevitable with the proxy-based architecture of the pattern.

We should investigate if, once the delegatecall proxy pattern is detected, we could filter some of the results.

montyly:
Fixed with https://github.com/crytic/slither/pull/370

PR #370, opened 2019-11-13
Uninitialized state variable detector: remove FP due to delegatecall proxy

montyly:
Fix https://github.com/crytic/slither/issues/178


montyly:
Example:
```solidity
// Do not use this code, it is buggy

contract Initializable{}

contract C is Initializable{

    address owner;

    function use_owner() public {
        require(owner == msg.sender);
    }

}

contract Proxy{

    address owner = msg.sender;

    function () external{
        assembly{
            let ret := delegatecall(
                    0,                
                    0,           
                    0,    
                    0,        
                    0,                 
                    0                   
                )
        }
    }

}
```

`C.owner` is not more detected as an uninitialized variable

PR #371, opened 2019-11-14
Updated pip3 for Macos

enderphan94:
Install by pip does not work in Macos


montyly:
Good catch. Thank you for your contribution @enderphan94

PR #376, opened 2019-11-25
--filter-paths option now handles regex and relative paths

erib3:
I convert each element in source_mapping_elements with normpath() (collapses redundant separators and up-level references)

Now using regular expression search to see if pattern matches the src_mapping. Each pattern is run through method that handles relative paths.

I also broke up list comprehensions to multiple lines for readability

### DIFF
--- a/slither/core/slither_core.py
+++ b/slither/core/slither_core.py
@@ -4,6 +4,7 @@
 import os
 import logging
 import json
+import re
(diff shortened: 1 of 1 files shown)


montyly:
Hi, @erib3 

Thanks for the work. I made a minor change (catch and print `re` errors) and merged the PR in `dev`.

PR #378, opened 2019-11-27
Add upper and lower bound comparison for solc detector

### LINKED ISSUE #320: solc-version detector: Missing 0.5.11

uivlis:
Should fix #320

### DIFF
--- a/slither/detectors/attributes/incorrect_solc.py
+++ b/slither/detectors/attributes/incorrect_solc.py
@@ -53,7 +53,7 @@ Use Solidity 0.4.25 or 0.5.3. Consider using the latest version of Solidity for 
             return self.LESS_THAN_TXT
         version_number = '.'.join(version[2:])
         if version_number not in self.ALLOWED_VERSIONS:
-            if version_number > self.ALLOWED_VERSIONS[-1]:
+            if list(map(int, version[2:])) > list(map(int, self.ALLOWED_VERSIONS[-1].split('.'))) :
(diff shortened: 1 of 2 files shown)


slither/detectors/attributes/incorrect_solc.py
```diff
@@ -54,7 +53,7 @@ def _check_version(self, version):
             return self.LESS_THAN_TXT
         version_number = '.'.join(version[2:])
         if version_number not in self.ALLOWED_VERSIONS:
-            if version_number in self.TOO_RECENT_VERSIONS:
+            if version_number > self.ALLOWED_VERSIONS[-1]:
```
montyly: I am not sure this comparison is going to be really robust.
For example if we have as the latest allowed version "0.5.10", and the file analyze uses "0.5.9", the message reported will be that the version used is too recent (`"0.5.9" > "0.5.10"` returns true).

A more robust approach is probably to:
- Check that the version starts with `0.5`
- Extract the minor version number, and compare it as an integer

yxliang01:
```suggestion
            if list(map(int, version[2:])) > list(map(int, self.ALLOWED_VERSIONS[-1].split('.'))) :
```

montyly:
I merged the branch in `dev`, thanks to @uivlis and @yxliang01 for the work!

PR #379, opened 2019-11-27
Split const_functions into asm and state cases. Updated detectors and tests.

uivlis:
Absolutely not sure if tests will pass, I haven't found any info in the docs on how to proceed with the tests. Fingers crossed!


uivlis:
Alright, there's just one little test that fails. Apparently, if you test `const-functions-asm` on `constant-0.5.1.sol` it detects the constant function having assembly code, although it shouldn't detect it, since `self._solc_version < 0.5.0". I wonder...

montyly:
Merged through #380

PR #396, opened 2020-01-29
Open source 4 detectors

montyly:
- boolean-cst
    - tautology
    - boolean-equal
    - divide-before-multiply


montyly:
`boolean-cst` has one type of FP that we should remove:
```solidity
while(true) {
  if(cond) return;
}
```

PR #411, opened 2020-03-20
Clean up IRs classes

montyly:
Clean up unused import in IRs and add `function.slithir_ssa_operations` method


montyly:
Additionally, this PR fixes the dapp regression tests (addition of one new finding)

PR #429, opened 2020-04-02
Fixed cfg_to_dot printing to file

aaronyoo:
Fixed a small error that caused contents of dot file not to be written to the specified file. For an example usage see `examples/scripts/export_to_dot.py`.


montyly:
Hi @aaronyoo ,

Thanks for pointing this issue. `slithir_cfg_to_dot` actually needed to return the content for `printers/functions/cfg.py`.

I ended up splitting the function into two versions: one returns the dot content, one writes the file #432

PR #437, opened 2020-04-10
Echidna printer: improve constant function detection

montyly:
Heuristic:
- If view/pure with Solidity >= 0.5 -> Return true
- If it contains assembly -> Return false (Slither doesn't analyze asm)
- Otherwise, check the rules from
    https://solidity.readthedocs.io/en/v0.5.0/contracts.html?highlight=pure#view-functions
- One exception: internal dynamic calls are not correctly handled, so we consider them as non-constant

### DIFF
--- a/slither/core/declarations/function.py
+++ b/slither/core/declarations/function.py
@@ -599,7 +599,7 @@ class Function(ChildContract, ChildInheritance, SourceMapping):
         """
             list(SolidityFunction): List of Soldity calls
         """
-        return list(self._internal_calls)
+        return list(self._solidity_calls)
 
     @property
     def high_level_calls(self):
--- a/slither/printers/guidance/echidna.py
+++ b/slither/printers/guidance/echidna.py
@@ -7,13 +7,14 @@ from typing import Dict, List, Set, Tuple
 
 from slither.core.cfg.node import Node
 from slither.core.declarations import Function
-from slither.core.slither_core import Slither
-from slither.core.variables.variable import Variable
-from slither.printers.abstract_printer import AbstractPrinter
 from slither.core.declarations.solidity_variables import SolidityVariableComposed, SolidityFunction, SolidityVariable
-from slither.slithir.operations import Member, Operation
-from slither.slithir.operations.binary import Binary, BinaryType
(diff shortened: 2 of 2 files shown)


montyly:
In addition, this PR adds:
- The type for the constant used. If there is an explicit type conversion (ex : `address(0x41)`) , the constant is the destination's type (ex : `0x41, address` ):
```json
    "constants_used": {
        "Ownable": {
            "transferOwnership(address)": [
                [
                    {
                        "value": 0,
                        "type": "address"
                    }
                ]
            ]
        },
```
- The constructor of each contract
```json
    "constructors": {
        "Ownable": "Ownable()",
        "BasicToken": "Ownable()",
```
- The list of functions impacted by a function (if `f` write a state variable read by `g`, then `f` impacts `g`). The output contains both `impacts` and `is_impacted_by` to simplify lookup for third-parties tools.
```json
            "approve(address,uint256)": {
                "impacts": [
                    "transferFrom(address,address,uint256)",
                    "approve(address,uint256)",
                    "allowance(address,address)"
                ],
                "is_impacted_by": [
                    "transferFrom(address,address,uint256)",
                    "approve(address,uint256)",
                ]
            },
```

montyly:
I added two additional info:
- `have_external_calls`: list of functions that call external contracts
- `call_a_parameter`: if a parameter is directly use as an external call.

Example:
```solidity
contract C{
    function test() public;
}

contract D{
    C c;

    function f() public{
        c.test();
    }

    function f2(address addr) public{
        C(addr).test();
    }

}
```

Will give
```
    "have_external_calls": {
        "D": [
            "f()",
            "f2(address)"
        ]
    },
    "call_a_parameter": {
        "D": [
            {
                "function": "f2(address)",
                "parameter_idx": 0,
                "signature": "test()"
            }
        ]
```

Note: ` "signature"` will be null for low level calls

montyly:
Added: 
- `use_balance`:  list of functions using `balance`
```solidity
contract C{

    function test() public{
        address(this).balance;
    }
}
```

```
    "use_balance": {
        "C": [
            "test()"
        ]
    }
```

Given the number of new features of this printer, the json must be documented.

PR #507, opened 2020-06-16
Implement support for calculating storage layout

samczsun:
Since some contracts can be extended by more than one derived contract, storage layout information can't be stored in the state variable itself.

Also, `state_variables_ordered` had duplicates which seemed kinda weird so I "fixed" that.

Tested by reading [this doc](https://github.com/ethereum/solidity/blob/develop/docs/internals/layout_in_storage.rst) and comparing the output with `solc --storage-layout`, but some automated tests would probably be nice

### DIFF
--- a/slither/core/solidity_types/array_type.py
+++ b/slither/core/solidity_types/array_type.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Tuple
 
 from slither.core.expressions import Literal
 from slither.core.expressions.expression import Expression
@@ -38,7 +38,7 @@ class ArrayType(Type):
         return self._length_value
 
     @property
-    def storage_size(self):
+    def storage_size(self) -> Tuple[int, bool]:
         if self._length_value:
             elem_size, _ = self._type.storage_size
             return elem_size * int(self._length_value.value), True
--- a/slither/core/solidity_types/array_type.py
+++ b/slither/core/solidity_types/array_type.py
@@ -37,6 +37,13 @@ class ArrayType(Type):
     def lenght_value(self) -> Optional[Literal]:
         return self._length_value
 
+    @property
+    def storage_size(self):
+        if self._length_value:
+            elem_size, _ = self._type.storage_size
+            return elem_size * int(self._length_value.value), True
+        return 32, True
+
     def __str__(self):
         if self._length:
             return str(self._type) + "[{}]".format(str(self._length_value))
--- a/slither/core/solidity_types/elementary_type.py
+++ b/slither/core/solidity_types/elementary_type.py
@@ -172,6 +172,13 @@ class ElementaryType(Type):
             return int(t[len("bytes") :])
         return None
 
+    @property
+    def storage_size(self):
+        if self._type == 'string' or self._type == 'bytes':
+            return 32, True
+
+        return int(self.size / 8), False
+
     def __str__(self):
         return self._type
 
--- a/slither/core/solidity_types/elementary_type.py
+++ b/slither/core/solidity_types/elementary_type.py
@@ -1,5 +1,5 @@
 import itertools
-from typing import Optional
+from typing import Optional, Tuple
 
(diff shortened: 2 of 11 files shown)


slither/solc_parsing/slitherSolc.py
```diff
@@ -420,4 +425,33 @@ def _convert_to_slithir(self):
             contract.fix_phi()
             contract.update_read_write_using_ssa()
 
+    def _compute_storage_layout(self):
```
montyly, line 428: I think the storage layout should be in `core`. Anything that is meant to be used by analyses/detectors/printers should not be in `solc_parsing`.

slither/core/solidity_types/array_type.py
```diff
@@ -32,6 +32,13 @@ def type(self):
     def length(self):
         return self._length
 
+    @property
```
montyly, line 35: Is there a type that should not implement `storage_size`? If not I think we should extend `Type` from abc, and make it an `abc.abstractmethod`

It would be nice also to add types and an explanation why it returns a boolean

slither/core/solidity_types/elementary_type.py
```diff
@@ -4,6 +4,7 @@
 
 
 # see https://solidity.readthedocs.io/en/v0.4.24/miscellaneous.html?highlight=grammar
+from slither.exceptions import SlitherException
```
montyly, line 7: I think the exception is not used

slither/core/solidity_types/array_type.py
```diff
@@ -32,6 +32,13 @@ def type(self):
...
+    def storage_size(self):
+        if self._length_value:
+            elem_size, _ = self._type.storage_size
+            return elem_size * int(self._length_value.value), True
```
montyly, line 39: I am not sure `_length` is always defined here. It could be none for dynamic array (I am not sure).
We need to handle this case to avoid a crash

slither/solc_parsing/declarations/contract.py
```diff
@@ -224,7 +224,9 @@ def parse_structs(self):
     def parse_state_variables(self):
         for father in self.inheritance_reverse:
             self._variables.update(father.variables_as_dict())
-            self._variables_ordered += father.state_variables_ordered
+            for var in father.state_variables_ordered:
```
montyly, line 227: Do we have an example where this could lead to a duplicate?

If so it might be worth looking why it happens, it might hide some weird bugs in another location

samczsun: For the following contract:
```
contract A1 {
    uint a1;
}

contract B1 is A1 {
    uint b1;
}

contract B2 is A1 {
    uint b2;
}

contract C1 is B1, B2 {
    uint c1;
}

contract D1 is C1 {
    uint d1;
}
```

On master:
```
D1:
+-------+---------+
|  Name |   Type  |
+-------+---------+
| A1.a1 | uint256 |
| A1.a1 | uint256 |
| B1.b1 | uint256 |
| A1.a1 | uint256 |
| B2.b2 | uint256 |
| A1.a1 | uint256 |
| A1.a1 | uint256 |
| B1.b1 | uint256 |
| A1.a1 | uint256 |
| B2.b2 | uint256 |
| C1.c1 | uint256 |
| D1.d1 | uint256 |
+-------+---------+
```

On this branch
```
D1:
+-------+---------+------+--------+
|  Name |   Type  | Slot | Offset |
+-------+---------+------+--------+
| A1.a1 | uint256 |  0   |   0    |
| B1.b1 | uint256 |  1   |   0    |
| B2.b2 | uint256 |  2   |   0    |
| C1.c1 | uint256 |  3   |   0    |
| D1.d1 | uint256 |  4   |   0    |
+-------+---------+------+--------+
```

montyly:
Note for the next release note: the documentation of `state_variables_ordered` stated that it returns the shadowed variables, so this PR contains a breaking change.

I checked our usages of the property, I don't see any place where this change creates a negative impact

PR #544, opened 2020-07-21
[WIP] Add slither-mutate

montyly:
`slither-mutate` will inject fault to a codebase. Based on https://arxiv.org/abs/2006.11597.

Only three mutators are implemented for now. The tool follows the detectors architecture to ease the add of new mutators.

It will generate git patches for the mutations. We could generate entire files, or overwrite existing ones - it will depend on its usage.

Example:
```solidity
contract C{

    uint a = 0;
    uint b = a + 1;

    function f(bool condition) public {
        uint c = 10;
        c = 20;

        if (condition) {
            c = 10;
        }
        return;
    }

}
```
```
$ slither-mutate test.sol 
--- ./test.sol
+++ ./test.sol
@@ -7,7 +7,7 @@
         uint c = 10;
         c = 20;
 
-        if (condition) {
+        if (true) {
             c = 10;
         }
         return;

--- ./test.sol
+++ ./test.sol
@@ -1,7 +1,7 @@
 contract C{
 
     uint a = 0;
-    uint b = a + 1;
+    uint b ;
 
     function f(bool condition) public {
         uint c = 10;

--- ./test.sol
+++ ./test.sol
@@ -1,10 +1,10 @@
 contract C{
 
-    uint a = 0;
+    uint a ;
     uint b = a + 1;
 
     function f(bool condition) public {
-        uint c = 10;
+        uint c ;
         c = 20;
 
         if (condition) {
```


agroce:
What are the three operators, exactly?  The code gives me an idea, but I'm not sure

Issue #549, opened 2020-07-22
Add Github's super-linter as an Action


dguido:
https://github.com/github/super-linter

elopez:
I'm working on this one, I'll send a PR soon :smile:

montyly:
Done with https://github.com/crytic/slither/pull/614

PR #554, opened 2020-07-26
Integrate Github Super Linter

elopez:
This series of commits add an action to lint the codebase using Github
Super Linter. Unfortunately the Python support in the Super Linter is
lacking[1] and there is no simple way to perform installation of
dependencies before linting, so a lot of import errors occur. To work
around this problem, we also introduce an independent pylint run in
the Actions workflow and disable the Super Linter one.

[1]: https://github.com/github/super-linter/issues/157

Closes: #549

### DIFF
--- a/.github/linters/.python-lint
+++ b/.github/linters/.python-lint
@@ -1,0 +1,543 @@
+[MASTER]
+errors-only=
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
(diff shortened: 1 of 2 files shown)


elopez:
You can see the linting workflow in action here: https://github.com/crytic/slither/runs/912512264?check_suite_focus=true

dguido:
This looks good at first glance!

I'd want to clear as many of the linter issues as possible before merging. It looks like there are a bunch of easy markdown, bash, and docker fixes.

The pylint configuration looks like a good place to start. There are lots of disabled checks but that's to be expected.

elopez:
Ah, I just saw your message. I already pushed the simple fixes on a separate PR, but I can bring them here as well if that's what you prefer :smile:

dguido:
Nah that's fine. Thanks for the extra follow through!

montyly:
Hi @elopez 

Thank you for your contribution. I ended up changing a bit the config here: https://github.com/crytic/slither/pull/614

PR #555, opened 2020-07-26
Fix linting problems

elopez:
With the introduction of linting (see #549, #554) several warnings
and errors appeared. This PR fixes all of the Docker, shell and
Markdown ones and some of the Python ones.


montyly:
That's great, thanks @elopez

Issue #579, opened 2020-08-18
DoS when contract contains literals with exponents with a mantissa of zero


Xenomega:
The following contract will cause a DoS in slither:
```
contract DoS {
    function f() public pure {
        int a = 0E123456789;
    }
}
```

The culprit is this line: `int a = 0E123456789;`

montyly:
Fixed with #608

PR #608, opened 2020-08-28
Fixing DOS large exponent in scientific notation constants.

laudiacay:
Fixes issue #579.

The issue was that slither was computing 10**HUGE_NUMBER eagerly in order to multiply it with the first part of the scientific notation.... which takes a while, and eats up lots of memory, because we're using bignums for this. I changed the behavior so that, if the first part is zero, just return zero. Otherwise, if the exponent is greater than 80, raise a ValueError, because there's no way the result will fit into a solidity integer type unless they're just writing it out in a really really troll-y way.

This will raise on something like `0.{80 zeros}1E80`. Not sure if solc supports that, but if we wanted to, we could add some sort of behavior with alternately multiplying the base by 10 and decrementing the exponent until it's zero... but that'd require some checks too to make sure the exponent isn't still too large.


laudiacay:
also fixing #587 by just adding another decimal cast

PR #614, opened 2020-08-31
Try github super linter

montyly:
Related to https://github.com/crytic/slither/pull/554


montyly:
Currently `super-linter` does not allow for black config file. I will wait that https://github.com/github/super-linter/pull/662 is merged to re-enable black

PR #636, opened 2020-09-14
Enable github super linter as a cron job

montyly:


montyly:
Looks like black was downgraded in github super linter, which cause all our PRs to fail: https://github.com/github/super-linter/issues/722

PR #639, opened 2020-09-19
Fix mixup of ordering and remove extra description

chanhosuh:


montyly:
Good catch @chanhosuh, thanks!

PR #646, opened 2020-09-28
feature to skip assembly

samczsun:


slither/solc_parsing/declarations/function.py
```diff
@@ -849,6 +849,8 @@ def _parse_statement(self, statement: Dict, node: NodeSolc) -> NodeSolc:
                 asm_node = self._new_node(NodeType.ASSEMBLY, statement["src"])
                 self._function.contains_assembly = True
                 # Added with solc 0.4.12
+                if "AST" in statement:
+                    asm_node.underlying_node.add_inline_asm(statement["AST"])
```
montyly, line 853: This returns a dictionary, right? If so, we should convert it to string, or skip it.

Otherwise, it will break some usages, ex: https://github.com/crytic/slither/blob/9f64778a85e749c4410a217b46687965f90b0564/slither/printers/summary/human_summary.py#L195-L197

montyly (changes requested):

PR #651, opened 2020-09-29
fix wiki link for naming conventions

gorgos:


montyly:
Good catch, thanks @gorgos

PR #666, opened 2020-10-14
slither-simil implementation draft

spilehchiha:
This is the initial commit for the most up-to-date implementation of `slither-simil.` Everything client-related has been removed and public datasets in the form of train / dev / test datasets will be added in the next commit.

### DIFF
--- a/slither/core/declarations/solidity_variables.py
+++ b/slither/core/declarations/solidity_variables.py
@@ -25,6 +25,7 @@ SOLIDITY_VARIABLES_COMPOSED = {
     "block.number": "uint256",
     "block.timestamp": "uint256",
     "block.blockhash": "uint256",  # alias for blockhash. It's a call
+    "block.chainid": "uint256",
     "msg.data": "bytes",
     "msg.gas": "uint256",
     "msg.sender": "address",


slither/tools/similarity/README.md
```diff
@@ -0,0 +1,30 @@
+# SlithIR Extraction and Processing
+
+There is a wiki page available for `slither-simil` on this [page.](https://github.com/crytic/slither/wiki/Code-Similarity-detector)
+
+
+## Installation adn Setup
```
FlashSheridan, line 6: “adn” should presumably be “and”.

slither/tools/similarity/README.md
```diff
@@ -0,0 +1,30 @@
...
+https://github.com/spilehchiha/crytic-compile/tree/dev-zip
+```
+
+On git-lfs: Go to this [link.](https://docs.github.com/en/github/managing-large-files/installing-git-large-file-storage)
+
+Some SlithIRs have already been extracted and are located as `datasets/output.csv`.
+
+Take the `output.csv` file and the `datasets/15kdump_model.bin` file already in the repository and use it as follows:
+```
+slither-simil test [datasets/15kdump_model.bin] --filename [file_name; can be a singleton file name or a directory path containing multiple smart ontracts either in .sol or .zip formats or a c ombincation of both] --fname [function_name with or without the preciding contract name (as in contractname[.]functionname) or a simple dot mark, meaning all the functions in the contracts provided by the previous argument flag;] --input [cache.npz or output.csv] --ntop [n; an ineteger; e.g. 3]
```
FlashSheridan, line 29: - “ontracts” should be “contracts”
- “c ombincation” should be “combination”
- “preciding” should be “preceding”
- “flag;”:  The semicolon is not needed, and semicolons on the command line make me nervous, even if they are supposed to be removed.
- “ineteger” should be “integer”

PR #694, opened 2020-11-19
Fixed crytic_revert_transfer_to_user_ERC20PropertiesTransferable() property

ggrieco-tob:
The `crytic_revert_transfer_to_user_ERC20PropertiesTransferable()` property was incorrect:

* If the balance cannot be increased, then it should revert, not return anything (because this property passed when it reverts).
* If the balance can be increased, and the call to transfer fails to revert, then returning true (or false in fact) is enough to make this property fail (not reverting).


slither/tools/properties/properties/ercs/erc20/properties/transfer.py
```diff
@@ -155,10 +155,10 @@
         description="Transfering more tokens than the balance will revert.",
         content="""
 \t\tuint balance = this.balanceOf(msg.sender);
-\t\tif (balance == (2 ** 256 - 1))
-\t\t\treturn true;
-\t\tbool transfer_other = transfer(crytic_user, balance+1);
-\t\treturn transfer_other;""",
+\t\tif (balance == (uint(-1)))
+\t\t\trevert();
+\t\ttransfer(crytic_user, balance+1);
```
montyly, line 160: This will not catch the ERC20 tokens that return false instead of reverting. We should return the result from  `transfer`  here

montyly (changes requested):

Issue #708, opened 2020-11-26
Slither 6.14 assertion error


Ferparishuertas:
After adding some new functionalities, to the smart contract, the following error raises. previous version with 6.14, works.

It seems to me the problem is the new functionality introduced on solidity 7.4, constants at file level

https://blog.soliditylang.org/2020/10/19/solidity-0.7.4-release-announcement/

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 693, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 79, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 62, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/usr/local/lib/python3.9/site-packages/slither/slither.py", line 81, in __init__
    self._parser.parse_contracts_from_loaded_json(ast, path)
  File "/usr/local/lib/python3.9/site-packages/slither/solc_parsing/slitherSolc.py", line 132, in parse_contracts_from_loaded_json
    assert contract_data[self.get_key()] in [
AssertionError
ERROR:root:None
ERROR:root:Error in .
ERROR:root:Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 693, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 79, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.9/site-packages/slither/__main__.py", line 62, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/usr/local/lib/python3.9/site-packages/slither/slither.py", line 81, in __init__
    self._parser.parse_contracts_from_loaded_json(ast, path)
  File "/usr/local/lib/python3.9/site-packages/slither/solc_parsing/slitherSolc.py", line 132, in parse_contracts_from_loaded_json
    assert contract_data[self.get_key()] in [
AssertionError

elenadimitrova:
[...]

Issue #619, opened 2020-09-03
Add support for top-level functions


montyly:
https://solidity.ethereum.org/2020/09/02/solidity-0.7.1-release-announcement/

Similar to https://github.com/crytic/slither/issues/487

We should probably have some re-org in the way the contracts-related object exist. As the codebase relies heavily on the assumption that everything is attached to a contract, the reorg will require some work

montyly:
We should consider during the refactor that the other elements are also going to become top-level in the future

https://github.com/ethereum/solidity/issues/9671
https://github.com/ethereum/solidity/issues/9737

PR #728, opened 2020-12-14
Improve top level objects support

montyly:
This is a first step towards better handling of top-level objects

- Split declaration.Enum/Structure to EnumContract/EnumToplevel (same for structure)
- Remove the placeholder top level contract
- Update the parsing accordingly

This PR changes the API, so we will need to evaluate the impact of the public/private detectors. 

Related: https://github.com/crytic/slither/issues/619, https://github.com/crytic/slither/issues/708.


montyly:
Added support for top-level variable and functions

The support for top-level functions is partial. Among other, what is missing:
- Legacy ast support (impact mostly `dapp` codebases)
- `import  X as B` is not correctly supported if `B` is used to refer to top level objects. For example, `B.function` will lead to generate a High level call instead of an internal call at the IR level

PR #730, opened 2020-12-15
Ensure MemberAccess.member_type is of type Type

snd:
I needed to access the `member_type` attribute of a `MemberAccess` object, expected it to be of type `Type` but found that it was of type `str`.

There's a related `TODO` and commented out `assert` in the code:
https://github.com/crytic/slither/blob/9f64778a85e749c4410a217b46687965f90b0564/slither/core/expressions/member_access.py#L12-L13

Fixed it by parsing the `member_type` before passing it to `MemberAccess(...)`.
This caused `parse_type` to fail on the following inputs:

```
function (address[] memory) external
```

```
function (uint256) pure returns (function (bytes memory) payable returns (bool,bytes memory))
```

Modified `parse_type` to parse those correctly.

Finally commented in:

https://github.com/crytic/slither/blob/9f64778a85e749c4410a217b46687965f90b0564/slither/core/expressions/member_access.py#L12


snd:
Let me know if this requires additional parser tests and I'll add them.

snd:
shall I leave the `TYPE_CHECKING` in and also only `assert isinstance(member_type, Type)` if `TYPE_CHECKING`?

https://github.com/crytic/slither/pull/730/commits/50afec2e2d2705637553d458693fe80042e9dff3#diff-a9f31b75625f7229950bf3a5fa92bc0c18bd085e27ccdec43bd7006fd7a77191L6

montyly:
The import can be removed. We can also change
https://github.com/crytic/slither/blob/50afec2e2d2705637553d458693fe80042e9dff3/slither/core/expressions/member_access.py#L27
to
` def type(self) -> Type: `

snd:
Whoops, I forgot to add my git credentials on the new work  partition. Going to try to fix it.

snd:
fixed author with rebase

PR #744, opened 2020-12-29
[WIP] Add "eth-receive" printer which lists functions per contract that can receive ETH as well as requires/state-changes related to msg.sender and msg.value

snd:
I didn't find an existing printer that included this output.

Will update the printers documentation in the wiki once this is merged.

- [x] print all functions that are payable
- [ ] print "Related Statements" in those functions that involve `msg.sender` and `msg.value`
- [ ] ensure requires/asserts and state variable writes that involve `msg.sender` or `msg.value` are recognizable
- [ ] ensure modifiers are included correctly
- [ ] detect cases where ether is forwarded rather than kept
- [ ] communicate if "Related Statements" are in called functions

Example:
```
slither --print eth-receive 0xc32f9b0292965c5dd4a0ea1abfcc1f5a36d66986
...
INFO:Printers:
TimelockInterface:
+----------------------------------------------------------+
|                           Name                           |
+----------------------------------------------------------+
| executeTransaction(address,uint256,string,bytes,uint256) |
+----------------------------------------------------------+

DualGovernorAlpha:
+------------------+
|       Name       |
+------------------+
| execute(uint256) |
+------------------+

INFO:Slither:0xc32f9b0292965c5dd4a0ea1abfcc1f5a36d66986 analyzed (5 contracts)
```

### DIFF
--- a/slither/printers/all_printers.py
+++ b/slither/printers/all_printers.py
@@ -17,4 +17,4 @@ from .summary.require_calls import RequireOrAssert
 from .summary.constructor_calls import ConstructorPrinter
 from .guidance.echidna import Echidna
 from .summary.evm import PrinterEVM
-from .functions.payable import Payable
+from .functions.eth_receive import EthReceivePrinter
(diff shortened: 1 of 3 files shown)


montyly:
As no progress has been made since 2021, I am closing the PR.

The idea was interesting, but further work is needed. For example in
```solidity
contract A{
	uint v;
	function set() internal{
		v += msg.value;
	}
	function f() payable public{
		require(msg.value==0);
	}
}
```

The side effect was not shown.

PR #769, opened 2021-01-14
Fix multi-file --filter-paths bug

Eenae:
It looks like multiple comma-separated file paths provided in the `--filter-paths` argument don't work since the `source_mapping_elements` mapping object get exhausted after the first iteration of the loop here https://github.com/crytic/slither/blob/705f8db526ace58fec9bfd9f025765d9bc4860a3/slither/core/slither_core.py#L342.

Suggesting a straightforward fix.


0xalpharush:
This was replaced by https://github.com/crytic/slither/pull/1196

PR #781, opened 2021-02-01
Add hardhat to the list of supported projects

fvictorio:
As far as I know, slither works fine in Hardhat projects, but since it's not indicated on the readme people often ask about it on our discord.


montyly:
Good catch @fvictorio, thanks for the contribution

PR #791, opened 2021-02-23
Rename __init__.y.py to __init__.py - fixes #790

### LINKED ISSUE #790: `slither-flat` crashes for installed package
When run from an installed `slither-analyzer` package, `slither-flat` crashes with the following exception:

```
File "//path/to//venv/lib/python3.9/site-packages/slither/tools/flattening/flattening.py", line 15, in <module>
    from slither.tools.flattening.export.export import (
ModuleNotFoundError: No module named 'slither.tools.flattening.export'
```
Further inspection shows that the `slither.tools.flattening.export` package/directory is not actually included in the PYPI package (at least for v0.7.0), and is similarly missing when installing from a cloned repo via `pip install .`

gnattishness:
Fix flattening.export package so that it correctly is identified as a python package.
This should ensure it is included in the `setuptools.find_packages()` result.


gnattishness:
fixes #790

montyly:
Hi @gnattishness 

Thanks for the fix. It's crazy that we did not catch this earlier on. I will add a CI test before the next release.

PR #821, opened 2021-04-07
Update `solidity_variables.py`

sobolev-igor:
Hi!
My contracts failed to parse and it seems like it's the only thing required to be fixed)


montyly:
Hi @sobolev-igor 

Nice catch, thanks for your contribution

PR #822, opened 2021-04-09
add unused-return-transfers detector + update unused-return

### LINKED ISSUE #816: Specializing unused-return detector
The unused return detector gives usually good results: https://github.com/crytic/slither/wiki/Detector-Documentation#unused-return

But I think we can split it into two detectors:
- Lack of return check on transferFrom/transfer
- All the other cases

The reasoning is that the lack of return check for ERC20 tokens is usually a high severity issue that can lead to a significant loss. Specializing this detector will help to triage the results

rmi7:
fixes #816 

- adds a new detector `unused-return-transfers`
- updates `unused-return` to exclude `transfer`/`transferFrom` calls, since they are caught by the new detector.


montyly:
I made the following changes:
- Fix unused return (it was not detecting unused return from state variable calls anymore)
- Inherit from Unused return (avoid duplicate code)
- Rename to unchecked-return
- Update description/recommendations

Additionally removed txt_description from all the UnusedReturn derived
class (it's not used anymore)

Issue #559, opened 2020-07-28
Add Solidity 0.7 support


montyly:
https://solidity.ethereum.org/2020/07/28/solidity-v0.7.0-release-announcement/

- [ ] Inline Assembly: Disallow . in user-defined function and variable names.
- [ ] Inline Assembly: Slot and offset of storage pointer variable x are accessed via x.slot and x.offset instead of x_slot and x_offset.
- [ ] JSON AST: Mark hex string literals with kind: "hexString".
- [ ] JSON AST: Remove members with null value from JSON output.
- [ ] Reference Resolver: using A for B only affects the contract it is mentioned in and not all derived contracts
- [ ] Parser: Introduce Unicode string literals: unicode"😃".

montyly:
I forgot to close this issue with https://github.com/crytic/slither/releases/tag/0.6.14. 

If I am not mistaken, only `slot` was missing, which is covered by https://github.com/crytic/slither/pull/833

If we find missing features, we will create individual gh issue.

PR #831, opened 2021-04-22
Tightened the supported versions of solidity

AntoineRondelet:
While using slither to analyze some contracts written with solidity `0.8` (see [here](https://github.com/clearmatics/zeth/tree/develop/zeth_contracts/contracts)), I got an error (i.e. `ERROR:root:unresolved reference to identifier vk.slot `) related to the `x.slot` syntax introduced in solidity 0.7 as a replacement of the syntax `x_slot` to retrieve the slot pointed to by the variable `x`.

See the difference between the [solidity 0.7 doc](https://docs.soliditylang.org/en/v0.7.0/assembly.html?highlight=mstore#access-to-external-variables-functions-and-libraries) and the [solidity 0.6 doc](https://docs.soliditylang.org/en/v0.6.0/assembly.html?highlight=mstore#access-to-external-variables-functions-and-libraries) for further details.

That is somewhat confusing, provided that the README states that contracts written with solidity >= 0.4 can be analyzed with the tool, however, I realized (beyond the error above) that the syntax for solidity 0.7 is not yet fully supported: https://github.com/crytic/slither/issues/559

As a consequence, this PR clarifies which solidity versions are currently supported, as a way to remove potential future issues for other users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Slither is a Solidity static analysis framework written in Python 3. It runs a s
 * Easily integrates into continuous integration and Truffle builds
 * Built-in 'printers' quickly report crucial contract information
 * Detector API to write custom analyses in Python
-* Ability to analyze contracts written with Solidity >= 0.4
+* Ability to analyze contracts written with Solidity `>= 0.4` and `< 0.7`
 * Intermediate representation ([SlithIR](https://github.com/trailofbits/slither/wiki/SlithIR)) enables simple, high-precision analyses
 * Correctly parses 99.9% of all public Solidity code
 * Average execution time of less than 1 second per contract


montyly:
Hi @AntoineRondelet. 

We support Solidity 0.7, but it looks like we missed the `.slot` change from the changelog. https://github.com/crytic/slither/pull/833 will fix that.

Let us know if you have other issues with Slither

AntoineRondelet:
> Hi @AntoineRondelet.
> 
> We support Solidity 0.7, but it looks like we missed the `.slot` change from the changelog. #833 will fix that.
> 
> Let us know if you have other issues with Slither

Thanks @montyly, no problem. Looking forward to #833 being merged then :)

PR #832, opened 2021-04-23
Improve AST parsing tests + add 0.8

montyly:


montyly:
Additionally, this PR adds all the AST tests as precompiled artifacts to speed up CI

We can run `python ./tests/test_ast_parsing.py --compile` to generate the artifacts when updating the AST tests

PR #850, opened 2021-05-12
Add support of ERC1155 for slither-check-erc tool.

smonicas:
This PR add the support to check for the correct implementation of [ERC1155](https://eips.ethereum.org/EIPS/eip-1155).


montyly:
This is great, thanks @smonicas for the contribution

Issue #735, opened 2020-12-17
False positives (function modifiers not understood)


aaronovz1:
It seems like Slither doesn't take into account function modifiers?

`onlyOwner` not taken into account:
```
function kill() external onlyOwner {
    selfdestruct(msg.sender);
}
```
`kill() allows anyone to destruct the contract`

`nonReentrant` not taken into account.
Using Openzeppelin ReentrancyGuard and nonReentrant modifier, I still get many reentrancy warnings from Slither.

Could Slither be smarter and analyze function modifiers?

montyly:
Hi @aaronovz1, thanks for reporting this.

For the `onlyOwner`, could you share the code of the modifier? 

For `nonReentrant`, we are aware of the current limitation, and are working on an improvement (https://github.com/crytic/slither/issues/683)

aaronovz1:
Thanks. `onlyOwner` is from standard OpenZeppelin Ownable contract:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol

thedavidmeister:
I see there's a merged branch that is supposed to fix false positives on `nonReentrant` code but i'm getting a lot of detected reentrancy even with the modifiers

https://github.com/rainprotocol/rain-protocol/actions/runs/4067065472/jobs/7003956483

for example

```
Reentrancy in Sale.buy(BuyConfig) (contracts/sale/Sale.sol#554-741):
	External calls:
	- canLive0Store_.set(DEFAULT_STATE_NAMESPACE,canLive0KVs_) (contracts/sale/Sale.sol#574)
	- calculateBuyStore_.set(DEFAULT_STATE_NAMESPACE,calculateBuyKVs_) (contracts/sale/Sale.sol#613-616)
	State variables written after the call(s):
	- remainingTokenInventory = context_[CONTEXT_BUY_COLUMN][CONTEXT_BUY_TOKEN_BALANCE_AFTER_ROW] (contracts/sale/Sale.sol#664-666)
	Sale.remainingTokenInventory (contracts/sale/Sale.sol#208) can be used in cross function reentrancies:
	- Sale._end() (contracts/sale/Sale.sol#414-429)
	- Sale._previewCanLive() (contracts/sale/Sale.sol#382-407)
	- Sale.initialize(SaleConfig,SaleRedeemableERC20Config) (contracts/sale/Sale.sol#261-339)
[...]

Issue #683, opened 2020-11-10
Reentrancy: remove false alarm due to the nonreentrant modifier


montyly:
We could remove the results if:
- The function has the `nonreentrant` modifier
- All the other functions that don't have the modifier cannot be exploited by this reentrancy

snd:
this would be great. slither generates a lot of false positives for projects make heavy use of `nonReentrant`

PR #854, opened 2021-05-18
[Draft] Allow to list modifiers that are being used to protect against reentrancy to avoid fake positive

alexon1234:
Hi, 

Thinking about the fake positive when running the reentrancy detectors, I thought that an easy what would be to enable the users to list all the modifiers they use to avoid this exploit.

This is pretty straightforward and could be a way to work around until we can come with something smarter than can detect with modifiers that are actually protecting against reentrancy attacks.

Related to issues #683, #735

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -8,9 +8,10 @@ import json
 import logging
 import os
 import pstats
+from slither.detectors.reentrancy.reentrancy import Reentrancy
 import sys
 import traceback
-from typing import Optional
+from typing import List, Optional
 
 from pkg_resources import iter_entry_points, require
 
@@ -64,7 +65,7 @@ def process_single(target, args, detector_classes, printer_classes):
         ast = "--ast-json"
     slither = Slither(target, ast_format=ast, **vars(args))
 
-    return _process(slither, detector_classes, printer_classes)
+    return _process(slither, detector_classes, printer_classes, args)
 
 
 def process_all(target, args, detector_classes, printer_classes):
@@ -92,7 +93,7 @@ def process_all(target, args, detector_classes, printer_classes):
     )
 
 
-def _process(slither, detector_classes, printer_classes):
+def _process(slither: Slither, detector_classes, printer_classes, args):
     for detector_cls in detector_classes:
         slither.register_detector(detector_cls)
 
@@ -103,6 +104,14 @@ def _process(slither, detector_classes, printer_classes):
 
     results_detectors = []
     results_printers = []
+
+    # We add the modifier guard to all the reentrancy detectors 
+    for detector in slither.detectors:
+        if isinstance(detector, Reentrancy):
+            for modifier in args.guard_modifiers.split(","):
+                detector.add_guard(modifier)
+                
+
 
     if not printer_classes:
         detector_results = slither.run_detectors()
@@ -314,6 +323,14 @@ def parse_args(detector_classes, printer_classes):  # pylint: disable=too-many-s
         action=ListDetectors,
         nargs=0,
         default=False,
+    )
+
+    group_detector.add_argument(
+        "--include-guard-modifiers",
+        help="Comma-separated list of modifiers that avoid reentrancy, default nonReentrant,onlyOwner",
+        action="store",
+        dest="guard_modifiers",
+        default=defaults_flag_in_config["guard_modifiers"],
     )
 
     group_printer.add_argument(
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -11,7 +11,7 @@ import pstats
 from slither.detectors.reentrancy.reentrancy import Reentrancy
 import sys
 import traceback
-from typing import List, Optional
+from typing import Optional
 
 from pkg_resources import iter_entry_points, require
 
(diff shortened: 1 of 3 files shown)


montyly:
Hi @alexon1234,

Thank you for the contribution. This is a nice solution!

I am not sure we should have `nonReentrant` as a default. the reasoning is that it's not because a function is protected with `nonReentrant` that the reentrancy is not exploitable, and you must review all the entry points. 

For example, consider:
```solidity
function withdraw(uint amount) nonReentrant public{
     require(amount <= balances[msg.sender]);
     Receiver(msg.sender).send_funds({value: amount});
     balances[msg.sender] -= amount;
}

function withdraw_all() public{
     uint amount = balances[msg.sender];
     balances[msg.sender] = 0;
     Receiver(msg.sender).send_funds({value: amount});
}
```

Here there is a reentrancy pattern in `withdraw`, but because `withdraw_all` is not protected, the reentrancy can be exploited through `withdraw_all`. So even with `nonReentrant` the reentrancy must be carefully reviewed.

I will think about this a bit. We have a [`modifiers` printer](https://github.com/crytic/slither/wiki/Printer-documentation#modifiers) that can help to detect these corner cases, but maybe we could specialize it.

alexon1234:
> I am not sure we should have nonReentrant as a default. the reasoning is that it's not because a function is protected with nonReentrant that the reentrancy is not exploitable, and you must review all the entry points.

I agree, maybe having default modifiers is not the best approach. 

Related to
> Here there is a reentrancy pattern in withdraw, but because withdraw_all is not protected, the reentrancy can be exploited through withdraw_all. So even with nonReentrant the reentrancy must be carefully reviewed.

`withdraw_all ` can suffer from reentrancy, but because is using checks effects interactions pattern, no funds could be still, right? 
When running the current version of Slither over:

```
pragma solidity 0.8.3;

contract WithDraw {
  mapping(address => uint256) balances;
  modifier nonReentrant() {
    _;
  }

  function withdraw(uint256 amount) public nonReentrant {
    require(amount <= balances[msg.sender]);
    msg.sender.call{ value: amount }("");
    balances[msg.sender] -= amount;
  }

  function withdraw_all() public {
    uint256 amount = balances[msg.sender];
    balances[msg.sender] = 0;
    msg.sender.call{ value: amount }("");
  }
}
```

Slither does not inform about the reentrancy in `withdraw_all`

```
Reentrancy in WithDraw.withdraw(uint256) (withdraw.sol#9-13):
        External calls:
        - msg.sender.call{value: amount}() (withdraw.sol#11)
        State variables written after the call(s):
        - balances[msg.sender] -= amount (withdraw.sol#12)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#reentrancy-vulnerabilities
INFO:Detectors:
WithDraw.withdraw(uint256) (withdraw.sol#9-13) ignores return value by msg.sender.call{value: amount}() (withdraw.sol#11)
WithDraw.withdraw_all() (withdraw.sol#15-19) ignores return value by msg.sender.call{value: amount}() (withdraw.sol#18)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unchecked-low-level-calls
INFO:Detectors:
Pragma version0.8.3 (withdraw.sol#1) necessitates a version too recent to be trusted. Consider deploying with 0.6.12/0.7.6
solc-0.8.3 is not recommended for deployment
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity
INFO:Detectors:
Low level call in WithDraw.withdraw(uint256) (withdraw.sol#9-13):
        - msg.sender.call{value: amount}() (withdraw.sol#11)
Low level call in WithDraw.withdraw_all() (withdraw.sol#15-19):
        - msg.sender.call{value: amount}() (withdraw.sol#18)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#low-level-calls
INFO:Detectors:
Function WithDraw.withdraw_all() (withdraw.sol#15-19) is not in mixedCase
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#conformance-to-solidity-naming-conventions
INFO:Detectors:
withdraw(uint256) should be declared external:
        - WithDraw.withdraw(uint256) (withdraw.sol#9-13)
withdraw_all() should be declared external:
        - WithDraw.withdraw_all() (withdraw.sol#15-19)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-external
```

montyly:
You can call `withdraw`, and reenter into `withdraw_all` ;)

The reentrancy is in `withdraw`, `nonReentrant` prevents you from re-entering directly into this function, but you can still re-enter into other entry points

alexon1234:
Yes, you're right! I will remove the default modifiers, but do you think we could add a feature like this draft?

montyly:
I think it's a great feature @alexon1234 and it should be included in the next release. Can you sign the cla, so that we can merge the PR?

riordant:
Hi guys, just thinking about this problem too. Is it only an issue if `nonRentrant` does not exist on other state-changing functions? In that case, if the modifier is detected on the re-entrant function with `nonRentrant`, can we just check for the same modifer on the other state changing functions?

montyly:
Hi @riordant. 

That's the idea for https://github.com/crytic/slither/issues/683, we can check if the variables affected are read/write in other functions without the modifier

montyly:
As the CLA was never signed, we can't merge this PR. Because this was done in 2021, I am closing the PR.

For info, we have ongoing discussion on how to filter better the results from the reentrancy to prevent FP due to the non reentrant modifier, and improvements should be added soon.

PR #859, opened 2021-06-02
updated reference wiki address

enderphan94:
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#function-initializing-state-variables is invalid


montyly:
Thanks @enderphan94, it's a good catch

PR #862, opened 2021-06-06
add `region`s to enable folding support in editors

hacker-DOM:
Before: https://i.imgur.com/mMuHwyn.png - Multi-line strings break editor folding 

After: https://i.imgur.com/4nFuvl8.png - Folding works!

More context:

In theory the Python language server could provide semantic folding support. However, it seems this [is not a priority for e.g. Pylance (VS code LPSP)](https://github.com/microsoft/pylance-release/issues/372).


montyly:
Thanks @hacker-DOM !

PR #869, opened 2021-06-20
Fix assignment to type in class Contract

hacker-DOM:


montyly:
Good catch, thanks @hacker-DOM

PR #876, opened 2021-06-23
Updated WIKI reference link

enderphan94:
Updated WIKI reference link for unused-state-variable


enderphan94:
@montyly can you check?

PR #879, opened 2021-07-01
move function type assignment from _analyze_attributes to _analyze_type

hacker-DOM:
This moves `_function_type` assignment from `_analyze_attributes` to `_analyze_type`. I am fairly confident this doesn't break anything, because:
- `_analyze_type` is called exactly once in the codebase on `FunctionSolc.__init__`.
- `_analyze_attributes` is called exactly once for a function or modifier in `FunctionSolc|ModifierSolc.analyze_params()`. This is called in:
  - `FunctionSolc.analyze_params()`:
    - `_analyze_params_top_level_function`, which is a node in `_analyze_third_part`
  - Both:
    - `ContractParser._analyze_params_elements`, which is also a node in `_analyze_third_part`

Based on this analysis, it seems to me there will be no unintended side-effects from setting the function type correctly early

### DIFF
--- a/slither/solc_parsing/declarations/function.py
+++ b/slither/solc_parsing/declarations/function.py
@@ -208,6 +208,13 @@ class FunctionSolc:
             if self._function.name == self._function.contract_declarer.name:
                 self._function.function_type = FunctionType.CONSTRUCTOR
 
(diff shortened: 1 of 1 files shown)


montyly:
I am closing this PR, as it hasn't been updated since 2021, and did not pass the CI

PR #908, opened 2021-08-11
Removes crytic.io notice

noahlitvin:


montyly:
Hi @noahlitvin .

Nice catch, thanks for the PR

PR #920, opened 2021-08-20
Add detector use of msg.value inside a loop

smonicas:
This PR add a new detector to check for the usage of msg.value inside a loop

### DIFF
--- a/slither/detectors/statements/msg_value_in_loop.py
+++ b/slither/detectors/statements/msg_value_in_loop.py
@@ -1,0 +1,84 @@
+from slither.core.cfg.node import NodeType
+from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
+from slither.core.declarations import SolidityVariableComposed
+
+
+def detect_msg_value_in_loop(contract):
+    results = []
+    for f in contract.functions + contract.modifiers:
+        if f.is_implemented and f.payable:
+            msg_value_in_loop(f.entry_point, False, [], results)
+    return results
+
+
+def msg_value_in_loop(node, in_loop, visited, results):
+    if node in visited:
+        return
+    # shared visited
+    visited.append(node)
+
+    if node.type == NodeType.STARTLOOP:
+        in_loop = True
+    elif node.type == NodeType.ENDLOOP:
+        in_loop = False
+
+    if in_loop:
+        for ir in node.all_slithir_operations():
+            if SolidityVariableComposed("msg.value") in ir.read:
+                results.append(node)
+
+    for son in node.sons:
+        msg_value_in_loop(son, in_loop, visited, results)
+
+
+class MsgValueInLoop(AbstractDetector):
+    """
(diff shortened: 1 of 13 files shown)


slither/detectors/statements/msg_value_in_loop.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if node.type == NodeType.STARTLOOP:
+        in_loop = True
+    elif node.type == NodeType.ENDLOOP:
+        in_loop = False
+
+    if in_loop:
+        for ir in node.all_slithir_operations():
+            if SolidityVariableComposed("msg.value") in ir.read:
+                results.append(node)
```
montyly, line 28: Open question: should we use `node` or `ir.node` here?

If the `msg.value` usage is within an internal call, `node` gives us the internal call, while `ir.node` gives us the actual msg.value usage.

slither/detectors/statements/msg_value_in_loop.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def msg_value_in_loop(node, in_loop, visited, results):
+    if node in visited:
+        return
+    # shared visited
+    visited.append(node)
+
+    if node.type == NodeType.STARTLOOP:
+        in_loop = True
```
montyly, line 21: The true/false might not work with nested loops. Maybe we could use a counter instead?

I think this pattern is used in other detectors, so we could improve the other detectors also

montyly:
The current logic:
- Iterate over all the functions
- If the function is payable, look for the bug (a node in a loop that use `msg.value`)

This approach works, but has some limits, mostly, because `the function is payable` only works for public/external functions. For example, if the bad loop is in an internal function, the detector won't catch it:

```solidity
contract C{
    
    mapping (address => uint256) balances;

    function bad(address[] memory receivers) public payable{
        bad_intern(receivers);
    }   
    
    function bad_intern(address[] memory receivers) internal{
        for (uint256 i = 0; i < receivers.length; i++) {
            bad2_internal(receivers[i]);
        }
    }
    function bad2_internal(address a) internal {
        balances[a] += msg.value;
    }
    
}   
```

Maybe we could change a bit the logic to:
- Iterate over [`contract.functions_entry_points`](https://github.com/crytic/slither/blob/903799c02fb100803b4efb7295ce7b024af85fe6/slither/core/declarations/contract.py#L442), and take only the payable function. Note that here we don't need to iterate over the modifiers
- Have `msg_value_in_loop` go over the function's entry point
- If there is an internal call (`isinstance(ir, InternalCall)`, [InternalCall](https://github.com/crytic/slither/blob/master/slither/slithir/operations/internal_call.py)), then call `msg_value_in_loop` on the destination entry point (`ir.function`))


What do you think? Note that if there is another detector that has the same flaw, we can take the opportunity to improve it

I think the same comments apply for https://github.com/crytic/slither/pull/921

Issue #793, opened 2021-02-24
Add tool to read on-chain storage


montyly:
We can rely on the internal storage information and [`web3.get_storage_at`](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_storage_at) to return the value of a contract onchain

Example of API

- `slither-read-storage [codebase] 0x414141 canonical_variable_name`
- `slither-read-storage 0x414141 canonical_variable_name` (if the source code is on etherscan)

Potential flags:
- RPC-related to connect to local node, infrura or other
- --list-all to dump all the contract variables and their values (aside mapping and maybe array?)
- `--offset` / `--index` for mapping array

Elyx0:
Came from https://0xalpharush.github.io/slithering-through-the-dark-forest/ excited to see it!

jeffchen006:
Redirect interested people to https://github.com/crytic/slither/tree/master/slither/tools/read_storage

PR #931, opened 2021-08-27
core/solidity_types: fix storage size and min size for bytes

0xalpharush:
When a contract contains a state variable such as `bytes32[3] data`, calling `data.type.storage_size` [returns](https://github.com/crytic/slither/blob/master/slither/core/solidity_types/array_type.py#L44) `12` bytes (`data.type` is `ArrayType`). This is because the size of each element in the array is [calculated](https://github.com/crytic/slither/blob/master/slither/core/solidity_types/elementary_type.py#L201) to be `4` bytes instead of `32` bytes. Changing the [size property](https://github.com/crytic/slither/blob/master/slither/core/solidity_types/elementary_type.py#L192) to return the number of bits (eg. 256) instead of bytes (eg. 32) fixes this issue, and `data.type.storage_size` correctly evaluates to `96` bytes, or three storage slots. I believe this is the desired behavior as `uint256[3] data` reports a size of `96` bytes currently. I ran into this error while working on [this issue](https://github.com/crytic/slither/issues/793) and using the size and offset to parse values from storage.

### DIFF
--- a/slither/core/solidity_types/elementary_type.py
+++ b/slither/core/solidity_types/elementary_type.py
@@ -127,10 +127,10 @@ Max_Byte = {k: 2 ** (8 * (i + 1)) - 1 for i, k in enumerate(Byte[2:])}
 Max_Byte["bytes"] = None
 Max_Byte["string"] = None
 Max_Byte["byte"] = 255
-Min_Byte = {k: 1 << (4 + 8 * i) for i, k in enumerate(Byte[2:])}
+Min_Byte = {k: 0 for k in Byte}
 Min_Byte["bytes"] = 0x0
-Min_Byte["string"] = None
-Min_Byte["byte"] = 0x10
+Min_Byte["string"] = 0x0
+Min_Byte["byte"] = 0x0
 
 MaxValues = dict(dict(Max_Int, **Max_Uint), **Max_Byte)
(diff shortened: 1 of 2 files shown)


slither/core/solidity_types/elementary_type.py
```diff
@@ -189,7 +189,7 @@ def size(self) -> Optional[int]:
         if t == "address":
             return int(160)
         if t.startswith("bytes"):
-            return int(t[len("bytes") :])
+            return int(t[len("bytes") :]) * 8
```
axic, line 192: Will this just result in 0 for `bytes`? Shouldn't it return `None` for that?

0xalpharush: Thanks for alerting me of this. We will need to check that the type is not exactly equal to `bytes` and then it will return `None`. 

```
if t.startswith("bytes") and t != "bytes":
```

Currently, Slither parses `bytes` (dynamic array in storage) as an instance of `ElementaryType`. Should it instead be an `ArrayType` like `uint8[]`?

axic: I'm just a person reporting bugs on this repository, so I can't really help whether this is good/bad/desired/undesired, but it did look unclear to me. I do see that `storage_size` has its own pre-checks though.

0xalpharush:
Replaced by #968

PR #935, opened 2021-09-04
Handle ignore markers when `exclude-dependencies` is set

ilya-bobyr:
`valid_results` behaviour is described in terms of "results are valid,
unless ...".  A matching implementation would be to only return `True` a
the end of the function, and have all the cases only return `False` if
they detect an invalid result.

`show-ignored-findings` is the only exception to the rule above.


montyly:
Hi @ilya-bobyr. Thanks for the PR. We (finally) merged it with #1317.

Issue #937, opened 2021-09-09
Dockerizing latest for Apple M1 Silicon


daithi-coombes:
I think the `Dockerfile` needs looking at, hasn't been updated since 2 Sep 2020. I'm working on an Apple Silicon M1 chip (arm64) and hitting issues.

Here's what I've done and why
  1. upgraded base ubuntu to xenial (no maifest for arm64 with bionic)
  2. install pip3 in `Dockerfile` to install `crytic-compile` and `prettytable` (`RUN ... setup.py...` step couldn't find them)
  3. now builds but with error:

```
% docker run --rm daithi/slither:local slither --help
Traceback (most recent call last):
  File "/home/slither/.local/bin/slither", line 9, in <module>
    load_entry_point('slither-analyzer==0.8.1', 'console_scripts', 'slither')()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/home/slither/.local/lib/python3.5/site-packages/slither_analyzer-0.8.1-py3.5.egg/slither/__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 954, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 896, in _find_spec
  File "<frozen importlib._bootstrap_external>", line 1139, in find_spec
  File "<frozen importlib._bootstrap_external>", line 1115, in _get_spec
  File "<frozen importlib._bootstrap_external>", line 1096, in _legacy_get_spec
  File "<frozen importlib._bootstrap>", line 444, in spec_from_loader
[...]

PR #938, opened 2021-09-09
Dockerizing latest for Apple M1 Silicon

daithi-coombes:
related issue: https://github.com/crytic/slither/issues/937

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:xenial
 
 LABEL name=slither
 LABEL src="https://github.com/trailofbits/slither"
@@ -8,7 +8,14 @@ LABEL desc="Static Analyzer for Solidity"
 
 RUN apt-get update \
   && apt-get upgrade -y \
-  && apt-get install -y git python3 python3-setuptools wget software-properties-common
+  && apt-get install -y build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev libsqlite3-dev wget libbz2-dev
+
+RUN wget https://www.python.org/ftp/python/3.8.12/Python-3.8.12.tgz \
+  && tar -xf Python-3.8.12.tgz \
+  && cd Python-3.8.12 \
+  && ./configure --enable-optimizations \
+  && make -j 8 \
+  && make altinstall
 
 RUN wget https://github.com/ethereum/solidity/releases/download/v0.4.25/solc-static-linux \
  && chmod +x solc-static-linux \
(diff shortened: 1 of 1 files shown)


daithi-coombes:
Some issues with Github Actions:

```
  slither --help
  shell: sh -e {0}
Traceback (most recent call last):
  File "/home/slither/.local/bin/slither", line 33, in <module>
    sys.exit(load_entry_point('slither-analyzer==0.8.1', 'console_scripts', 'slither')())
  File "/home/slither/.local/bin/slither", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: slither-analyzer
Error: Process completed with exit code 1.
```

Related yaml:
```yaml
  sec-slither:
    runs-on: ubuntu-latest
    container:
      image: daithi/slither:latest
    steps:
      - uses: actions/checkout@master
      - run: echo "${HOME}/.local/bin" >> $GITHUB_PATH
      - run: slither --help
```

`docker run --rm daithi/slither slither --help` works
testing the gh yaml with [act](https://github.com/nektos/act) also works.

I'm thinking maybe some linking is missing. `$HOME/.local/bin` is in the path (checked also in gh workflow) but doesn't have `slither-analyzer`

montyly:
@elopez : can you take a look at this? I haven't followed the progress on M1, and I am not sure we still need to update this

elopez:
Hey @montyly! Slither hasn't had Docker images built for a while as far as I can tell (last I see is ~3y ago, and the slither README suggests using eth-security-toolbox and not this dockerfile), so I don't know if anyone is using this 🤔

Personally I'd prefer to not to build Python from scratch as it is done in this PR, as that involves adding extra dependencies and build products which make the image heavier and the dockerfile more complex. Updating the base image to a modern ubuntu (like 20.04 or 22.04) so that we have newer python (and maybe adding some build tools for pysha3) should be enough to get M1 going.

NB: solc is also not available for aarch64 linux, I'm guessing this dockerfile would only work on M1 Macs with the emulation, and not real linux aarch64. If we're revamping this dockerfile, it might be a good time to use solc-select instead of an old solc binary 😄

montyly:
I agree with @elopez , let's simplify our docker and switch to solc-select (https://github.com/crytic/slither/issues/1330).

Thanks [daithi-coombes](https://github.com/daithi-coombes) for the work; we will do the improvements in a separate PR

Issue #580, opened 2020-08-18
assert patch_type == 'line_removal' failed error when running slither-flat in the Balancer contracts 


ggrieco-tob:
The latest revision of the Balancer contracts (https://github.com/balancer-labs/balancer-core/commit/f4ed5d65362a8d6cec21662fb6eae233b0babc1f) fails to converted by slither-flat (latest release and latest revision from `master`):

```
$ slither-flat . --convert-external
ERROR:CryticCompile:- Fetching solc version list from solc-bin. Attempt #1

Traceback (most recent call last):
  File "/home/g/.local/bin/slither-flat", line 11, in <module>
    load_entry_point('slither-analyzer==0.6.12', 'console_scripts', 'slither-flat')()
  File "/home/g/.local/lib/python3.6/site-packages/slither/tools/flattening/__main__.py", line 52, in main
    private_to_internal=args.convert_private)
  File "/home/g/.local/lib/python3.6/site-packages/slither/tools/flattening/flattening.py", line 37, in __init__
    self._get_source_code(contract)
  File "/home/g/.local/lib/python3.6/site-packages/slither/tools/flattening/flattening.py", line 112, in _get_source_code
    assert patch_type == 'line_removal'
AssertionError
```

PR #964, opened 2021-10-15
Fix slither-flat external/private options

bearpebble:
The options --convert-external and --convert-private are currently
broken from what looks like some automatic conversion according to git
blame.
Also fixes issue #580


montyly:
Thanks @bearpebble !

Issue #932, opened 2021-08-30
Create thorough tests for storage_size


montyly:
We can rely on the solc storage output (from solc >=0.6.5) to test our storage size API

It will help us to catch bugs such that https://github.com/crytic/slither/pull/931

PR #968, opened 2021-10-28
WIP: feature to read contract storage data

0xalpharush:
Closes https://github.com/crytic/slither/issues/793 and https://github.com/crytic/slither/issues/932
- [x] state variables
- [x] read from array at index/ key
- [x] support mapping
- [x] support deep mapping
- [x] support mapping with structs
- [x] add tests https://github.com/crytic/slither/issues/932
- [x] add ci tests for tool
- [x] test multidimensional arrays
- [x] Refactor imports so web3 isn't a dependency
- [x] Fix dynamic and fixed-size arrays
- [ ] Parse enums?
- [ ] Handle string/ bytes values greater than a single word
- [x] Add support for `ContractStructure` types
- [x] add documentation of CLI and API


setup.py
```diff
@@ -11,8 +11,9 @@
     install_requires=[
         "prettytable>=0.7.2",
         "pysha3>=1.0.2",
-        # "crytic-compile>=0.2.1",
+        #"crytic-compile>=0.2.1",
         "crytic-compile",
+        "web3",
```
montyly, line 16: I think we should not make this a permanent dep. `get_storage_slot_and_val` is already using a try/catch approach, the only other place where we used it is `get_storage_data`

We can probably move the `try/catch` import at the `read_storage.py` top-level, and have the tool not loading at all if the dep is not installed (see the comment on read_storage.py)

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,309 @@
+import sys
+import logging
+from math import floor
+
+from hexbytes import HexBytes
```
montyly, line 5: A lot of dependencies in this file are not present in the setup.py (including hexbytes, all the eth_*)

Maybe we could group them into an import statement, like

```python
def load_dependencies():
   try:
    # Try to import every dep and
    # print a message if it's not installed


hexbytes, eth_utils,  eth_utils= load_dependencies()


```

0xalpharush: They're web3 dependencies so they'll be installed with it. I will follow this suggestion

slither/tools/read_storage/__main__.py
```diff
@@ -0,0 +1,112 @@
...
+"""
+import argparse
+
+from crytic_compile import cryticparser
+
+from slither import Slither
+from slither.tools.read_storage.read_storage import get_storage_layout, get_storage_slot_and_val
+
+
+def parse_args():
```
montyly, line 12: Add return type

slither/tools/read_storage/__main__.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+
+    parser.add_argument("--max-depth", help="Max depth to search in data structure", default=20)
+
+    cryticparser.init(parser)
+
+    return parser.parse_args()
+
+
+def main():
```
montyly, line 84: Add return type

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,521 @@
...
+logger = logging.getLogger("Slither-read-storage")
+logger.setLevel(logging.INFO)
+
+
+class SlitherReadStorageException(Exception):
+    pass
+
+
+# pylint: disable=too-many-arguments
+def _all_struct_variables(var, contracts, address, rpc_url, storage_address=None, key=None):
```
montyly, line 43: Add parameters and return types

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,521 @@
...
+            key=key,
+            struct_var=elem.name,
+        )
+        data[elem.name] = {"slot": slot, "value": val, "type_string": type_string}
+
+    return data
+
+
+# pylint: disable=too-many-branches,too-many-locals,too-many-nested-blocks
+def get_storage_layout(
```
montyly, line 66: Add return type

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,521 @@
...
+            storage_address,
+            key=key,
+            struct_var=elem.name,
+        )
+        data[elem.name] = {"slot": slot, "value": val, "type_string": type_string}
+
+    return data
+
+
+# pylint: disable=too-many-branches,too-many-locals,too-many-nested-blocks
```
montyly, line 65: I think we can split this function into sub-functions to not remove all these warnings

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,521 @@
...
+
+                tmp[var_name]["elems"] = elems
+
+            data[contract] = tmp
+
+    with open(f"{address}_storage_layout.json", "w", encoding="utf-8") as f:
+        json.dump(data, f, indent=4)
+
+
+# pylint: disable=too-many-statements,too-many-branches,inconsistent-return-statements
```
montyly, line 161: Similarly, I think we can split this function into sub-functions to not remove all these warnings

slither/tools/read_storage/read_storage.py
```diff
@@ -0,0 +1,521 @@
...
+            # Account for storage packing
+            offset_hex_bytes = get_offset_value(hex_bytes, offset, size)
+            value = coerce_type(type_to, offset_hex_bytes)
+
+            log += f"\nName: {var_log_name}\nType: {type_to}\nValue: {value}\nSlot: {int.from_bytes(slot, byteorder='big')}\n"
+            logger.info(log)
+
+            return int_slot, value, type_to
+
+    if not found:
```
montyly, line 304: It looks like we can remove `found` and raise a exception if we reached this part of the code, is that correct?

montyly: that would also avoid having an undefined return

PR #973, opened 2021-11-08
fixed call-graph printer output filename

dev1644:
It fixes the filename of diagrams generated by `call-graph` printer.

Before: `..all_contracts.call-graph.dot`
After: `all_contracts.call-graph.dot`

### DIFF
--- a/slither/printers/call/call_graph.py
+++ b/slither/printers/call/call_graph.py
@@ -233,10 +233,10 @@ class PrinterCallGraph(AbstractPrinter):
             info += f"Call Graph: {all_contracts_filename}\n"
 
             # Avoid duplicate functions due to different compilation unit
-            all_functions = [
+            all_functionss = [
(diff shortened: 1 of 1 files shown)


slither/printers/call/call_graph.py
```diff
@@ -227,11 +232,11 @@ def output(self, filename):
         with open(all_contracts_filename, "w", encoding="utf8") as f:
             info += f"Call Graph: {all_contracts_filename}\n"
 
-            # Avoid dupplicate funcitons due to different compilation unit
-            all_functionss = [
+            # Avoid duplicate functions due to different compilation unit
+            all_functions = [
```
montyly, line 236: So actually the double `s` is here because if the same variable is used twice, it will have two different types (list of list of functions, versus a list of functions).

Using `ss` is a shortcut to have two variables without having a lot of different names and it is used in different parts of the codebase, but it's not actually documented

montyly (changes requested):

PR #986, opened 2021-12-01
Update broken wiki link for slither python API

0xalpharush:


montyly:
Good catch @0xalpharush

PR #988, opened 2021-12-02
Validate input from --markdown-root flag

### LINKED ISSUE #868: Sanitization of input from `--markdown-root` flag
Check, sanitize (if necessary) & warn for the input provided through `--markdown-root`  for following cases -

- Check whether the url contains `tree` keyword in the git link instead of `blob`, print a warning regarding it.
- If forward slash (`/`) is missing at the end of the input, append it and print a warning.

dev1644:
Closes #868

### DIFF
--- a/slither/utils/command_line.py
+++ b/slither/utils/command_line.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import logging
 from collections import defaultdict
 from crytic_compile.cryticparser.defaults import (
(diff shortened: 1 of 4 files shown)


slither/utils/command_line.py
```diff
@@ -332,3 +333,29 @@ def output_printers_json(printer_classes):
         table.append({"index": idx, "check": argument, "title": help_info})
         idx = idx + 1
     return table
+
+
+def check_and_sanitize_markdown_root(markdown_root: str) -> str:
+    # Regex to check whether the markdown_root is a GitHub URL
+    match = re.search(
+        r"(https://)github.com/([a-zA-z-]+)([:/][A-Za-z0-9_.-]+[:/]?)([A-Za-z0-9_.-]*)(.*)",
```
montyly, line 341: Lgtm is complaining `This regular expression includes duplicate character 'z' in a set of characters`,  looks like it is right and there is `z` that can be removed from the regex right?

dev1644: Hmm right, there seems to be a typo `A-Z` resulting in incorrect pattern matching of org/profile name of the URL. 
Fixed it.

montyly:
Thanks @dev1644

PR #1013, opened 2022-01-07
Added Additional CMD Flag for Specifying Custom Solc Binary

epociask:
**Changes:**
- Added new string argument to arg parser: `--solc-binary`
-  If flag not specified, defaults to `solc` for compiler command value


montyly:
Hi @epociask, thank you for your contribution.

I believe that this feature already exists with the following flag:
```
Solc options:
  --solc SOLC           solc path
```

That slither inherits from [crytic-compile](https://github.com/crytic/crytic-compile).

epociask:
Whoops, you are correct 😄 .  Closing this out!

PR #1018, opened 2022-01-20
Update __main__.py

zjzh:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/slither/tools/erc_conformance/__main__.py
+++ b/slither/tools/erc_conformance/__main__.py
@@ -89,7 +89,7 @@ def main():
         contract = contracts[0]
         # First elem is the function, second is the event
         erc = ERCS[args.erc.upper()]
-        generic_erc_checks(contract, erc[0], erc[1], ret)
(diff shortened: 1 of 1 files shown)


disconnect3d:
And now people have to use their brains to think what does `*something` or `something[0:2:1]` mean in `generic_erc_checks(contract, *erc[0:2:1], ret)`.

This is not more pythonic IMHO: it exploits cool Python features, but pure `erc[0]` and `erc[1]` is much more readable.

If I would do any change here to make this more readable, I would maybe change this to:

```python
a, b = ERCS[args.erc.upper()]
generic_erc_checks(contract, a, b, ret)
```

Providing some meaningful names to `a` and `b` of course; or, if there are more items in the `ERCS` values then:
```python
a, b, *others = ERCS[args.erc.upper()]
generic_erc_checks(contract, a, b, ret)
```

Of course, probably providing a meaningful name for `others` as well.

PR #1020, opened 2022-01-22
ElementaryType: `size` for `bytes` returns bit width

ilya-bobyr:
For `unitXX`, `intXX`, `bool` and `address` types `size` property returns
size in bits.  But for `bytesXX` type it was returning size in bytes.  It
makes calculation in `storage_size` break, as well as makes comparing
`size` values for different types more tricky.

Also, `bytes` (without any width) is a valid variable width type.  It
should return `None`.

### DIFF
--- a/slither/core/solidity_types/elementary_type.py
+++ b/slither/core/solidity_types/elementary_type.py
@@ -188,8 +188,8 @@ class ElementaryType(Type):
             return int(8)
         if t == "address":
             return int(160)
-        if t.startswith("bytes"):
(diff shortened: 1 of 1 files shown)


0xalpharush:
Thanks for taking the time to submit a PR! We'll merge PR #968 soon which addresses this issue https://github.com/crytic/slither/pull/968/files#diff-d7d3929d267223881511c461c1bc7a614a6b2fc2bf494d039821267c701c6ceeR191-R192

Issue #1010, opened 2022-01-04
Filter external visibility recommendation for public functions to solc < 0.6.9


0xalpharush:
Update this [detector](https://github.com/crytic/slither/blob/master/slither/detectors/functions/external_function.py) to filter based on the compiler version. Also, only output warning if one of the parameters is memory.
Ref: https://ethereum.stackexchange.com/questions/107578/does-using-external-over-public-in-a-library-reduce-any-gas-costs

htadashi:
Working on this issue. Thanks for the clarification @0xalpharush :)

0xalpharush:
`Prior to version 0.6.9 data location for reference-type arguments was limited to calldata in external functions, memory in public functions`. Based off the docs you linked, it is possible to have a public function with memory params.
 Heuristic: If a public function is not called internally and has memory params, recommend that the user change the function to external visibility and the param data location to calldata.

htadashi:
Ops, I wrote it wrong - sorry for the confusion. I meant to write that it is impossible to have a public function with `calldata` parameters (prior to 0.6.9). Thus public functions prior to 0.6.9 always have memory params and we don't need to check if its parameters are declared as `memory`.

Checking if a memory parameter can be changed to calldata is already detected by this line, isn't it?
https://github.com/crytic/slither/blob/168492ad899e074724b97f99cfa4d4f25934a8e5/slither/detectors/functions/external_function.py#L142-L145

duckki:
Ultimately, it depends on the `pragma solidity` setting of the file. I wouldn't suppress this report unless the lowest compatible Solidity version is 0.6.9 or up (not the version of current compiler being used).

PR #1022, opened 2022-01-24
Filter recommendation to < 0.6.9 and adds information about calldata parameter

htadashi:
Solves #1010. Needs to add more tests and update the tests for checking against false-positives.


0xalpharush:
Can you add tests [here](https://github.com/crytic/slither/tree/master/tests/detectors/external-function)? Good tests should not recommend this optimization and "bad" ones should. So, for versions of solidity greater than 0.6.9 all tests should be good. We should have good functions that have memory args or do not use memory types in lower versions of solidity that are not detected as well.

0xalpharush:
Any way we can help you out?

htadashi:
Hey @0xalpharush, thanks for the help and sorry for the late reply! Currently I'm moving to another country so my time got limited, but I'll try to write the tests as soon I get back to my normal routine :)

htadashi:
@0xalpharush Thanks for your guidance and your patience :) 
I created the additional tests in `external_function_3.sol`. By the way, your FAQ in PR #1047 (in particular, the paragraph about updating an existing detector) helped me to understand how the tests work.

Should this detector also detect when `memory` arguments could be changed to `calldata` when solc >= 0.6.9? This seems an independent problem from detecting when the function can be changed from `public` to `external`. For instance, in contract
```
pragma solidity 0.6.11;

contract Test {

   function test(uint256[] memory x) public {
       x[0] = 1;
   }
}
```
the function `test` can be changed from `public` to `external`, but even so, `x` can't be changed to `calldata`.

0xalpharush:
For this PR, we can recommend that a public function with memory args prior to 0.6.9 be changed to an external function with calldata args. If you'd like, you could implement a separate PR that recommends changing memory args to calldata args in public/ external functions if the solidity version 0.6.9 or later.

htadashi:
Sorry for the late reply. I am still adjusting myself to the new country and had some issues with my notebook in the past few days.

Since detecting a public function with memory args before 0.6.9 is equivalent to detecting a public function (a public function with calldata arguments won't compile prior to 0.6.9), I think it is ready for review now.

0xalpharush:
This detector should only flag a function if there is at least one memory arg in the function. In its current state, it flags all external functions even if the args do not have a memory location e.g.: 
```
function test(uint arg) public {
    return uint;
}
```
See https://docs.soliditylang.org/en/v0.6.9/types.html?highlight=data%20location#data-location for all args that can havea data location

slither/detectors/functions/external_function.py
```diff
@@ -108,6 +108,10 @@ def function_parameters_written(function):
     def _detect(self):  # pylint: disable=too-many-locals,too-many-branches
         results = []
 
+        # After solc 0.6.9, calldata arguments are allowed in public functions
+        if self.compilation_unit.solc_version >= "0.7." or self.compilation_unit.solc_version in ["0.6.9","0.6.10","0.6.11"]:
+            return []
```
0xalpharush, line 113:
```suggestion
            return results
```

slither/detectors/functions/external_function.py
```diff
@@ -20,9 +20,9 @@ class ExternalFunction(AbstractDetector):
     WIKI = "https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-external"
 
     WIKI_TITLE = "Public function that could be declared external"
-    WIKI_DESCRIPTION = "`public` functions that are never called by the contract should be declared `external` to save gas."
+    WIKI_DESCRIPTION = "`public` functions that are never called by the contract should be declared `external` and its immutable parameters should be located in `calldata` to save gas."
```
0xalpharush, line 23:
```suggestion
    WIKI_DESCRIPTION = "`public` functions that are never called by the contract should be declared `external`, and its immutable parameters should be located in `calldata` to save gas."
```

htadashi:
@0xalpharush Thanks! I didn't notice this case.

By the way, I was trying to use Slither's `var.location` to detect the location of the args, but it reports `memory` for `int` as well for other non-reference types. Is this the expected behavior?

0xalpharush:
Elementary types are implicitly located memory, so this is expected. Arrays (including bytes and string), mappings, and structs can have a data location of `calldata` declared explicitly. So, you'd want to first detect is this an array,  bytes array, string, mapping, or struct. And then, if it's also in memory, recommend putting it in calldata.

slither/detectors/functions/external_function.py
```diff
@@ -135,6 +151,14 @@ def _detect(self):  # pylint: disable=too-many-locals,too-many-branches
             # Next we'll want to loop through all functions defined directly in this contract.
             for function in contract.functions_declared:
 
+                # If all of the function arguments are non-reference type or calldata, we skip it.        
+                reference_args = []
+                for arg in function.parameters:
+                    if self.is_reference_type(arg) and arg.location == 'memory':
+                        reference_args.append(arg)
+                if len(reference_args) == 0 and len(function.parameters) > 0:
```
0xalpharush, line 159: What's the reasoning behind this check `len(function.parameters) > 0` ?

htadashi: Ops, my mistake! We can skip functions without arguments there too. I removed the check in the next commit.

0xalpharush:
Can you run the lints in `CONTRIBUTING.md` and add test cases that should not be detected. For instance: 
```
function good(bytes calldata x) external {};
```

htadashi:
> Can you run the lints in `CONTRIBUTING.md` and add test cases that should not be detected? For instance:
> 
> ```
> function good(bytes calldata x) external {};
> ```

Added more tests cases in `external_function3.sol` and `pylint`/`black` worked locally in the previous commits :)

slither/detectors/functions/external_function.py
```diff
@@ -105,9 +107,31 @@ def get_all_function_definitions(base_most_function):
...
+            return True
+        if str(parameter.type) in ["bytes", "string"]:
+            return True
+        return False
+
     def _detect(self):  # pylint: disable=too-many-locals,too-many-branches
         results = []
 
+        # After solc 0.6.9, calldata arguments are allowed in public functions
+        if self.compilation_unit.solc_version >= "0.7." or self.compilation_unit.solc_version in [
```
0xalpharush, line 128: I think there likely needs to be a lower bound since data location doesn't exist in 0.4.25, for instance.

slither/detectors/functions/external_function.py
```diff
@@ -105,9 +107,31 @@ def get_all_function_definitions(base_most_function):
     def function_parameters_written(function):
         return any(p in function.variables_written for p in function.parameters)
 
+    @staticmethod
+    def is_reference_type(parameter):
+        if isinstance(parameter.type, ArrayType):
```
0xalpharush, line 112: Can you add tests that hit each of these conditions i.e. an array (mapping etc.) in storage should not be detected but a memory one should?

0xalpharush:
If you rebase against the latest dev branch, I think that will fix the lint issues.

slither/detectors/functions/external_function.py
htadashi: If I understood correctly, it does [exist](https://docs.soliditylang.org/en/v0.4.25/types.html#:~:text=There%20is%20also%20a%20third%20data%20location%2C%20calldata%2C%20which%20is%20a%20non%2Dmodifiable%2C%20non%2Dpersistent%20area%20where%20function%20arguments%20are%20stored.%20Function%20parameters%20(not%20return%20parameters)%20of%20external%20functions%20are%20forced%20to%20calldata%20and%20behave%20mostly%20like%20memory.), but I think the `calldata` keyword was added later (probably in 0.5.0). 

So I think it would be nice to still detect if a function can be changed from external to public in these contexts, since this would make the compiler force the appropriate parameters to be located in calldata.

slither/detectors/functions/external_function.py
htadashi: Sure! Now that you reviewed it, I noticed that is not possible to pass a mapping as a function parameter and passing a struct as a function parameter is only possible after 0.8.0 (or if `pragma experimental ABIEncoderV2` is used).

If you agree, I will remove these lines: https://github.com/crytic/slither/blob/166eaed76178acd7657a773e25b37e68545fb9ff/slither/detectors/functions/external_function.py#L114-L115

and add the test for the structure type.

0xalpharush: Sounds good to me. I just want to make sure we have coverage for true positives for each case and also true negatives (if the function is already using calldata don't alert)

slither/detectors/functions/external_function.py
0xalpharush: You're correct

slither/detectors/functions/external_function.py
```diff
@@ -131,6 +155,14 @@ def _detect(self):  # pylint: disable=too-many-locals,too-many-branches
             # Next we'll want to loop through all functions defined directly in this contract.
             for function in contract.functions_declared:
 
+                # If all of the function arguments are non-reference type or calldata, we skip it.
+                reference_args = []
```
0xalpharush, line 159: Can you include which parameters need to be change for memory to calldata in the result? Right now it only specifies the function name, but it'd be helpful if it said "change `memory myarray` to `calldata myarray` in `function A()`

htadashi: Added in https://github.com/crytic/slither/pull/1022/commits/c878a8fc25b8a82fb41b4400f2c97f0ec6061fc6

htadashi:
@0xalpharush I had tried to rebase the PR, but some CI tests are still not passing. I am not sure why the tests are failing. Could I try to open a new PR after the review is complete?

slither/detectors/functions/external_function.py
htadashi: Sorry, I misunderstood. Added true positives in the tests in https://github.com/crytic/slither/pull/1022/commits/2e5aa1a20e88b548698fd2959a215dea6b3f08c0

htadashi:
> Some of the CI tests check how many results are output from slither:
> 
> https://github.com/crytic/slither/blob/f962d6c4eefcd4d5038a781875b826948f222b31/scripts/ci_test_embark.sh#L20
> 
> 
> So this "result count" needs to be updated when detectors are modified.
> Also, the test cases you've written don't have any true positives. Could you add "bad" test cases that recommend the optimization? Right now, it's only running against one's that the detector won't recommend optimizing.
> 
> I don't think there's any need to open a new PR.

Thanks for the review! 🙇🏻‍♂️
I am not sure if I understood the part of the result count, but it seems that embark's workflow is not currently enabled: https://github.com/crytic/slither/blob/f962d6c4eefcd4d5038a781875b826948f222b31/.github/workflows/ci.yml#L29

montyly:
Merged with https://github.com/crytic/slither/pull/1318. Thanks [htadashi](https://github.com/htadashi) for the PR

PR #1025, opened 2022-01-26
feat: add arbitrary-send-erc20 and arbitrary-send-erc20-permit detectors

0xalpharush:
This PR modifies the `from-TransferFrom` detector by splitting it into arbitrary-send-erc20 and arbitrary-send-erc20-permit, so that the guidance/ exploit scenario is different. Also, I changed the `arbitrary-send` detector to `arbitrary-send-eth` for clarity.


0xalpharush:
TODO: Need to fix false negative in #block-slither before merging

PR #877, opened 2021-06-23
Refactor source mapping + add new source code APIs

montyly:
- core.source_mapping.SourceMapping is now a class (fix #695)
- Add slither.offset_to_objects to retrieve the object on a given offset (WIP)
- Add slither.utils.source_mapping APIs to allow retrieving definition/implementations/references informations from an offset (WIP)

PR #1042, opened 2022-02-09
Prefix line numbers with ':'

smoelius:
This PR makes Slither's output VSCode-clickable.

The change works when I run Slither manually, but I haven't figured out how to get the tests to pass. Is it an easy fix?


montyly:
Mhh actually if we do it there, it will break other integrations (like the markdown's integration). Maybe we need a flag to specify the output format?  Or have an output dependant of the context (like `--sarif` versus  the markdown output)

We started a large refactoring of the source mapping class a couple of months ago (https://github.com/crytic/slither/pull/877/files), we should revisit if this can be used

Issue #1014, opened 2022-01-11
Missing-Zero-Check detector not working for contract types


Jaime-Iglesias:
Running slither's missing-zero-check detector on the following contract

```solidity
//SPDX-License-Identifier: Unlicense
pragma solidity = 0.8.10;

import "@openzeppelin/contracts/token/ERC20/IERC20.sol";

contract ZeroCheck {
    address private _owner;
    IERC20 private _token;

    function setOwner(address owner) external {
        _owner = owner; // will detect
    }

    function setToken(IERC20 token) external {
        _token = token; // will not detect
    }
}
```

Will yield the following result

```
ZeroCheck.setOwner(address).owner (contracts/ZeroCheck.sol#10) lacks a zero-check on :
                - _owner = owner (contracts/ZeroCheck.sol#11)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#missing-zero-address-validation
```

Jaime-Iglesias:
Added support in - https://github.com/crytic/slither/pull/1043

PR #1043, opened 2022-02-09
Add support for contract types to the missing-zero-check detector

Jaime-Iglesias:
fixes https://github.com/crytic/slither/issues/1014


montyly:
As we are making broader coverage with this change, we probably need to revisit the heuristic to avoid having too many FPs.

For example, this is now raising a warning:
```solidity
library Lib{
    function f(address a) external returns(address){
        return a;
    }   
}   

contract Test {

    function test(address addr) public returns(address){
        return Lib.f(addr);
    }

}
```
```
Test.test(address).addr (Test.sol#9) lacks a zero-check on :
		- Lib.f(addr) (Test.sol#10)
```

You can also see this in the metacoin from truffle (it is why the CI is failing):
```
MetaCoin.getBalanceInEth(address).addr (MetaCoin.sol#28) lacks a zero-check on :
		- ConvertLib.convert(getBalance(addr),2) (MetaCoin.sol#29)
```

Jaime-Iglesias:
> 

I will take a look into it.

Issue #1029, opened 2022-02-01
Strange false positive with unprotected-upgradeable-contract


Amxx:
Let's consider this source code:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.11;

import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";

contract MyToken is Initializable, ERC20Upgradeable, OwnableUpgradeable, UUPSUpgradeable {
    /// @custom:oz-upgrades-unsafe-allow constructor
    constructor() initializer {}

    function initialize() initializer public {
        __ERC20_init("MyToken", "MTK");
        __Ownable_init();
        __UUPSUpgradeable_init();
    }

    function _authorizeUpgrade(address newImplementation)
        internal
        onlyOwner
        override
    {}
}
```

Running slither on it will raise a red
```
MyToken (contracts/MyToken.sol#9-24) is an upgradeable contract that does not protect its initiliaze functions: MyToken.initialize() (contracts/MyToken.sol#13-17). Anyone can delete the contract with: UUPSUpgradeable.upgradeTo(address) (node_modules/@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol#74-77)UUPSUpgradeable.upgradeToAndCall(address,bytes) (node_modules/@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol#87-90)Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unprotected-upgradeable-contract
```

However, the initialize function is protected by the `initializer` modifier, and the constructor does use this constructor to lock the base instance.

The upgradeTo path is protected by `onlyOwner`, which is locked unless Ownable was initialized, which requires `initialize()` to be executed, which is secured.

Am I missing something ?

montyly:
Hi @Amxx,

You're right, this is a FP, thanks for reporting it. We will improve the detector to discard this

Jaime-Iglesias:
Should be fixed by - https://github.com/crytic/slither/pull/1046

Jaime-Iglesias:
[...]

PR #1046, opened 2022-02-10
fix issue-1029, FP on unprotected-upgrade detector

Jaime-Iglesias:
Tentative fix for https://github.com/crytic/slither/issues/1029

### DIFF
--- a/slither/detectors/statements/unprotected_upgradeable.py
+++ b/slither/detectors/statements/unprotected_upgradeable.py
@@ -75,36 +75,36 @@ class UnprotectedUpgradeable(AbstractDetector):
 
         for contract in self.compilation_unit.contracts_derived:
             if contract.is_upgradeable:
-                    if not _has_initializer_modifier(contract.constructors) or not _has_protected_initialize(contract.functions):
-                        functions_that_can_destroy = _can_be_destroyed(contract)
-                        if functions_that_can_destroy:
-                            initiliaze_functions = [f for f in contract.functions if f.name == "initialize"]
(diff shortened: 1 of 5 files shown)


slither/detectors/statements/unprotected_upgradeable.py
```diff
@@ -21,6 +21,23 @@ def _can_be_destroyed(contract) -> List[Function]:
...
 
+def _has_initializer_modifier(functions: List[Function]) -> bool:
+    for f in functions:
+        for m in f.modifiers:
+            if m.name == "initializer":
+                return True
+    return False
+
+
+def _has_protected_initialize(functions: List[Function]) -> bool:
```
montyly, line 32: Can you detail why adding this check? I think `_has_initializer_modifier` on the constructors should be sufficient to remove the FP's category

If there is a situation where `initializer` is not applied to the init function, I think this would be a separate access control issue, not related to the unprotected upgrade detector

Jaime-Iglesias: I added the check becasue I found odd that something like

```solidity
contract C is Initializable {
  constructor() initializer {}
  
  function initialize() external {
     require(owner == address(0));
     owner = msg.sender;
  }
  
  function kill() external {
   require(msg.sender == owner);
   // kill contract
  }
}
```
Would be skipped.

Jaime-Iglesias: I removed the `_has_protected_initialize` check.

PR #1047, opened 2022-02-10
add FAQ and useful info to contribution guidelines

0xalpharush:
Started a brief FAQ. Going to think of others and add those (open to suggestions). 

I also updated the testing section so it's more clear how to update existing detectors. I'd appreciate feedback of the "project layout" section since there's likely better descriptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,8 +22,8 @@ Some pull request guidelines:
 
 ## Directory Structure
 
-Below is a rough outline of slither's design: 
-```
+Below is a rough outline of slither's design:
+```text
 .
 ├── analyses # Provides additional info such as data dependency 
 ├── core # Ties everything together
@@ -35,6 +35,8 @@ Below is a rough outline of slither's design:
 ├── visitors # Parses expressions and converts to slithir
 └── ...
 ```
+
+A code walkthrough is available [here](https://www.youtube.com/watch?v=EUl3UlYSluU).
 
 ## Development Environment
 Instructions for installing a development version of Slither can be found in our [wiki](https://github.com/crytic/slither/wiki/Developer-installation).
@@ -62,7 +64,7 @@ For each new detector, at least one regression tests must be present.
 - Create a test in `tests`
 - Update `ALL_TEST` in `tests/test_detectors.py`
 - Run `python ./tests/test_detectors.py --generate`. This will generate the json artifacts in `tests/expected_json`. Add the generated files to git.
-    - If updating an existing detector, identify the respective json artifacts and then delete them, or run `python ./tests/test_detectors.py --overwrite` instead.
+  - If updating an existing detector, identify the respective json artifacts and then delete them, or run `python ./tests/test_detectors.py --overwrite` instead.
 - Run `pytest ./tests/test_detectors.py` and check that everything worked.
 
 To see the tests coverage, run `pytest  tests/test_detectors.py  --cov=slither/detectors --cov-branch --cov-report html`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -205,6 +205,26 @@ Feel free to stop by our [Slack channel](https://empireslacking.herokuapp.com) (
 
 * The [SlithIR documentation](https://github.com/trailofbits/slither/wiki/SlithIR) describes the SlithIR intermediate representation.
 
+# FAQ
+
+* How do I exclude mocks or tests?
```
montyly, line 210: Maybe we can link to the wiki here? (https://github.com/crytic/slither/wiki/Usage#path-filtering)

README.md
```diff
@@ -205,6 +205,26 @@ Feel free to stop by our [Slack channel](https://empireslacking.herokuapp.com) (
...
+    ```
+    slither . --filter-paths tests|mocks
+    ```
+* How do I fix "unknown file" or compilation issues?
+    
+    Because slither requires the solc AST, it must have all dependencies available. If a contract has dependencies, `slither contract.sol` will fail. Instead, use `slither .` in the parent directory of `contracts/` (you should see `contracts/` when you run `ls`). If you have a `node_modules/` folder, it must be in the same directory as `contracts/`. To verify that this issue is related to slither, run the compilation command for the framework you are using e.g `npx hardhat compile`. That must work successfully; otherwise, slither's compilation engine, crytic-compile, cannot generate the AST. 
+
+* Why does slither detect reentrancy on functions with the `nonReentrant` modifier?
+
+    Slither is a static analysis tool and does not have the same information as a state machine, namely whether a mutex is locked or unlocked. Additionally, reentrant calls can enter *other* functions that lack the `nonReentrant` modifier, so these results should be carefully reviewed before marking them as false positives.
```
montyly, line 226: I think this is incorrect regarding Slither's limitation. We can filter out the results and improve the reentrancy detector to be smarter when dealing with `nonReentrant`

Maybe we can instead explain that slither leads to some false positive, the user can use `//slither-disable-next-line DETECTOR_NAME` (https://github.com/crytic/slither/wiki/Usage#path-filtering), and if there is a recurring pattern leading to FPs, they can open a github issue?

CONTRIBUTING.md
```diff
@@ -45,6 +61,7 @@ For each new detector, at least one regression tests must be present.
 - Create a test in `tests`
 - Update `ALL_TEST` in `tests/test_detectors.py`
 - Run `python ./tests/test_detectors.py --generate`. This will generate the json artifacts in `tests/expected_json`. Add the generated files to git.
+    - If updating an existing detector, either delete the old json artifacts or run `python ./tests/test_detectors.py --generate` instead.
```
montyly, line 64: "delete the old json artifacts", this might be confusing as we don't want people to delete all the json artifacts - this would lead to a lot of unnecessary changes in the commit

Maybe `denfity the json artifacts related to your changes, and delete them` , or something like that?

CONTRIBUTING.md
```diff
@@ -20,6 +20,22 @@ Some pull request guidelines:
 - Fill out the pull request description with a summary of what your patch does, key changes that have been made, and any further points of discussion, if applicable.
 - Title your pull request with a brief description of what it's changing. "Fixes #123" is a good comment to add to the description, but makes for an unclear title on its own.
 
+## Project Layout
```
montyly, line 23: We can add a link to https://www.youtube.com/watch?v=EUl3UlYSluU

README.md
```diff
@@ -205,6 +205,26 @@ Feel free to stop by our [Slack channel](https://empireslacking.herokuapp.com) (
 
 * The [SlithIR documentation](https://github.com/trailofbits/slither/wiki/SlithIR) describes the SlithIR intermediate representation.
 
+# FAQ
```
montyly, line 208: We need to update the list of section at the top of the file to link to the FAQ section

montyly (commented):
This is a great step forward :)

However, I am a bit unsure about the FAQ section, as it seems to be mostly a duplicate of https://github.com/crytic/slither/wiki/Usage. Maybe we should move the missing parts there?

README.md
0xalpharush: update table of contents

montyly:
I think once we fixed the CI (mostly `markdownlint` issues), we can merge this PR

PR #1057, opened 2022-02-12
Calldata yul

### LINKED ISSUE #1056: Yul calldata offset and length support

In solidity 0.7.5, `.offset` and `.length` were added to calldata variables of dynamic arrays in yul as described in https://blog.soliditylang.org/2020/11/18/solidity-0.7.5-release-announcement/.

It appears that currently slither is missing support for those cause code that makes use of that to fail inside the Yul parser. Support should be similar to the existing support of `.slot` and `.offset` for storage variables.

I'm planning on submitting a PR with a proposed fix, but I wanted to open an issue as well in case their are issues with the PR

hkalodner:
This PR fixes #1056

This was my first PR to this repo and I think I followed the directions attaching the binaries produced by `test_ast_parsing.py`, but let me know if I need to adjust anything there.

I don't think this PR accomplished exactly what's needed since it selects for any calldata variable, which does allow it to parse the `.offset` and `.length` variables, but I believe it's currently overly broad since it should only apply to dynamic calldata arrays. I'm not sure what the proper way is to restrict the check to just that type.

Further since the check is likely to be more complicated it might be worth adding a `is_dynamic_calldata_array` property to `LocalVariable`  class similar to the `is_storage` property


montyly:
Thanks @kalodner. I merged this PR through #1085

PR #1065, opened 2022-02-19
Fix path confusion on Windows

elopez:
This fixes an issue when running Slither on Windows. Some paths were
being stringified as Windows-style (backslash) while the rest of the
paths use POSIX-style (forward slash). This caused an issue when
trying to match paths, and Slither would complain the dependency
was not present.

After this change, the paths for import directives are always expressed
as POSIX-style paths when stringified.

See the following issue for a complete error message:

    https://github.com/crytic/crytic-compile/issues/241


montyly:
Is this still a draft, or is it ready for review? (it seems ready to review)
If we merge `dev` into this branch, the CI should pass

elopez:
This fixed the issue I was seeing but I did not perform more extensive testing on other platforms. If you think it's safe to merge then go ahead  👍 

Maybe setting up CI on Windows would be a good thing for the future 🤔

PR #1092, opened 2022-03-07
Fixed infinite recursion in show_ignore_findings

ConstantinHvber:
"show_ignore_findings" calls itself instead of returning a similar named variable.
This causes an infinite recursion, resulting in an RecursionError when creating a Slither Object via the Python API


montyly:
This is a great catch, thanks @[ConstantinHvber](https://github.com/ConstantinHvber)

PR #1100, opened 2022-03-11
Ir ssa fix phi

hbrodin:
Fix two issues related to IR SSA Phi-nodes.
1. Phi-nodes where not correctly propagated.
2. Empty phi-nodes where removed from function entry. This caused some variable instances to not have a definition.


hbrodin:
Correct PR is in #1102

PR #1101, opened 2022-03-11
Test cases for SSA IR

hbrodin:
Attempts to verify a few basic properties of the generated SSA IR.
Test cases are prepared, but disabled, for those that fail. This is not an exhaustive list of test cases. It is rather an initial attempt at verifying the generated SSA IR.


hbrodin:
@montyly for some reason the linting fails on files I haven't touched.

.github/workflows/IR.yml
```diff
@@ -0,0 +1,54 @@
...
+          pip install deepdiff
+          pip install pytest
+
+          pip install "solc-select>=v1.0.0b1"
+          solc-select install all
+          solc-select use 0.8.11
+
+          cd tests/test_node_modules/
+          npm install hardhat
+          cd ../..
```
elopez, line 50: From a quick look at the PR it seems you are not using this project, these commands can probably be removed then.

```suggestion
```

tests/test_ssa_generation.py
```diff
@@ -0,0 +1,942 @@
...
+
+    def build_filename(i: int) -> pathlib.Path:
+        return SCRIPT_DIR / f"slither_test_{i}.sol"
+
+    # NOTE (hbrodin): There is a race condition here (TOCTOU), ignore that for now (only applies to test code).
+    i = 0
+    fname = build_filename(i)
+    while fname.exists():
+        i += 1
+        fname = build_filename(i)
```
elopez, line 249: You can probably still use `NamedTemporaryFile` on Windows if you want, by passing `delete=False`, closing after writing, and then unlinking manually when you are finished.

elopez (commented):
Hi! I have just two small comments here regarding the workflow & Windows bits.

tests/test_ssa_generation.py
hbrodin: I agree. I'll try that. It would make the could much more clear.

Issue #434, opened 2020-04-09
Incorrect SSA for indirect call to function modifying global state


aaronyoo:
SSA doesn't track global state modification through indirect call. Consider the following two test cases, one performs the correct behavior but the other doesn't. I have attached screenshots of the resulting SSA IR in graph form

## Correct:
```
pragma solidity >=0.4.16 <0.7.0;
contract Contract {
    int public a;
    function f() public {
        e();
        a += 1;
    }

    function e() public {
        a -= 1;
    }
}
```

![correct](https://user-images.githubusercontent.com/11811606/78846590-1d925600-79c1-11ea-905a-0370cdd4b8e4.png)

## Incorrect:
```
pragma solidity >=0.4.16 <0.7.0;
pragma solidity >=0.4.16 <0.7.0;
contract Contract {
    int public a;
    function f() public {
        g();
        a += 1;
    }

    function e() public {
        a -= 1;
    }

    function g() public {
        e();
    }
}
```

![image](https://user-images.githubusercontent.com/11811606/78846635-40bd0580-79c1-11ea-8d84-f5adffe8c9d0.png)

---

The correct version uses a phi node to track the alteration to global state through the function call. However, when we add one more layer of indirection, it no longer tracks that alteration.

PR #1102, opened 2022-03-11
SSA IR fix phi handling

hbrodin:
Fix two issues related to IR SSA Phi-nodes.

Phi-nodes where not correctly propagated in certain circumstances.
Empty phi-nodes where removed from function entry. This caused some variable instances to not have a definition.


hbrodin:
> Is this a duplicate of #1100? Also, on whichever version of this PR is the correct one, we should link it to any relevant issues. Is this addressing Issue #434?

#1100 is incorrect. This is the one I'll continue working on. I haven't checked #434 specifically. I will checck if it will be fixed by his change. Main driver for this change is that for some control flows we fail to add a phi-node and that causes the tools (using SSA IR) to miss some variables that are influenced by control flow.

bsamuels453:
#1683 related

Issue #1062, opened 2022-02-18
[False Positive]: Reentrancy when making a call to a view function previous to a state update.


hjorthjort:
### What happened?

A reentrancy vulnerability was reported for a call that looked like this:

```
MyLibrary.aViewCall(token);
```
The `aViewCall` library function is a `view`, and it calls the token contract externally via `staticcall`.

Since the external call can not modify any state there is no reentrancy vulnerability.

### Can you share code with us to reproduce this bug?

```solidity
// SPDX-License-Identifier: No
pragma solidity ^0.8.7;

interface IERC20 {
    function decimals() external view returns (uint8);
}

library MyLibrary {

  function aViewCall(address token) internal view {
     (bool success ,  ) = token.staticcall(
            abi.encodeWithSelector(IERC20.decimals.selector)
            );
     require(success, "call failed");
  }
}

contract A {
  uint256 private protectMe = 1;
  function foo() external {
    MyLibrary.aViewCall(0x6B175474E89094C44Da98b954EedeAC495271d0F);
    protectMe += 1;
  }
}
```

### Version

0.8.2

### Relevant log output

```shell
Reentrancy in A.foo() (Foo.sol#16-19):
	External calls:
	- MyLibrary.aViewCall(0x6B175474E89094C44Da98b954EedeAC495271d0F) (Foo.sol#17)
	State variables written after the call(s):
	- protectMe += 1 (Foo.sol#18)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#reentrancy-vulnerabilities-2
```

Jaime-Iglesias:
This looks like a False Positive in the reentrancy detector - thx for reporting it.

montyly:
Additionally we should update
https://github.com/crytic/slither/blob/de0d526cfd6a8e248a432887aa8d8ec3d3c95653/slither/core/declarations/function.py#L309-L311

To not return `True` if the `Call` is a `LowLevelCall` where `ir.function_name == "staticcall"`

montyly:
This issue was fixed with https://github.com/crytic/slither/pull/1119. The fix will be available in the upcoming release (by the end of the month)

PR #1119, opened 2022-03-16
reentrancy-no-eth: do not count staticcalls as reentrant

0xalpharush:
Fixes https://github.com/crytic/slither/issues/1062


slither/core/declarations/function.py
```diff
@@ -306,7 +306,11 @@ def can_reenter(self, callstack=None) -> bool:
         if self._can_reenter is None:
             self._can_reenter = False
             for ir in self.all_slithir_operations():
-                if isinstance(ir, Call) and ir.can_reenter(callstack):
+                if (
+                    isinstance(ir, Call)
+                    and ir.can_reenter(callstack)
+                    and ir.function_name != "staticcall"
```
montyly, line 312: `ir.function_name` is not a member of `Call`, so this can crash. I think we should add this check in `LowLevelCall.can_reenter`: https://github.com/crytic/slither/blob/5021f3e7a04e46dd32655f8d168b43a8bb000770/slither/slithir/operations/low_level_call.py#L59-L64

montyly, line 312: `ir.function_name` is not a member of `Call`, so this can crash. I think we should add this check in `LowLevelCall.can_reenter`: https://github.com/crytic/slither/blob/5021f3e7a04e46dd32655f8d168b43a8bb000770/slither/slithir/operations/low_level_call.py#L59-L64

montyly (changes requested):

slither/core/declarations/function.py
0xalpharush: I made this change and added a test case that hits this code.

PR #1128, opened 2022-03-17
Print all entry points not reaching whenNotPaused. Fix - 1114

feliam:
Prints entry points that are not can not reach the modifier **whenNotPause**


slither/printers/summary/when_not_paused.py
```diff
@@ -0,0 +1,72 @@
...
+
+    WIKI = "https://github.com/trailofbits/slither/wiki/Printer-documentation#when-not-paused"
+
+    @staticmethod
+    def _list_functions_not_reaching_modifier(
+        contract: Contract, modifier_name: str = "whenNotPaused"
+    ) -> List[Function]:
+        results: Set[Function] = set()
+        for entry_point in contract.functions_entry_points:
+            if entry_point.visibility not in ["public", "external"]:
```
montyly, line 26: Because this is iterating over `functions_entry_points`, there is no need to check for the visibility 

https://github.com/crytic/slither/blob/8344524cd35a1edf37c1cc791eb08c8ffff645a8/slither/core/declarations/contract.py#L487

slither/printers/summary/when_not_paused.py
```diff
@@ -0,0 +1,72 @@
...
+                continue
+            if entry_point.is_constructor or entry_point.view:
```
montyly, line 28: We probably want `view` and `pure` here

montyly (changes requested):

PR #1134, opened 2022-03-18
WIP allow checking contract source against known OZ source hashes

rmi7:


slither/utils/standard_libraries.py
```diff
@@ -65,6 +67,16 @@ def is_openzepellin(contract: "Contract") -> bool:
     return is_zep
 
 
+def is_openzepellin_strict(contract: "Contract") -> bool:
```
elopez, line 70: Small typo:
```suggestion
def is_openzeppelin_strict(contract: "Contract") -> bool:
```

Issue #1026, opened 2022-01-28
Error: Top level UserDefinedValueTypeDefinition not supported


bearpebble:
I'm trying to use slither on a project that makes use of user defined value types. Basically the example from [this post](https://blog.soliditylang.org/2021/09/27/user-defined-value-types/).
I'm using the latest slither release at the time (0.8.2).

Example contract that triggers the issue:
```solidity
//SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.8;

// Represent a 18 decimal, 256 bit wide fixed point type
// using a user defined value type.
type UFixed is uint256;

/// A minimal library to do fixed point operations on UFixed.
library FixedMath {
    uint constant multiplier = 10**18;

    /// Adds two UFixed numbers. Reverts on overflow, 
    /// relying on checked arithmetic on uint256.
    function add(UFixed a, UFixed b) internal pure returns (UFixed) {
        return UFixed.wrap(UFixed.unwrap(a) + UFixed.unwrap(b));
    }
    /// Multiplies UFixed and uint256. Reverts on overflow,
    /// relying on checked arithmetic on uint256.
    function mul(UFixed a, uint256 b) internal pure returns (UFixed) {
        return UFixed.wrap(UFixed.unwrap(a) * b);
    }
    /// Take the floor of a UFixed number.
    /// @return the largest integer that does not exceed `a`.
    function floor(UFixed a) internal pure returns (uint256) {
        return UFixed.unwrap(a) / multiplier;
    }
    /// Turns a uint256 into a UFixed of the same value.
    /// Reverts if the integer is too large.
    function toUFixed(uint256 a) internal pure returns (UFixed) {
        return UFixed.wrap(a * multiplier);
    }
}


contract Greeter {
    using FixedMath for UFixed;
    UFixed public someValue;

    constructor(string memory _greeting) {
    }
}

```

The output of `slither .` is
```
Traceback (most recent call last):
  File "/home/xxx/.local/lib/python3.10/site-packages/slither/__main__.py", line 743, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/home/xxx/.local/lib/python3.10/site-packages/slither/__main__.py", line 84, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/home/xxx/.local/lib/python3.10/site-packages/slither/__main__.py", line 67, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
[...]

Issue #951, opened 2021-09-28
Add 0.8.8 support


montyly:
https://blog.soliditylang.org/2021/09/27/solidity-0.8.8-release-announcement/

Most relevant:
- https://blog.soliditylang.org/2021/09/27/user-defined-value-types/
- min and max for Enums

fiddlemath:
Just a +1 here; Getting support for user-defined value types into Slither would save me and my team considerable grief.

0xalpharush:
Need to support
```
contract test {
        enum MinMax { A, B, C, D }

        function min() public returns(uint) { return uint(type(MinMax).min); }
        function max() public returns(uint) { return uint(type(MinMax).max); }
}
```

PR #1135, opened 2022-03-18
Add support for user defined types

montyly:
This is a WIP - the main blocker is how to handle `warp` operations without adding too much complexity to the parsing.


montyly:
Related: https://github.com/crytic/slither/issues/951
Fix : https://github.com/crytic/slither/issues/1026

PR #1141, opened 2022-03-25
fix typo in output

bohendo:
Fixes "Different versions of Solidity is used" -> "Different versions of Solidity are used" in slither output.

Also fixes the same string being used for comparisons in tests but I wasn't able to figure out how to run all the tests locally, I guess CI will run those checks for me..

I was able to run individual suites w specific calls into scripts eg `bash scripts/ci_test_data_dependency.sh` but some of these have suspicious output & still exit w code 0. I expected `bash scripts/ci_test.sh` to run all test suites w a clear indication when something failed but it exits w/out any output & exit code 0.. Maybe that means everything works as expected?

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -776,7 +776,7 @@ def main_impl(all_detector_classes, all_printer_classes):
         if args.checklist:
             output_results_to_markdown(results_detectors, args.checklist_limit)
 
-        # Dont print the number of result for printers
+        # Don't print the number of result for printers
         if number_contracts == 0:
             logger.warning(red("No contract was analyzed"))
         if printer_classes:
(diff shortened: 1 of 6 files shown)


bohendo:
What are the `id` values in the expected output files? They seem hashy because some changed when I updated the description string. I just changed them to what the tests expected them to be to get the tests passing, not sure if there'd be a better way to do that..

I see now that `pytest` is what I should use for testing rather than running the ci shell scripts, but is there an easy way to run just one test? I'd expect it to be something like below (which doesn't work):

```
pytest 'tests/test_detectors.py::test_detector[<class 'slither.detectors.attributes.constant_pragma.ConstantPragma'>: 0.4.25/pragma.0.4.25.sol]'
```

0xalpharush:
@bohendo whenever you modify a detector, you need to run `python ./tests/test_detectors.py --overwrite`

bohendo:
Looks like `python ./tests/test_detectors.py --overwrite` automatically does what I did by hand. I reverted a prev commit that had corrupted author data & force pushed a commit that uses the automated changes rather than my manual ones.

Everything should be good to go now pending review. Thanks for the feedback.

PR #1167, opened 2022-04-11
fix shell rendering in bug report template

holic:
The "Relevant log output" wasn't rendering as a code snippet due to some extra indentation


0xalpharush (approved):

montyly:
Good catch, thanks @holic

PR #1171, opened 2022-04-15
Prevent a TypeError on Function.__str__

### LINKED ISSUE #1170: Function.__str__ can raise a TypeError on special FunctionTypes
### Describe the issue:

For special function types like `FunctionType.CONSTRUCTOR_VARIABLES`, the `Function._name` member variable will be `None`, yet the `Function.name` property will return a string as expected (`"slitherConstructorVariables"`). `Function.__str__`, however, returns `self._name` rather than `self.name`. Returning `None` from `__str__` raises a `TypeError` at runtime.

Every function for which `Function._function_type == FunctionType.CONSTRUCTOR_VARIABLES` seems to have this problem.

### Version:

0.8.2

### Relevant log output:

_No response_

ESultanik:
`Function.__str__` used to return `self._name`. However, for special function types this value can be `None`, which causes `__str__` to raise a `TypeError` at runtime. This small change uses `self.name` instead, which is a property that returns valid strings for all special function types.

Fixes #1170

### DIFF
--- a/slither/core/declarations/function.py
+++ b/slither/core/declarations/function.py
@@ -1724,6 +1724,6 @@ class Function(SourceMapping, metaclass=ABCMeta):  # pylint: disable=too-many-pu
     ###################################################################################
 
     def __str__(self):
-        return self._name
+        return self.name
 
     # endregion


montyly:
Great catch, thanks @ESultanik

PR #1178, opened 2022-04-21
update setup.py with dev dependencies and update CI

0xalpharush:
This allows us to only update the setup.py without having to edit each CI test. Also it ensures that our development environments stay up to date (pylint and black, in particular).


setup.py
```diff
@@ -17,6 +17,18 @@
         # "crytic-compile>=0.2.2",
         "crytic-compile",
     ],
+    extras_require={
+        "dev": [
+            "black>=22.3.0",
```
montyly, line 22: We need to use a fixed version for black/pylint, otherwise we might have conflicts

.github/workflows/black.yml
```diff
@@ -29,8 +29,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install .
-          pip install deepdiff numpy
+          pip install ".[dev]"
+          pip install numpy
```
elopez, line 33: This one is a linting task, do we need to install things here?

.github/workflows/ci.yml
```diff
@@ -65,11 +65,8 @@ jobs:
       - name: Install dependencies
         run: |
```
elopez, line 67: nit:
```suggestion
```

.github/workflows/linter.yml
```diff
@@ -29,8 +29,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install .
-          pip install deepdiff numpy
+          pip install ".[dev]"
+          pip install numpy
```
elopez, line 33: Same here, I think we can drop these.

elopez (commented):
Just a few comments, see below 👍

setup.py
```diff
@@ -17,6 +17,18 @@
...
+            "pylint>=2.13.4",
+            "pytest==7.0.1",
+            "pytest-cov",
+            "deepdiff",
+            "typing_extensions==4.1.1",
+            "importlib_metadata==4.8.3",
+            "solc-select>=v1.0.0b1",
```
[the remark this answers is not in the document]
0xalpharush: I ran `pip install -e ".[dev]"` locally and didn't have issues. I'm using python 3.10

montyly:
Fix https://github.com/crytic/slither/issues/1194

PR #1186, opened 2022-04-22
Optimization: use `str.startswith(tuple)`, and `str.endswith(tuple)`, …

BoboTiG:
…when revelant


montyly:
Good catch @BoboTiG, thanks for the contribution

BoboTiG:
@0xalpharush a little bit late, and I'm lazy, so I redirect you to that commit where you will find numbers: https://github.com/ethereum/eth-utils/pull/223/commits/988762508003d922ae917e7199197cdce2bb3214 :)

Issue #1191, opened 2022-04-27
[Bug]: Slither path filtering on Windows is different than on *NIX


pcaversaccio:
### Describe the issue:

If you use the latest Slither version on Windows and your project contains a `slither.config.json` file, the options are ignored if you run the Slither analyzer. The same applies if the options are part of the command itself (e.g. `slither file.sol --detect arbitrary-send,pragma`).

### Code example to reproduce the issue:

Clone my repo [here](https://github.com/pcaversaccio/hardhat-project-template-ts) and run `slither .` in Windows and for comparison with Ubuntu using WSL.

**Windows Output:**
![image](https://user-images.githubusercontent.com/25297591/165497820-58e9c3b8-066a-4a58-bae4-7584b13efd1b.png)

**Ubuntu Output:**
![image](https://user-images.githubusercontent.com/25297591/165497912-4e6b683d-cb1e-4a31-a543-e1819a66f084.png)

### Version:

- Slither: `0.8.3`
- OS: Windows 11
- Shell: PowerShell 7.2.3

elopez:
Thanks for the report @pcaversaccio! I tested the repository you linked and I see Slither reads the config file correctly on Windows. There is however an issue with path filtering -- as Slither does it on absolute paths, which have `\` on windows vs `/` on Linux/WSL, and your `filter_paths` config has paths with `/`, it does not get applied as expected when you run Slither on Windows. I opened a PR with a possible fix (#1196). As a workaround for now, you can use `.` instead of `/`; that should match a character in both OSes and make filtering work the same in Linux and Windows (in the linked repository, you would have `"filter_paths": "contracts.lib|contracts.test|contracts.Create2DeployerLocal.sol|@openzeppelin|hardhat.console.sol|xdeployer"`

I also tested passing arguments via the command line (e.g. `slither . --detect arbitrary-send`) and they worked fine as well on Windows (0 findings on that repository you linked, vs several ones with just `slither .`)

Can you confirm if the different amount of issues / path filtering is the problem you were having or if you experience something else apart from that?

pcaversaccio:
@elopez thanks for your answer - yes confirmed it's due to the non-POSIX compliance of the Windows path before filtering. The workaround with `.` works smoothly, thx! You can close this issue once your PR is merged (I just left a small comment re unit tests in your PR). Thx a lot for your assistance!

PR #1196, opened 2022-04-30
Unify path filtering across POSIX and Windows

elopez:
Path filtering is performed on absolute paths, which causes regexps
to not work uniformly across OSes, due to the different path separator
(forward vs backward slashes). This converts paths to POSIX-style
before filtering, so that regexps can be written in a single style
and work across all systems.

Fixes: #1191

### DIFF
--- a/slither/core/slither_core.py
+++ b/slither/core/slither_core.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import pathlib
+import posixpath
 import re
 from typing import Optional, Dict, List, Set, Union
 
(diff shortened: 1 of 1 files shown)


pcaversaccio (commented):
Thanks for fixing this issue @elopez - however, if you think it's a good idea, we should implement an additional unit test for Windows at least (and some of the other POSIX-compliant OSes) so that this issue is always tested as part of the CI. What do you think?

pcaversaccio:
> Would you like to open a PR with a test @pcaversaccio?

@0xalpharush sure, it will take some time however before I can submit it probably. I quickly checked your [test](https://github.com/crytic/slither/tree/master/tests) path folder. A quick `grep` search revealed in none of the test files `filter_paths` is used. So I will probably make a dedicated folder with a dedicated `slither.config.json`. Currently your CI files only target:
- ubuntu-latest
- windows-2022

I will keep this for the tests otherwise I would need to rewrite all the impacted CI pipelines. In the meantime, you can merge this PR.

pcaversaccio:
@0xalpharush I just opened a PR with the test suite here: https://github.com/crytic/slither/pull/1303.

PR #1231, opened 2022-06-08
analyze internal calls after performing constant propagation

0xalpharush:
This detector currently under-approximates arbitrary from's in `transferFrom` as it does not consider internal calls and it's dependency analysis is not strict (if `msg.sender` is used anywhere in the contract it's considered okay despite other tainted values).

This PR addresses those issues in two ways:
1) It performs constant propagation on `address(this)` and `msg.sender` so that arguments passed to library/internal function calls represent the underlying value.

2) Whenever `high_level_calls` was being used, it wasn't returning internal nodes/IRs to analyze, so I added a recursive analysis whenever an IR is an `INTERNAL_CALL`.


slither/detectors/erc/erc20/arbitrary_send_erc20.py
```diff
@@ -28,66 +32,68 @@ def permit_results(self) -> List[Node]:
...
-                    in all_high_level_calls
-                ):
-                    ArbitrarySendErc20._arbitrary_from(f.nodes, self._permit_results)
-                else:
-                    ArbitrarySendErc20._arbitrary_from(f.nodes, self._no_permit_results)
+        for func in contract.functions_entry_points:
+            has_permit = False
+            for node in func.nodes:
+                for ir in node.all_slithir_operations():
+                    # FIXME this is not always in the same order and produces causes permits to fall into non-permit
```
0xalpharush, line 39: There's a weird non-determinism issue where calls to `permit` are sometimes out-of-order and it changes which detector flags the test case that I need to fix.

slither/slithir/variables/constant.py
```diff
@@ -75,3 +75,6 @@ def __lt__(self, other):
 
     def __repr__(self):
         return f"{str(self.value)}"
+
+    def __hash__(self) -> int:
+        return super().__hash__()
```
0xalpharush, line 80:
```suggestion
        return self._val.__hash__()
```

slither/detectors/erc/erc20/arbitrary_send_erc20.py
```diff
@@ -27,69 +57,56 @@ def no_permit_results(self) -> List[Node]:
...
+        ):
+            results.append(ir.node.function)
+
+    def _detect_arbitrary_from(self, instructions: List[Operation]):
+        has_permit = False
+        for ir in instructions:
+            if isinstance(ir, InternalCall):
+                self.simplify_and_analyze(ir.function.nodes)
+            if isinstance(ir, HighLevelCall):
+                sig = ir.function.solidity_signature
```
0xalpharush, line 87: This throws an exception on `StateVariable`

slither/detectors/erc/erc20/arbitrary_send_erc20.py
```diff
@@ -27,69 +59,57 @@ def no_permit_results(self) -> List[Node]:
...
+            and ir.arguments[1]
+            not in (SolidityVariableComposed("msg.sender"), SolidityVariable("this"))
+        ):
+            results.append(ir.node)
+
+    def _detect_arbitrary_from(self, instructions: List[Operation]):
+        has_permit = False
+        for ir in instructions:
+            if isinstance(ir, InternalCall):
+                self.simplify_and_analyze(ir.function.nodes)
```
montyly, line 87: Mhh I think the recursive call here can lead to an infinite loop if the functions have a cycle.

Additionally I am a bit unsure about the strategy for internal call here. If you have an entry point that has an internal call, where the parameter is `msg.sender`, I think the logic here will fail to capture it
```solidity

function wrapper(address from, ..) internal{
   token.transferFrom(from, ...);
}

function externalFunction() external{
    wrapper(msg.sender, ...);
}

```

Because the functiond are analyzed independently here, I think it will lead to FP (unless I missed something?).

slither/detectors/erc/erc20/arbitrary_send_erc20.py
```diff
@@ -27,69 +59,57 @@ def no_permit_results(self) -> List[Node]:
...
+            # High level calls (view functions) can also be `StateVariable`
+            if isinstance(ir, HighLevelCall) and isinstance(ir.function, Function):
+                sig = ir.function.solidity_signature
+                has_permit |= sig == "permit(address,address,uint256,uint256,uint8,bytes32,bytes32)"
```
montyly, line 91: There is no guarantee on the order of the IR (the list is created by iterating over the nodes - which don't have an order, as they coming from a graph).  
So I think `has_permit` should be assigned ahead of time, or explored through a recursive path exploration.

For example
```solidity
if(..){
   token.permit
   return;
}
token.transferFrom()
```

Here `has_permit` might be set to true, or to false when `token.transferFrom` is reached; it depends on the order of the nodes in `func.nodes`

montyly (changes requested):

PR #1242, opened 2022-06-17
Dockerfile: upgrade from Ubuntu Bionic (18.04) to Focal (20.04)

greenknot:
It fixes the following Python error when running slither:

```
  AttributeError: module 'typing' has no attribute '_SpecialForm'
```


montyly:
Thanks @greenknot

Issue #1263, opened 2022-06-27
Please support ERC1363 and ERC4524 in `slither-check-erc`


lukehutch:
### Describe the desired feature

ERC1363 and ERC4524 are ERC20 extensions that are much safer than ERC223 and much safer and simpler than ERC777. Please consider adding them to `slither-check-erc`.

(Both ERC223 and ERC777 don't check a return value from the callback hook functions, so the fallback function may be called instead; ERC777 calls the sender notification hook before updating account state, which can lead to reentrancy attacks)

PR #1274, opened 2022-07-07
Support of ERC1363, ERC4524 for slither-check-erc

h00p30:
As requested in #1263


montyly:
This is great, thanks @h00p30 for the contribution

PR #1280, opened 2022-07-09
Fix: virtual internal function in delegatecall-loop and msg-value-loop

### LINKED ISSUE #1279: [Bug-Candidate]: delegatecall-loop and msg-value-loop in abstract contract cause crash
### Describe the issue:

`delegatecall-loop` and `msg-value-loop` detectors cause crash if the abstract class has a payable function that calls the virtual internal method.

As I understand it, the reason is that [there](https://github.com/crytic/slither/blob/db703d82f4876e30f1ca10fb78d85b4dd4c31ec1/slither/detectors/statements/delegatecall_in_loop.py#L37) is no check to see if the method is implemented.

```
abstract contract FailingContract {
    function _virtualFunction() internal virtual;

    function failingFunction(
    ) external payable {
        _virtualFunction();
    }
} 
```

### Version:

0.8.3

### Relevant log output:

```shell
if node.type == NodeType.STARTLOOP:
AttributeError: 'NoneType' object has no attribute 'type'
```

vladyan18:
Fix #1279

### DIFF
--- a/slither/detectors/statements/delegatecall_in_loop.py
+++ b/slither/detectors/statements/delegatecall_in_loop.py
@@ -34,7 +34,7 @@ def delegatecall_in_loop(
             and ir.function_name == "delegatecall"
         ):
             results.append(ir.node)
-        if isinstance(ir, (InternalCall)):
+        if isinstance(ir, (InternalCall)) and ir.function.is_implemented:
             delegatecall_in_loop(ir.function.entry_point, in_loop_counter, visited, results)
 
     for son in node.sons:
--- a/slither/detectors/statements/msg_value_in_loop.py
+++ b/slither/detectors/statements/msg_value_in_loop.py
@@ -30,7 +30,7 @@ def msg_value_in_loop(
     for ir in node.all_slithir_operations():
         if in_loop_counter > 0 and SolidityVariableComposed("msg.value") in ir.read:
             results.append(ir.node)
-        if isinstance(ir, (InternalCall)):
+        if isinstance(ir, (InternalCall)) and ir.function.is_implemented:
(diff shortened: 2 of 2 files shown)


0xalpharush:
I think a more robust solution would be to filter abstract contracts from `contracts_derived` since we're likely on interested in contracts which inherit from the abstract contract and implement the virtual functions. This doesn't cause an error because only `FailingContract` is analyzed:
` slither test.sol`
```
abstract contract FailingContract {
    uint x;
    function _virtualFunction() internal virtual;

    function failingFunction(
    ) external payable {
    for (uint i; i < 10; i++) {
        _virtualFunction();
    }
    }
} 
contract Test is FailingContract {
    function _virtualFunction() internal override {
        x = 1;
    }
}

```
Right now anywhere where a similar pattern is used e.g. `calls-loop` will fail on just a single abstract contract:
`slither test.sol --detect calls-loop`
```
abstract contract FailingContract {
    uint x;
    function _virtualFunction() internal virtual;

    function failingFunction(
    ) external payable {
    for (uint i; i < 10; i++) {
        _virtualFunction();
    }
    }
} 
```

vladyan18:
@0xalpharush 

> I think a more robust solution would be to filter abstract contracts from `contracts_derived` 

Indeed, if all sections of the code that use `contracts_derived` do not imply abstract contracts, this can simultaneously get rid of unwanted triggering of some detectors  in unflattened projects, such as `unimplemented-functions`, `unused-state-variable` etc.

On the other hand, this excludes some use cases, such as checking packages with contracts intended exclusively for inheritance in other projects.

montyly:
Hi @vladyan18, thanks for doing this. I ended up taking a different approach to fix the issue in #1307, and to fix the type associated with `entry_point`

PR #1282, opened 2022-07-09
Fix: binary operation with type(...).max and "ternary if"

### LINKED ISSUE #1281: [Bug-Candidate]: binary operation (type(...).max, <ternary operation>) cause crash
### Describe the issue:

`SplitTernaryExpression` fails if there is a binary operation with the `type(...).max` as the first operand and the result of the "ternary if" as the second.

If the first operand is a "ternary if" and the second is a `type(...).max` , then everything is fine.

It looks like the reason is that the `ElementaryTypeNameExpression` is missing [here](https://github.com/crytic/slither/blob/db703d82f4876e30f1ca10fb78d85b4dd4c31ec1/slither/utils/expression_manipulations.py#L83).

Possibly related to [1153 issue comment](https://github.com/crytic/slither/issues/1153#issuecomment-1127717023)

```
contract FailingContract {
    function failingFunction(
        bool condition
    ) external returns(uint256 res) {
        res = type(uint256).max / (condition ? 10 : 1) ;
    }
}
```

### Version:

0.8.3

### Relevant log output:

```shell
Ternary operation not handled uint256(<class 'slither.core.expressions.elementary_type_name_expression.ElementaryTypeNameExpression'>)
```

vladyan18:
Fix #1281

### DIFF
--- a/slither/utils/expression_manipulations.py
+++ b/slither/utils/expression_manipulations.py
@@ -9,6 +9,7 @@ from slither.core.expressions.assignment_operation import AssignmentOperation
 from slither.core.expressions.binary_operation import BinaryOperation
 from slither.core.expressions.call_expression import CallExpression
 from slither.core.expressions.conditional_expression import ConditionalExpression
+from slither.core.expressions.elementary_type_name_expression import ElementaryTypeNameExpression
 from slither.core.expressions.expression import Expression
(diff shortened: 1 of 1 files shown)


montyly:
This is a great fix. Thanks @vladyan18

montyly:
Replaced by #1306 (same but with a testcase)

PR #1291, opened 2022-07-17
fix: typo in slither-read-storage cli helper text

0xfarhaan:
note:
- required argument is --value but helper text says --values


0xalpharush:
In #1311

PR #1292, opened 2022-07-17
Detector for encodeWithSelector wrong argument count :jetski:

feliam:
The Solc Compiler does not help checking that the arguments used in when building the transaction calldata with encodeWhitSelector matches the target function. This simple detector simply tries to detect that. Miss-matching encodeWithSelector arguments and target function arguments.

### DIFF
--- a/slither/detectors/compiler_bugs/wrong_selector_with_selector.py
+++ b/slither/detectors/compiler_bugs/wrong_selector_with_selector.py
@@ -1,0 +1,116 @@
+import pdb
+
+from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
+from slither.slithir.operations import SolidityCall
+from slither.core.declarations.solidity_variables import SolidityFunction
+from slither.slithir.operations.assignment import Assignment
+from slither.slithir.variables.reference import ReferenceVariable
+from slither.slithir.variables.constant import Constant
+
+from slither.utils.function import get_function_id
+def get_signatures(c):
+    """Build a dictionary mapping function ids to signature, name, arguments for a contract"""
+    result = {}
+    functions = c.functions
+    for f in functions:
+        if f.visibility not in ["public", "external"]:
+            continue
+        if f.is_constructor or f.is_fallback:
+            continue
+        result[get_function_id(f.full_name)] = (f.full_name, *f.signature[:2])
+
+    variables = c.state_variables
+    for variable in variables:
+        if variable.visibility not in ["public"]:
+            continue
+        name = variable.full_name
+        result[get_function_id(name)] = (name, (),())
+
+    return result
+
+
+class WrongEncodeWithSelector(AbstractDetector):
+    """
+    Detect calls to abi.encodeWithSelector that may result in unexpected
+    calldata encodings
+    """
+
+    ARGUMENT = "encodeselector"  # slither will launch the detector with slither.py --mydetector
+    HELP = "abi.encodeWithSelector with unexpected arguments"
+    IMPACT = DetectorClassification.MEDIUM
+    CONFIDENCE = DetectorClassification.HIGH
+
+    WIKI = "https://github.com/trailofbits/slither/wiki/encode-with-selector"
+    WIKI_TITLE = "Encode With Selector uses unexpected parameters"
+    WIKI_DESCRIPTION = "Plugin example"
+    WIKI_EXPLOIT_SCENARIO = """
+```solidity
+contract Test {   
+    event Val(uint, uint);
+    function f(uint a, uint b) public {
+        emit Val(a, b);
+    }
+}
+contract D {
+    function g() public {
+        Test t = new Test();
+        address(t).call(abi.encodeWithSelector(Test.f.selector, "test"));
+    }
+}
+```
+The compiler will not check if the parameters of abi.encodeWithSelector match the arguments expected at the destination 
+function signature.
+"""
+    WIKI_RECOMMENDATION = ".."
+
+    def _detect(self):
+        #gather all known funcids
+        func_ids = {}
+        for contract in self.compilation_unit.contracts:
+            func_ids.update(get_signatures(contract))
+        #todo: include func_ids from the public db
+
+        results = []
+        for contract in self.compilation_unit.contracts:
+            for func, node in check(contract, func_ids):
+                info = [func, " calls abi.encodeWithSelector() with wrong arguments at", node ]
+                json = self.generate_result(info)
+                results.append(json)
+
+        return results
+
+def check(contract, func_ids):
+    """ check if contract has an ecodeWhitSelector that uses a selector
+        for a method with an unmatching number of arguments
+    """
+    result = []
+    for function in contract.functions_and_modifiers_declared:
+        for node in function.nodes:
+            for ir in node.irs:
+
+                if isinstance(ir, SolidityCall) and ir.function == SolidityFunction(
+                    "abi.encodeWithSelector()"
+                ):
+
+                    #build reference bindings dict
+                    assigments = {}
+                    for ir1 in node.irs:
+                        if isinstance(ir1, Assignment):
+                            assigments[ir1.lvalue.name] = ir1.rvalue
+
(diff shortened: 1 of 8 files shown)


slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,118 @@
...
+    function g() public {
+        Test t = new Test();
+        address(t).call(abi.encodeWithSelector(Test.f.selector, "test"));
+    }
+}
+```
+The compiler will not check_contract if the parameters of abi.encodeWithSelector match the arguments expected at the destination 
+function signature.
+"""
+    WIKI_RECOMMENDATION = "Make sure that encodeWithSelector is building a calldata that matches the target function signature"
```
0xalpharush, line 62:
```suggestion
    WIKI_RECOMMENDATION = "Make sure that arguments passed to abi.encodeWithSelector have the same types as the target function signature."
```

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,118 @@
...
+        #gather all known funcids
+        func_ids = {}
+        for contract in self.contracts:
+            func_ids.update(get_signatures(contract))
+        #todo: include func_ids from the public db
+
+        results = []
+        for contract in self.contracts:
+            for func, node in check_contract(contract, func_ids):
+                info = [func, " calls abi.encodeWithSelector() with wrong arguments at", node ]
```
0xalpharush, line 74:
```suggestion
                info = [func, " calls abi.encodeWithSelector() with arguments of incorrect type:", node ]
```

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if len(argument_types) != len(arguments):
+            result.append((function, node))
+
+        # Todo check_contract unmatching argument types for correct count
+    return result
+
+def check_contract(contract, func_ids):
+    """ check_contract if contract has an ecodeWhitSelector that uses a selector
+        for a method with an unmatching number of arguments
```
0xalpharush, line 110:
```suggestion
    """ Check contract's usage of abi.encodeWithSelector to ensure that the number of arguments 
        and their type match the function signature of the given selector.
```

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,118 @@
...
+
+        assert isinstance(selector, Constant)
+
+        _, _, argument_types = func_ids[selector.value]
+        arguments = ir.arguments[1:]
+
+        if len(argument_types) != len(arguments):
+            result.append((function, node))
+
+        # Todo check_contract unmatching argument types for correct count
```
0xalpharush, line 105: In addition to the number of arguments, can the types also be checked?

0xalpharush (changes requested):

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,118 @@
...
+    Detect calls to abi.encodeWithSelector that may result in unexpected calldata encodings
+    """
+
+    ARGUMENT = "wrongencodeselector"
+    HELP = "abi.encodeWithSelector with unexpected arguments"
+    IMPACT = DetectorClassification.MEDIUM
+    CONFIDENCE = DetectorClassification.HIGH
+
+    WIKI = "https://github.com/trailofbits/slither/wiki/encode-with-selector"
+    WIKI_TITLE = "Encode With Selector uses unexpected parameters"
```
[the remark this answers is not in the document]
feliam: As a note. This thing as is, only counts the number of arguments. Same number of wrong type of argumentes is not going to be detected. (I think) 
We should clarify  it in a TODO or in a diff issue.

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,126 @@
...
+        address(t).call(abi.encodeWithSelector(Test.f.selector, "test"));
+    }
+}
+```
+abi.encodeWithSelector's arguments do not match the types expected in the function signature.
+function signature.
+"""
+    WIKI_RECOMMENDATION = "Make sure that arguments passed to abi.encodeWithSelector have the same types as the target function signature."
+
+    def _detect(self):
```
0xalpharush, line 64: We've been trying to add type annotations recently. Here you'll want to do `-> List[Output]` and add the import `from slither.utils.output import Output`

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,126 @@
...
+                    " calls abi.encodeWithSelector() with arguments of incorrect type or with an incorrect number of arguments:",
+                    node,
+                ]
+                json = self.generate_result(info)
+                results.append(json)
+
+        return results
+
+
+def check_ir(function, node, ir, func_ids):
```
0xalpharush, line 85: Can you add type annotations and maybe change this function name to be more descriptive please?

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _detect(self):
+        # gather all known funcids
+        func_ids = {}
+        for contract in self.contracts:
+            func_ids.update(get_signatures(contract))
+        # todo: include func_ids from the public db
+
+        results = []
+        for contract in self.contracts:
```
0xalpharush, line 72: Here we can use `contracts_derived` instead of `contracts` so that we don't reanalyze a contract that is only inherited from

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
0xalpharush, line 1: Can you rename the file slither/detectors/compiler_bugs/wrong_encode_with_selector.py

0xalpharush:
Can we add a test case that uses an array type as an argument please?

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
tuturu-tech: Changed to `self.compilation_unit.contracts_derived`

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
tuturu-tech: Renamed

slither/detectors/compiler_bugs/wrong_selector_with_selector.py
tuturu-tech: Added type annotations and changed the name to `check_wrong_selector_encoding`

tests/e2e/detectors/snapshots/detectors__detector_WriteAfterWrite_0_8_0_write_after_write_sol__0.txt
0xalpharush, line 1: I think this file should not have been deleted and is causing tests to fail

slither/detectors/statements/wrong_encode_with_selector.py
```diff
@@ -0,0 +1,145 @@
...
+        # Does not handle LocalVariables
+        if not isinstance(selector, Constant):
+            if isinstance(selector.expression, TypeConversion) and isinstance(
+                selector.expression.expression, Literal
+            ):
+                selector = selector.expression.expression.value
+            else:
+                return result
+
+        _, _, argument_types = func_ids[selector.value]
```
0xalpharush, line 118: This causes an error on 0x9078f5996746f134f761b93939b0fd595f68845e. We should return early if we cannot determine which function the selector map backs to until we integrate a selector db

0xalpharush (changes requested):

PR #1297, opened 2022-07-24
Fix #1296

### LINKED ISSUE #1296: [Bug-Candidate]: slither-flat a bug of --convert-external flag, which converts external to public in an interface
### Describe the issue:

```
slither-flat 0xda2c338350a0E59Ce71CDCED9679A3A590Dd9BEC --strategy MostDerived --convert-external
```

You will see that external function was converted to public function in an interface called `IUniswapV2Factory` in a file having `Stake_FXS_WETH_` prefix.

`slither-flat 0xda2c338350a0E59Ce71CDCED9679A3A590Dd9BEC --strategy MostDerived --convert-external`

### Version:

0.8.3

### Relevant log output:

_No response_

TheStarBoys:
contract.contract_kind is always None. So use contract.is_interface instead of contract_kind.

### DIFF
--- a/slither/tools/flattening/flattening.py
+++ b/slither/tools/flattening/flattening.py
@@ -111,7 +111,7 @@ class Flattening:
 
         to_patch = []
         # interface must use external
-        if self._external_to_public and contract.contract_kind != "interface":
+        if self._external_to_public and not contract.is_interface:
             for f in contract.functions_declared:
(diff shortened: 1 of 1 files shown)


montyly:
Thanks @TheStarBoys, good catch. I also fixed `contract_kind` in #1308

PR #1239, opened 2022-06-15
correct instances of typo throughout repo: initiliaze -> initialize

kylebakerio:
WIP, though, and would like advising--when replacing all instances, 4 related tests fail.

Advice on how to get those tests passing to get this corrected welcome.

```python
~/code/forks/slither$ pytest ./tests/test_detectors.py
========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.9.5, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/kyle/code/forks/slither
plugins: mock-3.7.0, cov-3.0.0
collected 291 items                                                                                                                                                                                      

tests/test_detectors.py .......................................................................................................................................................................... [ 58%]
......................F..F..F..F.........................................................................................                                                                          [100%]

================================================================================================ FAILURES ================================================================================================
_________________________________________ test_detector[<class 'slither.detectors.statements.unprotected_upgradeable.UnprotectedUpgradeable'>: 0.4.25/Buggy.sol] _________________________________________

test_item = <tests.test_detectors.Test object at 0x7fba0cb16280>

    @pytest.mark.parametrize("test_item", ALL_TESTS, ids=id_test)
    def test_detector(test_item: Test):
        test_dir_path = pathlib.Path(
            pathlib.Path().absolute(),
            "tests",
            "detectors",
            test_item.detector.ARGUMENT,
            test_item.solc_ver,
        )
        test_file_path = str(pathlib.Path(test_dir_path, test_item.test_file))
        expected_result_path = str(pathlib.Path(test_dir_path, test_item.expected_result).absolute())
    
        set_solc(test_item)
        sl = Slither(test_file_path)
        sl.register_detector(test_item.detector)
        results = sl.run_detectors()
    
        with open(expected_result_path, encoding="utf8") as f:
            expected_result = json.load(f)
    
        results_as_string = json.dumps(results)
    
        for additional_file in test_item.additional_files:
            additional_path = str(pathlib.Path(test_dir_path, additional_file).absolute())
            additional_path = additional_path.replace("\\", "\\\\")
            results_as_string = results_as_string.replace(additional_path, GENERIC_PATH)
        test_file_path = test_file_path.replace("\\", "\\\\")
        results_as_string = results_as_string.replace(test_file_path, GENERIC_PATH)
        results = json.loads(results_as_string)
    
        diff = DeepDiff(results, expected_result, ignore_order=True, verbose_level=2)
        if diff:
            pprint(diff)
[...]

PR #1302, opened 2022-07-26
Fix broken tests introduced with #1239

montyly:


montyly:
Additionally this PR removes the etherlime test. This framework is not maintained anymore, and the test is taking forever on the CI

PR #1310, opened 2022-07-28
Update event_call.py

zjuchenyuan:
should be comma `,` rather than dot `.` when separating Event arguments


montyly:
Good catch, thanks @zjuchenyuan

PR #1334, opened 2022-08-09
implement detector for erc20 tokens that have function collision with…

0xalpharush:
… DOMAIN_SEPARATOR
- Also implements solidity sig and return type for public/external state variables

Heuristic: warn if external or public function/ state variable name collides with hash of "DOMAIN_SEPARATOR()" or the return type is not `bytes32`
cc @transmissions11


transmissions11:
amazing thank you so much @0xalpharush!

slither/detectors/functions/permit_domain_signature_collision.py
```diff
@@ -0,0 +1,55 @@
...
+    WIKI_RECOMMENDATION = "Remove or rename the function that collides with DOMAIN_SEPARATOR()."
+
+    def _detect(self):
+        results = []
+        domain_sig = get_function_id("DOMAIN_SEPARATOR()")
+        for contract in self.compilation_unit.contracts_derived:
+            if contract.is_erc20():
+                for func in contract.functions:
+                    if (
+                        func.name != "DOMAIN_SEPARATOR"
```
montyly, line 44: Maybe we should compare the solidity signature instead of the name?

We probably can't have a collision with `DOMAIN_SEPARATOR(arg1, arg2, ...)` but better safe than sorry

tests/test_detectors_mocked_contract.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    solc_select.switch_global_version("0.8.0", always_install=True)
+    sl = Slither("tests/mock_permit_domain_collision.sol")
+    assert len(sl.contracts_derived) == 1
+    contract = sl.contracts_derived[0]
+    # This will memoize the solidity signature and mutating the name
+    # won't change the function selector calculated
+    func = contract.get_function_from_signature("DOMAIN_SEPARATOR()")
+    # Change the name to mock
+    func.name = "MOCK_COLLISION"
```
montyly, line 18: Similar to https://github.com/crytic/slither/pull/1231, we should not change the properties on the fly, as its likely to break internal invariants ;)

.github/workflows/features.yml
```diff
@@ -50,3 +50,4 @@ jobs:
           pytest tests/test_functions_ids.py
           pytest tests/test_function.py
           pytest tests/test_source_mapping.py
+          pytest tests/test_detectors_mocked_contract.py
```
montyly, line 53: I think we should use the same testing strategy for all the detectors, otherwise it will be difficult to maintain. We can use `fopwCDKKK()` to simulate a collision with `DOMAIN_SEPARATOR()`

montyly (changes requested):
The detector is great, but I think we should not use a custom testing strategy for it.

We can use `fopwCDKKK()` to simulate a collision with

.github/workflows/features.yml
0xalpharush: I hadn't found a collision on 4bytes, which is why I mocked it and mutated the property. Thanks for finding this as I can test it normally.

slither/detectors/functions/permit_domain_signature_collision.py
0xalpharush: The signature is checked in the next half of the clause `get_function_id(func.solidity_signature) == domain_sig`. I checked that the name is not

slither/detectors/functions/permit_domain_signature_collision.py
```diff
@@ -0,0 +1,55 @@
...
+    # endregion wiki_exploit_scenario
+
+    WIKI_RECOMMENDATION = "Remove or rename the function that collides with DOMAIN_SEPARATOR()."
+
+    def _detect(self):
+        results = []
+        domain_sig = get_function_id("DOMAIN_SEPARATOR()")
+        for contract in self.compilation_unit.contracts_derived:
+            if contract.is_erc20():
+                for func in contract.functions:
```
montyly, line 42: we should iterate over `functions_entry_points` instead of function, to avoid reporting on internal functions

montyly (changes requested):
In addition, we could:
- Look for collision with public state variables (in case someone creates a state variable name `fopwCDKKK`)
- And maybe check the return type of `DOMAIN_SEPARATOR()` (`bytes32`)

Issue #1338, opened 2022-08-10
[Bug-Candidate]: False positive in costly-loop detector when evaluating an if statement with assignment and return


varunsrin:
### Describe the issue:

A costly-loop warning triggers when evaluating a loop that contains an if statement that includes both an assignment to a state variable followed by a return. This shouldn't be flagged as costly because it is the most efficient way of assigning this value since it only triggers once before exiting the loop. 

### Code example to reproduce the issue:

```solidity
uint 256 storedIndex = 0; 
arrayOfValues =  [1640995200, 1672531200, 1704067200];

function currYear() public {
  for (uint256 i = 0; i < arrayOfValues.length; i++) {
        if (arrayOfValues[i] > block.timestamp) {
            storedIndex = i;
            return;
        }
    }
}
```

### Version:

0.8.3

### Relevant log output:

```shell
<function> has costly operations inside a loop:
	- _nextYearIdx = i 
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#costly-operations-inside-a-loop
```

plotchy:
@0xalpharush @montyly Would this be a good application for checking post-dominance of any form of loop breaking statement to remove costly operations false positives? In cases where a loop-breaker is immediately hit, the operation is only performed once

0xalpharush:
Yeah, I think you could check if the post dominator node is a return and drops those from the results 
https://github.com/crytic/slither/blob/ce9dbf650d7acfee51ddafd844929f4e8d345672/slither/detectors/statements/costly_operations_in_loop.py#L39
Would you like to work on this issue @plotchy ?

plotchy:
@0xalpharush I looked into this a little bit, but I think it's over my head.

From what I can see Slither uses this algo:
```
Naive implementation of Cooper, Harvey, Kennedy algo
    See 'A Simple,Fast Dominance Algorithm'
```
[A Simple,Fast Dominance Algorithm](https://www.cs.rice.edu/~keith/Embed/dom.pdf)

Skimmed the doc to see if they have a postdominance impl, and they wrote about difficulties doing it with jumps out of loops, which is what we'd be targetting...

> We also measured the time to compute postdominators and postdominance frontiers.3 The postdominance
computation has a different behavior, because the reversed cfg has a different shape. This gives us additional
insight into the behavior of the algorithms.
In cfgs generated from real-world codes, it is more common to encounter irreducible graphs when
[...]

PR #1343, opened 2022-08-12
add post-dominator functionality

plotchy:
Reference: https://github.com/crytic/slither/issues/1338
Changes made to create output from comment: https://github.com/crytic/slither/issues/1338#issuecomment-1213195429

### DIFF
--- a/examples/scripts/export_post_dominator_tree_to_dot.py
+++ b/examples/scripts/export_post_dominator_tree_to_dot.py
@@ -1,0 +1,16 @@
+import sys
+from slither.slither import Slither
+
+
+if len(sys.argv) != 2:
+    print("python export_post_dominator_tree_to_dot.py contract.sol")
+    sys.exit(-1)
+
+# Init slither
+slither = Slither(sys.argv[1])
+
+for contract in slither.contracts:
+    for function in list(contract.functions) + list(contract.modifiers):
+        filename = f"{sys.argv[1]}-{contract.name}-{function.full_name}_post_dom.dot"
+        print(f"Export {filename}")
+        function.post_dominator_tree_to_dot(filename)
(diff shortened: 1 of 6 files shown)


plotchy:
I'm new to PRs -- just wanted to be able to share the changes I made. Not sure if this is the normal way to do so.

Some notes:

https://github.com/plotchy/slither/blob/41dd674d18deb42cf8ebe7bff5bab7ead6a96b95/slither/core/dominators/utils.py#L179-L184
- I'm unsure where to incorporate `break/assert/revert/throw/..` into the post-dom logic, seems this spot might be related?
- At the very least, should the `node.type/runner.type` logic be flipped as we're working in reverse order? 
```python
                    and node.type == NodeType.IF
                    and runner.type == NodeType.ENDIF
```

https://github.com/plotchy/slither/blob/41dd674d18deb42cf8ebe7bff5bab7ead6a96b95/slither/core/declarations/function.py#L576-L593
- On the topic of flipping logic when going in reverse order, does this `.entry_point` logic need to be flipped to something like `if isinstance(self.ir, RETURN)` ? So that it works off of return statements rather than `entry_point`s. Just spitballing here
```python
            #TODO plotchy this probably needs to be returns?
            if self.entry_point:
                self._compute_nodes_ordered_post_dominators(self.entry_point)
```

plotchy:
@0xalpharush Is all sons being postdominators of their fathers expected behavior? Maybe clear this up for me as cfg + nodes are all new to me.

```js
if (bool) {
    i = 1
} else {
    i = 2
}
print(i)
```
Is the `if (bool)` node postdominated by both the true & false sons, or neither of them?  Is it only postdominated by `ENDIF`?
Is the `print(i)` node dominated by both the true & false sons, or neither of them? Is `ENDIF` dominated by neither branch?

plotchy:
After speaking with @montyly , I realized I'd likely not be capable of creating + polishing a robust pdom implementation. Too far over my head atm. Thank you @0xalpharush for taking a look at this previously - appreciate your work.

Closing
🫡

PR #1350, opened 2022-08-17
Dev z

lovely011:
更新版本


montyly:
Hi @lovely011 . Thank you for the contribution, unfortunately we want to keep English for our output, and we do not support multiple languages at the moment.

Issue #1019, opened 2022-01-20
For loop prevents reentrancy detection


frangio:
This was reported before in https://github.com/crytic/slither/issues/940 but wasn't answered, so I wanted to bring attention to it again and give simple reproduction instructions.

In the following code I expect `reentrancy-no-eth` to detect both `r1` and `r2` as errors, but only `r1` is detected.

Is this a known limitation? Is it documented? Are there any plans to fix it? Thanks.

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract ITarget {
    function exec() public {}
}

contract Reenter {
    uint x;

    function r1(ITarget target) public {
        if (x > 0) {
            target.exec();
            x = 0;
        }
    }

    function r2(ITarget target) public {
        if (x > 0) {
            for (uint i = 0; i < 1; i++) {
                target.exec();
            }
            x = 0;
        }
    }
}
```

0xalpharush:
I'm not able to replicate this locally. When I run `slither test.sol --detect reentrancy-no-eth`:
```
Reentrancy in Reenter.r1(ITarget) (test.sol#11-16):
	External calls:
	- target.exec() (test.sol#13)
	State variables written after the call(s):
	- x = 0 (test.sol#14)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#reentrancy-vulnerabilities-1
```
Can you run `slither --version` and paste the result of slither when run on the contract above?

montyly:
Hi @frangio, thanks for pointing this out, somehow #940 was missed for the 0.8.2 release. 

We will investigate why it happens

Jaime-Iglesias:
> Hi @frangio, thanks for pointing this out, somehow #940 was missed for the 0.8.2 release.
> 
> We will investigate why it happens

The issue also affects `reentrancy-eth` as the detector code is very similar.

```solidity
    function bad3(ITarget target) public payable {
        if (notCalled = true) {
            for (uint256 i = 0; i < 10; i++) {
                target.exec{ value: msg.value }();
            }
        }
        notCalled = false;
    }
  ```

0xalpharush:
I found another instance of this. Slither detects that `nftCount` is updated after `_safeMint`, but it would miss it otherwise.
Test case: https://github.com/FrankieIsLost/smart-batched-auction/blob/dccadf0fdee5859f637ee29f46eb832e10a84b05/contracts/BatchAuction.sol#L102-L107
[...]

Issue #863, opened 2021-06-07
Reentrancy False Positive: Reported on internal function called by a nonReentrant external function


sudeepb02:
A Reentrancy vulnerability is reported by Slither in case an external nonReentrant function calls an internal function in the same contract.

For example:

```
function doSomething() internal {
    externalCallToAnotherContract();
}

function testExternal() nonReentrant external {
    doSomething();
}
```
Here as the `doSomething` is an internal function and only called inside a `nonReentrant` function, this should not be an issue.

montyly:
Hi @sudeepb02. Thank you for reporting this.

It's actually more tricky to consider this as a FP, and it depends on the other functions. For example if you have `doSomethingElse` function that does not have the `nonReentrant` modifier, an attacker might be able to exploit the reentrancy by re-entering into `doSomethingElse`. For example:
```solidity
function withdraw(uint amount) nonReentrant public{
     require(amount <= balances[msg.sender]);
     Receiver(msg.sender).send_funds({value: amount});
     balances[msg.sender] -= amount;
}

function withdraw_all() public{
     uint amount = balances[msg.sender];
     balances[msg.sender] = 0;
     Receiver(msg.sender).send_funds({value: amount});
}
```

However, we are working on a custom flag to tell Slither that all functions marked `nonReentrant` should be considered as safe: https://github.com/crytic/slither/pull/854

sudeepb02:
This is interesting, thanks for the info. But in my opinion allowing to set a custom flag would just mask the issue and not actually solve it.

Issue #765, opened 2021-01-11
Reentrant printer


montyly:
Print all the entry points that don't have the nonReentrant modifier, or are view/pure

PR #1351, opened 2022-08-17
Improve reentrancy detectors

montyly:
- Filter result based on `nonReentrant` modifier
  - `reentrancy-eth`/`reentrancy-no-eth`: do not warn if the function is non reentrant, and there is no other reentrant function that writes to the affected variable
  - `reentrancy-event`: do not warm in the function is non reentrant. 
  - Fix https://github.com/crytic/slither/issues/863, https://github.com/crytic/slither/issues/683. Replace https://github.com/crytic/slither/issues/765
- Fix incorrect detection on loop (fix https://github.com/crytic/slither/issues/1019)
- Improve python types
- Add relevant APIs for the reentrancy filtering:
   - `function.all_reachable_from_functions` list of all the functions in the call graph that leads to `function`
   - `function.is_reentrant`: check if the function is reentrant. Here we check if the `nonReentrant` modifier is directly applied, or for internal functions if one of the entry points (i.e. one of the public/external function that can reach the targeted function) is lacking `nonReentrant`

I might also improve the `reentrancy-event` filtering to detect cross-functions reentrancy on event - through this is usually only informational and have a limited impact. 

Additionally right now `nonReentrant` is only checked based on the modifier name. We could compare the source code hash of the known version to increase the confidence. 

This require additional internal testing before being merged.


montyly:
TODO: replace `is no other reentrant function that writes to the affected variable` to include read, and state variable that are public.

This will help reviewing cross protocol reentrancies

PR #1359, opened 2022-08-23
Add the `--no-fail-pedantic` flag

Boyan-MILANOV:
This PR adds the `--no-fail-pedantic` flag, which is the opposite of `--fail-pedantic`. It also changes the default which is now to set `--fail-pedantic` to `True`.

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -288,12 +288,6 @@ def parse_args(detector_classes, printer_classes):  # pylint: disable=too-many-s
     usage += "\t- 0x.. // a contract on mainet\n"
     usage += f"\t- NETWORK:0x.. // a contract on a different network. Supported networks: {','.join(x[:-1] for x in SUPPORTED_NETWORK)}\n"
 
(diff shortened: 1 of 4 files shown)


slither/__main__.py
```diff
@@ -402,6 +402,14 @@ def parse_args(detector_classes, printer_classes):  # pylint: disable=too-many-s
         default=defaults_flag_in_config["fail_pedantic"],
     )
 
+    group_detector.add_argument(
+        "--no-fail-pedantic",
+        help="Don't fail immediately if a finding is detected",
```
montyly, line 407: Maybe `Do not return the number of findings in the exit code?`, or something similar?

We could update the helper for `-fail-pendantic` to make it clear that both are related.

Slither does not fail immediately if this flag is not enabled ;)

Boyan-MILANOV: Ok. I based this on the previous help sentence for `--fail-pedantic`, we can change it

PR #1360, opened 2022-08-25
Added reference URLs to JSON output

### LINKED ISSUE #1224: Add reference urls to json output?

Slither only outputs reference urls in the txt output.  Would it be possible to add the urls to the json output too?  

Thanks

aranyadan:
Added relevant URL links to the wiki page in the JSON output for each
detector triggered.

Fixes #1224 

Dumps reference url in the json file for each detector object.


aranyadan:
so uh this breaks all the tests since I am changing the output format. The expected_result json no longer match

aranyadan:
should i replace all the expected jsons with the updated ones?

PR #1362, opened 2022-08-26
Adding poetry support for development

KholdStare:
It should be equivalent to what's already in setup.py, but should simplify
developer onboarding and reproducibility.
`poetry.lock` ensures anyone installing from scratch will get the exact
same versions of dependencies as everyone else.

Brief usage:

```
poetry install # install all dependencies in virtual env
poetry run pylint slither # run anything in the virtual env
poetry shell # start a shell in virtual env
```

See https://python-poetry.org/ for more docs

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,590 @@
+[[package]]
+name = "astroid"
+version = "2.12.4"
+description = "An abstract syntax tree for Python with inference support."
+category = "dev"
+optional = false
+python-versions = ">=3.7.2"
+
+[package.dependencies]
+lazy-object-proxy = ">=1.4.0"
+typing-extensions = {version = ">=3.10", markers = "python_version < \"3.10\""}
+wrapt = [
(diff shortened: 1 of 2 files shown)


KholdStare:
Let me know if this is unwelcome. I wanted to contribute to the project in general (for example adding mypy linting, and working through any errors there), and I am a big fan of using poetry for the development/build workflow - it makes setup and reproducibility that much easier.

0xalpharush:
@KholdStare I think we've decided to use Makefile for virtualenv management, linting, testing, etc  since none of the python tools like poetry are a catch all solution, so I'm closing this for now. This could change in the future though

KholdStare:
@0xalpharush I see. I think using poetry can go hand in hand with everything you pointed out. One of the main benefits is having a lock file. I don't see a dependency lockfile checked into the repo at the moment, and IMHO that can lead to reproducibility issues if some indirect dependencies drift. Poetry just replaced the "setup.py" portion with something more declarative.

Issue #1361, opened 2022-08-25
[Bug-Candidate]: Function-local (memory) variable shadowing not detected by Slither


mojtaba-eshghie:
### Describe the issue:

In solidity, it is possible to specify the name of the returned variable from the function as well as its value type. The presence of this feature cancels the need to both declare the variable in the function body as well as having a return statement. However, if the developer mistakenly declares the variable in the function body again, the newly declared variable will not shadow the previously declared variable in the function declaration statement. This leads to the function returning the default initial value for the named return variable instead of the intended value that is computed in the function body since it pertains to another variable only with the same name. The current solidity compiler (0.8.15) does not raise an error, and only shows a warning: `Warning: This declaration shadows an existing declaration for the declaration of the variable in the function body.`
 This is not correct since the declared variable in the function declaration statement will shadow the variable in the function body, not the other way around. This shadowing is not detected by Slither. In the accompanied code the `max` variable in the getMax function is shadowed by the default value max is initialized within the named return statement in the function declaration.

### Code example to reproduce the issue:
```
// SPDX-License-Identifier: MIT
pragma solidity >=0.6.0 <=0.9.0;

library ArrayUtils {
    function getMax(uint[] calldata a) public pure returns (uint max) {
        uint max = 0;
        uint max_index = 0;
        for (uint i = 0; i < a.length; i++) {
            if (a[i] >= max) {
                max_index = i;
                max = a[i];
            }
        }
    }

    function range(uint l) internal pure returns (uint[] memory r) {
        r = new uint[](l);      
        for (uint i = 0; i < l; i++) {
            r[i] = i;
        }
    }
}


contract ArraysUser {
    using ArrayUtils for *;

    function getMax(uint l) public pure returns (uint max) {
        max = ArrayUtils.range(l).getMax();
    }
}
```

### Version:

0.8.3

AndyBarnard:
Hey guys, I wrote this detector, currently trying to test it and getting stuck here:
`$ python ./tests/test_detectors.py --generate
Traceback (most recent call last):
  File "/home/andy/Dev/slither/./tests/test_detectors.py", line 399, in <module>
    all_detectors.ArbitrarySendEth,
AttributeError: module 'slither.detectors.all_detectors' has no attribute 'ArbitrarySendEth'. Did you mean: 'ArbitrarySend'?`
Not sure what's going on here, I'm a bit new to the slither codebase. Can anyone illuminate this for me? Is this a bug?

[...]

PR #1373, opened 2022-09-08
Adding a new detector: "Return Shadows Local"

xternet:
Detector for: [issue](https://github.com/crytic/slither/issues/1361)

Return functions without **return** statement, despite their logic, can return **default** values of **variables** declared inside **"returns()"**.

**Example**:

```solidity
pragma solidity 0.8.7;

contract Bug {
    function unnamed() external view returns(uint) {
        uint value = 1;
    } //returns 0 from "returns(uint)"

    function shadowed() external view returns(uint value) {
        uint value = 1;
    } //returns 0 from "returns(uint value)"
}
```

The detector reports when **"returns()" variables** are **unnamed** and **shadowed** (re-declared locally).


slither/detectors/shadowing/return_local.py
```diff
@@ -0,0 +1,117 @@
...
+        for var in function.returns:
+            if str(var)=='':
+                unnamed_return_vars.append(var)
+
+        return unnamed_return_vars
+
+    def no_return_statment_in(self, function): #1/4 ↑
+        if len(function.nodes)==0: return False #ignore inherited interfaces
+        for node in function.nodes:
+            if function.nodes and str(node.type)=='RETURN':
```
montyly, line 96: `function.nodes` is not needed here. Additionally, we should check against `NodeType.RETURN` instead of checking the `str` to make sure its robust

slither/detectors/shadowing/return_local.py
```diff
@@ -0,0 +1,117 @@
...
+
+    WIKI_RECOMMENDATION = """
+    1. Name return variables inside "returns(...)".
+    2. Don't re-declare the type of return variables.
+    3. Add a `return` statement at the end of the function."""
+
+    def info(self, err_unnamed, err_shadowed): #4/4 (end) ↰
+        info = []
+        if len(err_unnamed)>0:
+            info.append(self.generate_result('💥💥💥 DETECTED UNNAMED RETURN VARIABLE 💥💥💥\n'))
```
montyly, line 48: `generate_result` takes a list of `str` and slither's object. It should be called only once per result, otherwise every call to `generate_result` will lead to the creation of a separate finding.

So here, one call to `generate_result`, where the parameter is a list of information we want to show for this issue. For example we could have:
```python
result = [function, " has its return variable ", return_var, " shadowed by ", return_var_shadower]
info.append(self.generate_result(result))
```

Where `return_var` and `return_var_shadower` are the slither's object for the variable. Right now `return_var_shadower` is not passed in this function, but it could. Doing it this way will make easier to understand what was shadowed, and where are the variables definition.

montyly (changes requested):
This is great, thanks @xternet. 

I think this could be split into two detectors for clarity:
- Shadowing of return variable
- Unnamed return variable

PR #1375, opened 2022-09-16
Constants detector: Also detect Contract types

JorgeAtPaladin:
Presently the constants detector does not detect Contract types.

Eg.

```
IERC20 public token = IERC20(0x0000000000000000000000000000000000000000);
```

This is not a big deal as hardcoding addresses like that is generally an anti-pattern. However, if someone ever forks the detector for immutable flagging, than this will definitely come in handy!

Please do test before merging: we use the slither python API and not the detector framework ourselves.

This pull request specifically checks that if the candidate type is not Elementary, it must be a user defined Contract.

Let us know if it makes sense to invert the check and perhaps focus on ignoring dynamic types instead (because this PR still doesn't catch const enums I think for example, or re-typed int's, though that's also easily extendable)

### DIFF
--- a/slither/detectors/variables/possible_const_state_variables.py
+++ b/slither/detectors/variables/possible_const_state_variables.py
@@ -46,7 +46,7 @@ class ConstCandidateStateVars(AbstractDetector):
 
     @staticmethod
     def _valid_candidate(v):
-        return _is_valid_type(v) and not (v.is_constant or v.is_immutable)
+        return self._is_valid_type(v) and not (v.is_constant or v.is_immutable)
 
     # https://solidity.readthedocs.io/en/v0.5.2/contracts.html#constant-state-variables
     valid_solidity_function = [
--- a/slither/detectors/variables/possible_const_state_variables.py
+++ b/slither/detectors/variables/possible_const_state_variables.py
@@ -3,8 +3,10 @@ Module detecting state variables that could be declared as constant
 """
 from slither.core.compilation_unit import SlitherCompilationUnit
 from slither.core.solidity_types.elementary_type import ElementaryType
+from slither.core.solidity_types.user_defined_type import UserDefinedType
 from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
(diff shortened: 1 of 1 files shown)


slither/detectors/variables/possible_const_state_variables.py
```diff
@@ -29,9 +31,22 @@ class ConstCandidateStateVars(AbstractDetector):
...
+            type_t = t.type
+            if isinstance(type_t, Contract):
+                return True
+        return False
+        
+
     @staticmethod
     def _valid_candidate(v):
-        return isinstance(v.type, ElementaryType) and not (v.is_constant or v.is_immutable)
+        return _is_valid_type(v) and not (v.is_constant or v.is_immutable)
```
0xalpharush, line 49:
```suggestion
        return self._is_valid_type(v) and not (v.is_constant or v.is_immutable)
```

slither/detectors/variables/possible_const_state_variables.py
```diff
@@ -29,9 +31,22 @@ class ConstCandidateStateVars(AbstractDetector):
...
+
+    @staticmethod
+    def _is_valid_type(v):
+        t = v.type
+        if isinstance(t, ElementaryType):
+            return True
+        if isinstance(t, UserDefinedType):
+            type_t = t.type
+            if isinstance(type_t, Contract):
+                return True
```
0xalpharush, line 43:
```suggestion
        if isinstance(t, UserDefinedType) and isinstance(t.type, Contract):
            return True
```

JorgeAtPaladin:
@0xalpharush merged the two PRs, thanks for giving it a test run  :+1:

JorgeAtPaladin:
@0xalpharush blindly merged your fix (yet again), sorry for not properly reviewing these.

0xalpharush:
Can you run pylint and black https://github.com/crytic/slither/blob/master/CONTRIBUTING.md#linters please?

montyly:
Thanks @JorgeAtPaladin for the contribution. I made minor improvements and merged the PR with #1443

PR #1377, opened 2022-09-18
Improve constant parsing

elopez:
This PR improves and unifies constant parsing through the following:
* Using `Fraction` instead of `Decimal`, to eliminate precision loss and rounding on constants
* Unified parsing of integers through `convert_string_to_int/fraction`
* Correction of unary minus parsing (the negation was not being applied)
* Tests for rational and scientific notation constants


slither/core/expressions/literal.py
```diff
@@ -29,6 +31,10 @@ def subdenomination(self) -> Optional[str]:
     def __str__(self):
         if self.subdenomination:
             return str(convert_subdenomination(self._value, self.subdenomination))
+
+        if self.type in Int + Uint + Fixed + Ufixed + ["address"]:
```
elopez, line 35: Maybe fixed types shouldn't be in here

0xalpharush: They're a reserved keyword in solidity but not currently used.

0xalpharush (approved):

montyly:
This is great, thanks @elopez

PR #1384, opened 2022-09-26
Add tool to troubleshoot slither errors

elopez:
This introduces `slither-doctor`, a tool to aid users in troubleshooting problems encountered when running Slither. This tool helps users with:

* [x] checking that Slither detects and uses the correct compilation framework
* [x] checking that the project compiles successfully
* [x] checking that Slither, crytic-compile, solc-select are up to date 
* [ ] other features TODO

This tool is also meant as a one-stop shop for collecting information when users find a problem and need to file an issue on GitHub.

Screen Shot 2022-10-03 at 19 28 43


montyly:
@elopez - this looks really cool. Is the PR ready for review?

elopez:
@montyly there's more checks and docs it'd be nice to add, but I think the current state is fine as a sort of "MVP" and we can build upon this for extra features 👍

PR #1385, opened 2022-09-26
Read storage from block

bohendo:
Adds a new `--block` flag to `slither-read-storage` that allows the user to specify which block number to read storage from.

Note: requires an archive node to be provided as the rpc url.


slither/tools/read_storage/__main__.py
```diff
@@ -122,6 +128,12 @@ def main() -> None:
 
     srs = SlitherReadStorage(contracts, args.max_depth)
 
+    if args.block:
+        try:
+            srs.block = int(args.block)
```
montyly, line 133: `block` should be added as a member of `SlitherReadStorage`:  https://github.com/crytic/slither/blob/f0c87149f710ed80674e2b5136bb7f6657264054/slither/tools/read_storage/read_storage.py#L54-L65

as a `Union[str, int]`. We can use `latest` as the default value

montyly (changes requested):

PR #1392, opened 2022-10-01
Add detector for uint comparisons to greater than 0

cre-mer:
Comparing `uint > 0` has the same effect as comparing `uint != 0` since uints are always positive. But the latter operation uses less gas.


0xalpharush:
I think this would emit false positives unless it considers compiler version and settings. For example, these operations result in the same bytecode if solc 0.8.17 is used with optimizations enabled.

0xalpharush (changes requested):
Please filter results based on solc version and compiler optimization configurations and document the heuristic in the wiki.

slither/detectors/operations/comparing_uint_gt_zero.py
```diff
@@ -0,0 +1,91 @@
...
+
+    ARGUMENT = "comparing-uint-gt-zero"
+    HELP = "Comparing uint types greater than zero"
+    IMPACT = DetectorClassification.OPTIMIZATION
+    CONFIDENCE = DetectorClassification.HIGH
+
+    WIKI = "https://github.com/crytic/slither/wiki/Detector-Documentation#comparing-uint-gt-zero"
+
+    WIKI_TITLE = "Comparing uint greater than zero"
+    WIKI_DESCRIPTION = "uints are either 0 or greater. Comparing it different than 0 has the same effect but uses less gas."
```
0xalpharush, line 24: Would you add a Solidity snippet here like the other detectors have?

cre-mer:
This was fixed since the introduction of [tautology or contradiction](https://github.com/crytic/slither/wiki/Detector-Documentation#tautology-or-contradiction). No further work is needed.

PR #1402, opened 2022-10-05
new detector: Omitted Return Variables

xternet:
Detects when the return function omits to return the declared return variables.

Example:
```solidity
pragma solidity ^0.8.0;

contract Bug {
    function omitted() external view returns(uint val) {
        val = 1;
        return 0;
    } //returns 0
}


xternet:
Closing this stale PR because it no longer applies cleanly to current `master` and has not been maintained. If the detector is still useful, it should be re-evaluated and submitted against the current detector framework.

PR #1403, opened 2022-10-05
new detector: Unnamed Return Shadows Local

xternet:
Detects when return function without `return` statement has unnamed variables inside `returns`.

Example:
```solidity
pragma solidity ^0.8.0;

contract Bug {
    function unnamed() external view returns(uint) {
        uint val = 1;
    } //returns 0
}
```


xternet:
Closing this stale PR because it no longer applies cleanly to current `master` and has not been maintained. If the detector is still useful, it should be re-evaluated and submitted against the current detector framework.

PR #1404, opened 2022-10-05
new detector: Return Shadows Local

xternet:
Detector for: [issue](https://github.com/crytic/slither/issues/1361)
Detects when return function without `return` statement shadows self-related local variables.

Example:
```solidity
pragma solidity ^0.8.0;

contract Bug {
    function shadowed() external view returns(uint val) {
        uint val = 1;
    } //returns 0
}
```


0xalpharush:
@xternet This was a phenomenal idea, and we really appreciate your time! I realized this should be detected as part of an existing detector and fixed it in https://github.com/crytic/slither/pull/1510

PR #1406, opened 2022-10-05
fix: text for too recent version

mds1:
https://github.com/crytic/slither/pull/1389 changed the recommended solc versions but did not update the version text


montyly:
Good catch, thanks @mds1

PR #1409, opened 2022-10-06
new detector: Missing Initialization

godsangsoo:


godsangsoo:
Detector could check whether the state variables used for condition in the initialize() are initialized.

e.g. https://medium.com/valuedefi/vstake-pool-incident-post-mortem-4550407c9714

PR #1411, opened 2022-10-06
updated formatting of Unprotected upgradeable contract 

PatrickAlphaC:
so syntax highlighting applies in github wiki


montyly:
Great catch @PatrickAlphaC. Thanks for the contribution

Issue #1407, opened 2022-10-06
[Bug]: Slither fails with library events


yosriady:
### Describe the issue:

Slither fails when given a contract which uses a library event. See minimal example below.

### Code example to reproduce the issue:

```
// SPDX-License-Identifier: MIT
pragma solidity 0.8.16;

library Lib {
  event Event();
}

contract Test {
  function foo() external {
    emit Lib.Event(); // This line specifically
  }
}

```

### Version:

0.9.0

### Relevant log output:

```shell
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/slither/__main__.py", line 826, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.10/site-packages/slither/__main__.py", line 97, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.10/site-packages/slither/__main__.py", line 75, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/usr/local/lib/python3.10/site-packages/slither/slither.py", line 130, in __init__
    parser.analyze_contracts()
  File "/usr/local/lib/python3.10/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 503, in analyze_contracts
    self._convert_to_slithir()
  File "/usr/local/lib/python3.10/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 688, in _convert_to_slithir
    func.generate_slithir_and_analyze()
  File "/usr/local/lib/python3.10/site-packages/slither/core/declarations/function.py", line 1705, in generate_slithir_and_analyze
    node.slithir_generation()
  File "/usr/local/lib/python3.10/site-packages/slither/core/cfg/node.py", line 721, in slithir_generation
    self._find_read_write_call()
  File "/usr/local/lib/python3.10/site-packages/slither/core/cfg/node.py", line 919, in _find_read_write_call
    assert isinstance(ir.function, Function)
AssertionError
Error in contracts/Test.sol
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/slither/__main__.py", line 826, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.10/site-packages/slither/__main__.py", line 97, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
[...]

PR #1419, opened 2022-10-12
Fix detection of library events

medariox:
Fixes https://github.com/crytic/slither/issues/1407

### DIFF
--- a/slither/slithir/convert.py
+++ b/slither/slithir/convert.py
@@ -827,7 +827,7 @@ def extract_tmp_call(ins: TmpCall, contract: Optional[Contract]):  # pylint: dis
             # lib L { event E()}
             # ...
             # emit L.E();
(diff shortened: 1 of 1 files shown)


kristovatlas:
@medariox I'd like to test this branch against my Solidity code. If you have a moment, how would I run this branch's version of slither from the command line vs the one I installed with homebrew? The process is not obvious to me from the README.

medariox:
@kristovatlas 
You can use pip to install the patched version with: `pip install git+https://github.com/medariox/slither.git@patch-1`

kristovatlas:
Thanks. Can confirm that this resolves https://github.com/crytic/slither/issues/1407 with my code base.

montyly:
Thanks @medariox for this fix. I merged it in #1442. We will issue a minor release to include it

PR #1425, opened 2022-10-14
new detector: Missing Initialization

godsangsoo:


slither/detectors/operations/missing_initialization.py
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+```
+Bob calls `initialize`. However, Alice can also call `initialize`.
+"""
+    WIKI_RECOMMENDATION = (
+        "Check state variable should be initialized in initialize function"
+    )
+
+    def detect_initilize(self, func):
```
skimaharvey, line 44: `detect_initilize` sould probably be `detect_initialize`

godsangsoo: Thanks

Issue #1428, opened 2022-10-14
[Bug-Candidate]: Exception in `slither-flat` with `--solc-remap`


mdgeorge4153:
### Describe the issue:

When running `slither-flat` with `--solc-remap` I get the following error:
```
Traceback (most recent call last):
  File "/home/mdgeorge/.local/bin/slither-flat", line 8, in <module>
    sys.exit(main())
  File "/home/mdgeorge/.local/lib/python3.10/site-packages/slither/tools/flattening/__main__.py", line 116, in main
    flat = Flattening(
  File "/home/mdgeorge/.local/lib/python3.10/site-packages/slither/tools/flattening/flattening.py", line 79, in __init__
    self._get_source_code_top_level(compilation_unit.functions_top_level)
  File "/home/mdgeorge/.local/lib/python3.10/site-packages/slither/tools/flattening/flattening.py", line 84, in _get_source_code_top_level
    content = self._compilation_unit.core.source_code[src_mapping["filename_absolute"]]
TypeError: 'Source' object is not subscriptable
```

### Code example to reproduce the issue:

`slither-flat pkg/pool-stable/contracts/StablePool.sol --solc-remaps @balancer-labs=$(pwd)/node_modules/@balancer-labs/` in [balancer monorepo](https://github.com/balancer-labs/balancer-v2-monorepo)

### Version:

0.9.0

PR #1429, opened 2022-10-14
fix for issue #1428

mdgeorge4153:
Just a one-line change that seems to resolve issue #1428


montyly:
Hi @mdgeorge4153 . Thanks for the contributions. While this PR is correct, I ended up taking a slightly different approach to fix the issue with #1441

PR #1430, opened 2022-10-15
Update __main__.py

emretepedev:
changed mainet to mainnet

Type: documentation update
Issue: -
Breaking change: no


montyly:
Good catch, thanks @emretepedev

Issue #1322, opened 2022-08-02
[Bug-Candidate]: slither-read-storage: storage array slots miscalculated in slot layout for shadowed vars


plotchy:
### Describe the issue:

Weird edge case, though this happens with `uint256[50] __gap` quite often.

If there are instances of shadowed vars, slither-read-storage fails to differentiate them in the storage layout.
Only the first instance of the var is printed, and it uses the last instance's slot for the length holder.

### Code example to reproduce the issue:

```js
pragma solidity ^0.8.4;

contract BaseContract{
    uint256 one;
    uint256[50] private __gap;
    uint256 two;
}

contract DerivedContract is BaseContract{
    uint256 three;
    uint256[50] private __gap;
    uint256 four;
}
```

### Version:

0.8.3

### Relevant log output:

```shell
INFO:Slither-read-storage:
Contract 'DerivedContract'
BaseContract.one with type uint256 is located at slot: 0

INFO:Slither-read-storage:
Name: one
Type: uint256
Slot: 0

INFO:Slither-read-storage:
Contract 'DerivedContract'
DerivedContract.__gap with type uint256[50] is located at slot: 53 <-- ** Should be 1 **

INFO:Slither-read-storage:
Name: __gap
Type: uint256[50]
Slot: 53

INFO:Slither-read-storage:
Contract 'DerivedContract'
DerivedContract.__gap with type uint256[50] is located at slot: 53

INFO:Slither-read-storage:
Key: 0
Name: __gap
Type: uint256
Slot: 53

--- snip ---

INFO:Slither-read-storage:
Contract 'DerivedContract'
DerivedContract.__gap with type uint256[50] is located at slot: 53

INFO:Slither-read-storage:
Key: 19
Name: __gap
Type: uint256
Slot: 72

INFO:Slither-read-storage:
Contract 'DerivedContract'
BaseContract.two with type uint256 is located at slot: 51

INFO:Slither-read-storage:
Name: two
Type: uint256
Slot: 51

INFO:Slither-read-storage:
Contract 'DerivedContract'
DerivedContract.three with type uint256 is located at slot: 52

INFO:Slither-read-storage:
Name: three
Type: uint256
Slot: 52

** DerivedContract.__gap should be here with slot 53 ** 

INFO:Slither-read-storage:
Contract 'DerivedContract'
DerivedContract.four with type uint256 is located at slot: 103

INFO:Slither-read-storage:
Name: four
Type: uint256
Slot: 103
```

[...]

PR #1332, opened 2022-08-09
Return proper function names for `fallback()` and `receive()`

Boyan-MILANOV:
Makes slither return proper function names for `fallback()` and `receive()` instead of `()`. Implements the following suggestion: https://github.com/crytic/slither/issues/1331#issuecomment-1208637675 by @plotchy

Closes #1331

PR #1432, opened 2022-10-17
fix: include shadowed state variables in contract's state_variables property

0xalpharush:
fixes https://github.com/crytic/slither/issues/1322


montyly:
I am a bit unsure about adding an API breaking change here. I believe that #1332 can be fixed by using `state_variables_ordered`. So I would rather keep the current behavior and maybe document better `state_variables` and mention that this does not return shadowed variables - which are not accessible from the contract's perspective

0xalpharush:
> 

Okay I replaced this PR with 1444 and will update the docs later

PR #1434, opened 2022-10-18
parse_top_level_from_loaded_json: remove redundant get_scope calls

zhiqiangxu:


montyly:
Great catch @zhiqiangxu, thanks for the contribution

Issue #1354, opened 2022-08-17
Triage mode not prompting as expected


zarifpour:
### Describe the issue:

Running this command:
```
slither . --triage-mode
```
According to the [wiki](https://github.com/crytic/slither/wiki/Usage#triage-mode-1), slither is intended to triage as follows:
```
slither . --triage-mode
[...]
0: C.destination (test.sol#3) is never initialized. It is used in:
	- f (test.sol#5-7)
Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#uninitialized-state-variables
Results to hide during next runs: "0,1,..." or "All" (enter to not hide results): <user_input>
[...]
```

What I expect to happen:

1. Slither detects n issue(s) with the same detector filter
2. Slither asks the user which instance(s) of the issues they would like to ignore
3. Repeat steps for all detectors

However, in step 2, when slither expects an input to be given by a user after it detects the issue(s), I expect to be asked for an input with this message:
```
Results to hide during next runs: "0,1,..." or "All" (enter to not hide results): <user_input>
```
This is not written to the console (zsh) when the input is expected, instead it is written at the end of the findings. 

To be clear: instead of printing the expected message to cue the user for input, there is a blinking cursor that waits for user's input without telling the user what to do.

### Code example to reproduce the issue:

Run `slither . --triage-mode` on the root of any project.

### Version:

0.8.3

### Relevant log output:

```shell
0: DataUri.addAttributes(bytes32[],bytes32[]).data (contracts/serde/DataUri.sol#169) is a local variable never initialized
1: [redacted]
2: [redacted]
3: [redacted]
4: [redacted]
5: Multihash.decodeMultihash(bytes).hash (contracts/serde/Multihash.sol#113) is a local variable never initialized
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#uninitialized-local-variables
All

0: DataUri.encode(bytes) (contracts/serde/DataUri.sol#197-268) uses assembly
        - INLINE ASM (contracts/serde/DataUri.sol#216-265)
1: DataUri.decode(string) (contracts/serde/DataUri.sol#270-335) uses assembly
        - INLINE ASM (contracts/serde/DataUri.sol#285-332)
2: Multihash.decode512(uint64,uint64,bytes) (contracts/serde/Multihash.sol#217-236) uses assembly
        - INLINE ASM (contracts/serde/Multihash.sol#228-231)
[...]

Issue #922, opened 2021-08-21
Slither triage presents all options after running


transmissions11:
For some reason after 0.8.1 if I run `slither . --triage` it won't print the "results to hide during the next run" message for each grouping of findings, instead after you've gone through all of them it will print that message in one big blob like this (scroll horizontally and you'll see what i mean):

```

0: name() should be declared external:
        - ERC20.name() (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#64-66)
1: symbol() should be declared external:
        - ERC20.symbol() (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#72-74)
2: totalSupply() should be declared external:
        - ERC20.totalSupply() (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#96-98)
3: balanceOf(address) should be declared external:
        - ERC20.balanceOf(address) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#103-105)
4: transfer(address,uint256) should be declared external:
        - ERC20.transfer(address,uint256) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#115-118)
5: allowance(address,address) should be declared external:
        - ERC20.allowance(address,address) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#123-125)
6: approve(address,uint256) should be declared external:
        - ERC20.approve(address,uint256) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#134-137)
7: transferFrom(address,address,uint256) should be declared external:
        - ERC20.transferFrom(address,address,uint256) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#152-156)
8: increaseAllowance(address,uint256) should be declared external:
        - ERC20.increaseAllowance(address,uint256) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#170-173)
9: decreaseAllowance(address,uint256) should be declared external:
        - ERC20.decreaseAllowance(address,uint256) (node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol#189-192)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-external

. analyzed (28 contracts with 75 detectors), 91 result(s) found
[...]

Issue #885, opened 2021-07-07
Hidden results with triage mode doesn't work as expected.


mauricio1802:
I wanted to hide some specific results, but I didn't want to add project unrelated commets to the code, so I tried `--triage-mode` option. It hides the result when I run `slither .`, but if I use `--print human-summary` or `--json file.json` options those results are still counted. Is that ok? I would expect that any analysis run ignored the results in `slither.db.json`.

If I use the `//slither-disable-next-line DETECTOR_NAME` way it works as expected.

montyly:
Hi @mauricio1802 

Thank you for reporting this. Indeed currently the triage only works when the detectors are run directly, and not through the printers.

We will improve this

0xalpharush:
closed by [fix triage prompt not displaying #1435](https://github.com/crytic/slither/pull/1435)

PR #1435, opened 2022-10-18
fix triage prompt not displaying

0xalpharush:
Closes https://github.com/crytic/slither/issues/1354, https://github.com/crytic/slither/issues/922, https://github.com/crytic/slither/issues/885

Right now, all of the prompts do not show until the last result is output... adding a new line makes a prompt show for each detectors' results (idk why this is).


montyly:
Good catch. For info this is because `print` is buffered and not flushed directly without `/n` (or by adding [`,flush=True`](https://docs.python.org/3/library/functions.html#print) to `print`)

PR #1451, opened 2022-11-04
upgradeability: include inherited private variables, ignore immutables

0xalpharush:
The upgradeability check doesn't consider inherited, private variable declarations nor does it consider immutability. This PR addresses that by using `state_variables_ordered` and checking that variables aren't immutable prior to flagging them.
```
// test.sol 
pragma solidity ^0.8.0;
contract A {
    uint private a = 1; // not currently detected 
}
contract B is A {
    uint a = 2;
    uint immutable b = 3; // fine to delegatecall (same as constant)
}
```
Command: `slither-check-upgradeability test.sol B`
Before:
```
B.a (test.sol#6) is a state variable with an initial value.
B.b (test.sol#7) is a state variable with an initial value.
```
After fix:
```
A.a (test.sol#3) is a state variable with an initial value.
B.a (test.sol#6) is a state variable with an initial value.
```

### DIFF
--- a/slither/tools/upgradeability/checks/variable_initialization.py
+++ b/slither/tools/upgradeability/checks/variable_initialization.py
@@ -39,8 +39,8 @@ Using initialize functions to write initial values in state variables.
 
     def _check(self):
         results = []
-        for s in self.contract.state_variables:
-            if s.initialized and not s.is_constant:
+        for s in self.contract.state_variables_ordered:
+            if s.initialized and not (s.is_constant or s.is_immutable):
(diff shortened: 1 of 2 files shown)


0xalpharush:
Probably needs to be done here too https://github.com/crytic/slither/blob/b96beeaa5742d5d7a862e964bc72b3cab67623f4/slither/tools/upgradeability/checks/variables_order.py#L112

montyly:
@0xalpharush: can you make the changes in the other places within this PR?

Issue #1218, opened 2022-05-26
[Bug-Candidate]: Slither unable to generate IR


aga7hokakological:
### Describe the issue:

I have been trying to scan some files and projects with slither but getting same error: 

```
46 | contract NFTMarketplace is AccessControlEnumerable, ReentrancyGuard, IERC721Receiver
   | ^ (Relevant source part starts here and spans across multiple lines).

Impossible to generate IR for AccessControl.supportsInterface
Impossible to generate IR for AccessControl.supportsInterface
Impossible to generate IR for AccessControlEnumerable.supportsInterface
Impossible to generate IR for AccessControlEnumerable.supportsInterface
Impossible to generate IR for AccessControlEnumerable.supportsInterface
Impossible to generate IR for ERC721.supportsInterface
Impossible to generate IR for ERC721.supportsInterface
Impossible to generate IR for ERC721._checkOnERC721Received
Impossible to generate IR for ERC165.supportsInterface
Impossible to generate IR for NFTMarketplace.supportsInterface
Impossible to generate IR for NFTMarketplace.supportsInterface
Impossible to generate IR for NFTMarketplace.supportsInterface
Impossible to generate IR for NFTMarketplace.supportsInterface
Impossible to generate IR for NFTMarketplace.onERC721Received
Traceback (most recent call last):
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/__main__.py", line 744, in main_impl       
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/__main__.py", line 87, in process_all      
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/__main__.py", line 72, in process_single   
    return _process(slither, detector_classes, printer_classes)
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/__main__.py", line 113, in _process        
    detector_results = slither.run_detectors()
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/slither.py", line 201, in run_detectors    
    results = [d.detect() for d in self._detectors]
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/slither.py", line 201, in <listcomp>       
    results = [d.detect() for d in self._detectors]
  File "/home/saurabh/.local/lib/python3.9/site-packages/slither/detectors/abstract_detector.py", line 152, 
in detect
[...]

PR #1454, opened 2022-11-07
(deps) replace pysha3 with pycryptodome

0xalpharush:
move to a maintained package. also will probably help reduce namespace conflicts e.g. https://github.com/crytic/slither/issues/1218

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     python_requires=">=3.8",
     install_requires=[
         "prettytable>=0.7.2",
-        "pycryptodome>=3.15.0",
+        "pycryptodome>=3.4.6",
         "crytic-compile>=0.2.4",
         # "crytic-compile@git+https://github.com/crytic/crytic-compile.git@master#egg=crytic-compile",
     ],
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     python_requires=">=3.8",
     install_requires=[
         "prettytable>=0.7.2",
-        "pysha3>=1.0.2",
+        "pycryptodome>=3.15.0",
         "crytic-compile>=0.2.4",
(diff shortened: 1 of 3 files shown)


slither/utils/function.py
```diff
@@ -9,6 +9,6 @@ def get_function_id(sig: str) -> int:
     Return:
         (int)
     """
-    s = sha3.keccak_256()
-    s.update(sig.encode("utf-8"))
-    return int("0x" + s.hexdigest()[:8], 16)
+    hash = keccak.new(digest_bits=256)
+    hash.update(sig.encode("utf-8"))
+    return int("0x" + hash.hexdigest()[:8], 16)
```
0xalpharush, line 14:
```suggestion
    digest = keccak.new(digest_bits=256)
    digest.update(sig.encode("utf-8"))
    return int("0x" + digest.hexdigest()[:8], 16)
```

setup.py
```diff
@@ -13,7 +13,7 @@
     python_requires=">=3.8",
     install_requires=[
         "prettytable>=0.7.2",
-        "pysha3>=1.0.2",
+        "pycryptodome>=3.15.0",
```
elopez, line 16: Do we need such a strong version requirement? As far as can I see, fairly old versions of the library already support keccak. It might be wise to be flexible with it if it works for what we need, to avoid conflicts with other tools.

elopez (approved):
🎉  awesome! It's high time we replaced pysha3 with something maintained 👍

setup.py
0xalpharush: That is a good idea. What minimum version would you recommend?

elopez: Going by the changelog I would say `>=3.4.6` (may 2017) as that version [fixed a bug with Keccak](https://pycryptodome.readthedocs.io/en/latest/src/changelog.html#may-2017) and I don't see any newer releases mentioning any changes on Keccak support. Please confirm that everything works with the version we end up choosing, just in case 😄

PR #1455, opened 2022-11-07
add immutable-states detector

0xalpharush:
close https://github.com/trailofbits/slither-private/issues/333


slither/detectors/variables/possible_const_state_variables.py
```diff
@@ -71,54 +74,66 @@ def _constant_initial_expression(self, v: Variable) -> bool:
...
 
     def _detect(self) -> List[Output]:
-        """Detect state variables that could be const"""
-        results = []
+        """Detect state variables that could be constant or immutable"""
+        results = {}
+
+        variables = []
+        functions = []
+        for c in self.compilation_unit.contracts:
```
0xalpharush, line 98: filter out openzeppelin?

slither/detectors/variables/possible_const_state_variables.py
```diff
@@ -71,54 +75,66 @@ def _constant_initial_expression(self, v: Variable) -> bool:
...
-        all_variables: Set[StateVariable] = {
-            item for sublist in all_variables_l for item in sublist
+    def _detect(self) -> List[Output]:
+        """Detect state variables that could be constant or immutable"""
+        results = {}
+
+        variables = []
+        functions = []
+        for c in self.compilation_unit.contracts:
+            if is_openzeppelin(c):
```
montyly, line 98: Why removing openzepellin? 

Maybe we should iterate over `compilation_unit.contracts_derived` here, so that we only consider the most derived contracts, and we avoid FP on potential constant variable that are written in a derived contract

montyly (changes requested):
We are missing the candidates that are written in the constructor directly, ex:
```solidity
contract A{
	uint a;
	constructor(uint b) {
		a = b;
	}
}
```

Maybe we should split this detector in two? As adding the logic for the constructor will increase the complexity, it would make sense to differentiate variables can that be determined at compile time (constant), and variables that can be determined at deployment (immutable) into two distinct detectors. 

We can reuse the functions (`_constant_initial_expression` /  `_is_valid_type` / ..)  and mostly change the filtering logic

PR #1460, opened 2022-11-08
support abi.encodeCall

plotchy:
abi.encodeCall is not supported as a `SolidityFunction`/`SolidityCall`. This adds support

@0xalpharush / @montyly  If you have time, can you help explain how to set up an ast_parsing test to me? I plan to add tests, but could use a little guidance

### DIFF
--- a/slither/core/declarations/solidity_variables.py
+++ b/slither/core/declarations/solidity_variables.py
@@ -70,6 +70,7 @@ SOLIDITY_FUNCTIONS: Dict[str, List[str]] = {
     "abi.encodePacked()": ["bytes"],
     "abi.encodeWithSelector()": ["bytes"],
     "abi.encodeWithSignature()": ["bytes"],
+    "abi.encodeCall()": ["bytes"],
     "bytes.concat()": ["bytes"],
     "string.concat()": ["string"],
     # abi.decode returns an a list arbitrary types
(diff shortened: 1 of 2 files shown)


0xalpharush:
I would modify [units_and_global_variables-0.8.12.sol ](https://github.com/crytic/slither/blob/master/tests/ast-parsing/units_and_global_variables-0.8.12.sol). If necessary, create another test case in `tests/ast-parsing` with a newer version of Solidity (not sure if this [bug](https://blog.soliditylang.org/2022/03/16/encodecall-bug/) is relevant).

For each test case, we store a zip file of the compilation e.g. [units_and_global_variables-0.8.0.sol-0.8.12-compact.zip](https://github.com/crytic/slither/blob/master/tests/ast-parsing/compile/units_and_global_variables-0.8.0.sol-0.8.12-compact.zip).  

For existing test cases that are modified, it must be deleted and regenerated. 

For new test cases the zip file needs to be created for the first time. In addition, new test cases must be registered [here](https://github.com/crytic/slither/blob/a41f86739d9d2aad7596b9a0ac0ab2ce644fc2f3/tests/test_ast_parsing.py#L303) with the file name and which solidity versions it should be tested against.

Command to generate zip for test:
```
python ./tests/test_ast_parsing.py --compile
```
Then, you can generate the json file of slither's output:
```
python ./tests/test_ast_parsing.py --compile
```
You can copy commands from here in your local environment https://github.com/crytic/slither/blob/master/CONTRIBUTING.md#parser-tests

Here's what a similar PR ended up looking like https://github.com/crytic/slither/pull/1086. Lmk if you have other questions or need clarification.

montyly:
Thanks @plotchy

PR #1462, opened 2022-11-08
fail-on: rework feature

### LINKED ISSUE #1458: making exclude/ fail-on behavior more intuitive
There's some confusion about the default behavior and prioritization of `--exclude-*` and `--fail-on-*` flags. For example, since `--fail-on-pedantic` is enabled by default, one must pass `--exclude-informational  --exclude-optimization --no-fail-pedantic` to disable informational and optimization level detectors.

Current thinking:
- Decouple fail and exclude 
- Create --fail-on Enum[optimization, informational, low, medium, high]

ref: https://github.com/crytic/slither-action/issues/38, https://github.com/crytic/slither/issues/1408

elopez:
This adds a new `fail_on` config option that can be changed with a mutually exclusive group of argument flags. It also decouples the exclude_* and fail_on flags, so you can do things like `fail_on: pedantic` while disabling optimization findings. Additionally, this adds some new code to detect the old-style config options, migrate their settings, and alert the user.

Fixes #1458


elopez:
This has only been tested very lightly at this time

PR #1470, opened 2022-11-15
Fix `naming-convention` to flag single letter `O` or `I` variable

### LINKED ISSUE #1418: `naming-convention` does not flag `I` or `O` variable
### Describe the issue:

Slither has a detection for usage of

* l - Lowercase letter el
* O - Uppercase letter oh
* I - Uppercase letter eye

https://docs.soliditylang.org/en/v0.8.17/style-guide.html#names-to-avoid

However this detection fails on O and I, as the code considers them "uppercase with underscores" even though there are no underscores.

Relevant code:
* https://github.com/crytic/slither/blob/fba37f2c0c8196079719432d6324e42a1a974399/slither/detectors/naming_convention/naming_convention.py#L52-L54 (will match just uppercase letters as well)
* https://github.com/crytic/slither/blob/fba37f2c0c8196079719432d6324e42a1a974399/slither/detectors/naming_convention/naming_convention.py#L120-L137 (detection code)

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

contract C {
    bool constant l = false;
    bool constant O = false;
    bool constant I = false;
    function f() public pure {}
}
```

### Version:

0.9.0

### Relevant log output:

_No response_

ardislu:
The regex `^[lOI]$` inside `should_avoid_name()` will only match on the single letter variables `l`, `O`, or `I`, which are not recommended by the [style guide](https://docs.soliditylang.org/en/v0.8.17/style-guide.html#names-to-avoid). 

The inner `not is_upper_case_with_underscores()` check is unnecessary and also allows the single letter variables `O` and `I` to pass.

Fixes #1418

### DIFF
--- a/slither/detectors/naming_convention/naming_convention.py
+++ b/slither/detectors/naming_convention/naming_convention.py
@@ -119,22 +119,21 @@ Solidity defines a [naming convention](https://solidity.readthedocs.io/en/v0.4.2
 
             for var in contract.state_variables_declared:
                 if self.should_avoid_name(var.name):
-                    if not self.is_upper_case_with_underscores(var.name):
-                        info = [
-                            "Variable ",
-                            var,
-                            " used l, O, I, which should not be used\n",
-                        ]
+                    info = [
+                        "Variable ",
+                        var,
+                        " used l, O, I, which should not be used\n",
+                    ]
 
-                        res = self.generate_result(info)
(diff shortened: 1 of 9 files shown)


montyly:
great fix @ardislu. Thanks for updating the tests ;)

PR #1475, opened 2022-11-21
Exclude parent private variables from shadow detector

### LINKED ISSUE #1440: [Bug-Candidate]: shadow private variable
### Describe the issue:

I made a small change to the [state-variable-shadow](https://github.com/crytic/slither/wiki/Detector-Documentation#state-variable-shadowing) example. If the variable `_owner` in BaseContract is decorated by private and it cannot be inherited by DerivedContract. The `isOwner` is BaseContract will work properly.

## Suggestion
I suggest that the `shadowing-state` detector exclude private variable shadowing.

## When this case:

When I want to inherit the [openzepplin ERC1155 contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/ERC1155.sol) and override the `_uri` private variable as I would set a different URI for each tokenId and use other feature from oz's contract. 

```solidity
pragma solidity ^0.8.0;

contract BaseContract {
  address private _owner;

  modifier isOwner() {
    require(_owner == msg.sender);
    _;
  }
}

contract DerivedContract is BaseContract {
  address private _owner;

  constructor() {
    _owner = msg.sender;
  }

  function withdraw() external isOwner {
    payable(msg.sender).transfer(0);
  }
}
```

### Version:

0.9.0

### Relevant log output:

```shell
Compilation warnings/errors on 1.sol:
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> 1.sol



DerivedContract._owner (1.sol#13) shadows:
        - BaseContract._owner (1.sol#4)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#state-variable-shadowing

Pragma version^0.8.0 (1.sol#1) allows old versions
solc-0.8.17 is not recommended for deployment
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity

BaseContract._owner (1.sol#4) should be constant
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#state-variables-that-could-be-declared-constant
1.sol analyzed (2 contracts with 81 detectors), 4 result(s) found
```

devtooligan:
closes #1440 

Hello, first time contributor here.  Few questions.

1. wrt the approach, @0xalpharush had suggested using `state_variables` API instead of `state_variables_declared` but I tried changing that in both places in this function and it did not seem fix the issue.  Please let me know if my approach is ok.
2. I'm guessing I should write a test?  I'm sure I can figure out where to put the test.
3. Based on what I read in the contributing guidelines, I'm assuming w this pr I should only address the top comment in the [ticket](https://github.com/crytic/slither/issues/1440#issue-1422052660), and not @plotchy's [additional comment](https://github.com/crytic/slither/issues/1440#issuecomment-1290947529), right?  I can moved that other comment to another ticket and do a separate pr.
4. Are there any conventions you use for naming commits? Conventional commits ok?
5. Any conventions you use for naming branches?

Thanks


slither/detectors/shadowing/state.py
```diff
@@ -12,7 +12,7 @@ def detect_shadowing(contract: Contract):
     variables_fathers = []
     for father in contract.inheritance:
         if any(f.is_implemented for f in father.functions + father.modifiers):
-            variables_fathers += father.state_variables_declared
+            variables_fathers += [var for var in father.state_variables_declared if var.visibility != "private"]
 
     for var in contract.state_variables_declared:
```
0xalpharush, line 17: I think the heuristic should be updated to "are there any **public** state variables in the child contract that shadow **any** state variable in the parent". This would warn on instances where a variable is shadowed and they're not both private as the author suggested. Right now, this change no longer warns when a private variable is shadowed and there's a public one e.g. https://github.com/crytic/slither/blob/97b60d6115ec71264b9da4c63f60e92c5ae64c81/tests/detectors/shadowing-state/0.7.5/public_gap_variable.sol
```suggestion
            variables_fathers += father.state_variables_declared 

    for var in contract.state_variables_entry_points:
```
`state_variables_entry_points` is defined here https://github.com/crytic/slither/blob/master/slither/core/declarations/contract.py#L320

devtooligan: @0xalpharush   @noyyyy what do think about my comment above?  how should i proceed?

devtooligan: @montyly unless I'm mistaken, it appears the code wouldn't compile in cases where we would want this detector to to give a warning. and the only times it does compile and give the warning are cases where we want to suppress the warning

Issue #1476, opened 2022-11-21
Refactor `_uses_vulnerable_solc_version()` onto the base AbstractDetector class


devtooligan:
### Describe the desired feature

I was looking at ticket https://github.com/crytic/slither/issues/1424 and trying to figure out how we would limit the solc versions that the detector would run on.  

I noticed one place we already do this is with `_uses_vulnerable_solc_version()` in [enum_conversion.py](https://github.com/crytic/slither/blob/master/slither/detectors/compiler_bugs/enum_conversion.py#L10-L19).

I was thinking we could move that logic to the `AbstractDetector` base class and add a class attribute `VULNERABLE_SOLC_VERSIONS` that defaults to [].  In the main `detect()` on the base class we could add logic which checks if the `self.compilation_unit.solc_version` is included in the list. (Skip the check if the list is empty)

This way, whenever we want to limit the vulnerable solc versions for a detector, we just add the new `VULNERABLE_SOLC_VERSIONS` attribute to the subclass, and done.  No need to change any logic. 

I think there may be other detectors that need to be limited to certain solc versions like this too, so this should make that easier.

I am planning to do this work before working on #1424 so am documenting this here and will submit separate prs.

Issue #1424, opened 2022-10-14
Restrict the pre-declaration usage of local variables detector (`variable-scope`) to Solidity 0.4


CodeSandwich:
### Describe the desired feature

The pre-declaration usage of local variables detector checks a property which is impossible to violate since Solidity 0.5, see the [last point of the 0.5 changelog](https://docs.soliditylang.org/en/latest/050-breaking-changes.html#semantic-and-syntactic-changes). The [exploit scenario example](https://docs.soliditylang.org/en/latest/050-breaking-changes.html#semantic-and-syntactic-changes) doesn't even compile with Solidity >=0.5. Any `variable-scope`s found when using Solidity >=0.5 are obviously false positives and unfortunately the detector does that quite often. All it takes is a function with the same variable name used twice in two separate scopes, e.g.
```solidity
function foo() public {
    {
        uint a = 0;
    }
    {
        uint a = 1;
    }
}
```
It'd be great if this detector was disabled altogether when analyzing codebases using Solidity >=0.5.

montyly:
Hi @CodeSandwich , that's a good point. We have to update the detector

0xalpharush:
My mistake this wasnt fixed in the linked pull requests

0xalpharush:
closed by https://github.com/crytic/slither/pull/1731

PR #1477, opened 2022-11-21
refactor: add VULNERABLE_SOLC_VERSIONS and logic

devtooligan:
addresses #1476 
precursor to #1424

### DIFF
--- a/slither/detectors/abstract_detector.py
+++ b/slither/detectors/abstract_detector.py
@@ -64,7 +64,7 @@ class AbstractDetector(metaclass=abc.ABCMeta):
     # list of vulnerable solc versions as strings (e.g. ["0.4.25", "0.5.0"])
     # if this list is not empty then the detector will not run unless the solc version is on the list
     # an empty list means that the detector will run on any solc version
-    VULNERABLE_SOLC_VERSIONS = []
+    VULNERABLE_SOLC_VERSIONS: List[str] = []
 
     def __init__(
         self, compilation_unit: SlitherCompilationUnit, slither: "Slither", logger: Logger
--- a/slither/detectors/abstract_detector.py
+++ b/slither/detectors/abstract_detector.py
@@ -61,6 +61,11 @@ class AbstractDetector(metaclass=abc.ABCMeta):
 
     STANDARD_JSON = True
 
+    # list of vulnerable solc versions as strings (e.g. ["0.4.25", "0.5.0"])
+    # if this list is not empty then the detector will not run unless the solc version is on the list
+    # an empty list means that the detector will run on any solc version
+    VULNERABLE_SOLC_VERSIONS = []
+
     def __init__(
         self, compilation_unit: SlitherCompilationUnit, slither: "Slither", logger: Logger
     ):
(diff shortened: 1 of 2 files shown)


slither/detectors/abstract_detector.py
```diff
@@ -139,6 +144,12 @@ def _log(self, info: str) -> None:
         if self.logger:
             self.logger.info(self.color(info))
 
+    def _uses_vulnerable_solc_version(self) -> bool:
+        if not self.VULNERABLE_SOLC_VERSIONS:
+            return False
+
+        return self.compilation_unit.solc_version in self.VULNERABLE_SOLC_VERSIONS
```
0xalpharush, line 151:
```suggestion
    def _uses_vulnerable_solc_version(self) -> bool:
        if self.VULNERABLE_SOLC_VERSIONS:
            return self.compilation_unit.solc_version in self.VULNERABLE_SOLC_VERSIONS

        return True
```
The base case should be true so that the detector runs on all versions unless specified in the detector implementation.

0xalpharush:
I think this approach makes sense. There's probably other areas it could be used e.g. the [external function detector](https://github.com/crytic/slither/blob/master/slither/detectors/functions/external_function.py#L138).

slither/detectors/abstract_detector.py
devtooligan: oops, no wonder all my tests are failing. thank you 🙏🏽

devtooligan:
@0xalpharush fixed and ready for re-review.

question: do you prefer 1 commit per pr?  so `github commit --amend` and `github push --force-with-lease` for additional changes? or do you prefer to keep all the interim history commits?  i assumed the former.

0xalpharush:
@devtooligan There's not a strict style atm. I think force pushing is fine for small commits but if you make large changes and the commits aid in reviewing the PR, leaving the interim history makes sense.

0xalpharush (approved):

montyly:
I would recommend against force push, or changing anything that was pushed on github (see [Linus's point of view on rebasing](https://www.mail-archive.com/dri-devel@lists.sourceforge.net/msg39091.html))

Github allows to squash commit when we merge the PR, so it's ok to have multiple commits in the PRs ;)

slither/detectors/abstract_detector.py
```diff
@@ -61,6 +61,11 @@ class AbstractDetector(metaclass=abc.ABCMeta):
 
     STANDARD_JSON = True
 
+    # list of vulnerable solc versions as strings (e.g. ["0.4.25", "0.5.0"])
+    # if this list is not empty then the detector will not run unless the solc version is on the list
+    # an empty list means that the detector will run on any solc version
+    VULNERABLE_SOLC_VERSIONS = []
```
montyly, line 67: Maybe we could add the type here: `List[str]` to avoid confusion

montyly (commented):
This is a great idea, I think we can clean up a lot of detectors following this schema. Thanks @devtooligan

montyly:
Note: when merging the PR, we need to update https://github.com/crytic/slither/wiki/Adding-a-new-detector

slither/detectors/abstract_detector.py
devtooligan: good call, added.

devtooligan:
@montyly wrote:
> Note: when merging the PR, we need to update https://github.com/crytic/slither/wiki/Adding-a-new-detector

@0xalpharush maybe you or someone else can help with this.  Github wiki is not set up very well for open source contributions from outsiders 😄   

We will need to update the snippet in the wiki with:
```
VULNERABLE_SOLC_VERSIONS: List[str] = []
```
Then down below in the notes maybe add something like:
```
`VULNERABLE_SOLC_VERSIONS` lets you specify which versions of solc
are vulnerable.  If this list is not empty, the detector will
only run if the solc version is included on the list.
```

devtooligan:
@0xalpharush @montyly Requested changes made.  Aside from updating the wiki, this should be good for final re-review and merge.

PR #1484, opened 2022-11-28
Recommend reading variable without additional STATICCALL 

0xalpharush:
Closes https://github.com/trailofbits/slither-private/issues/319. Whenever the this.variable syntax is used, it results in an additional STATICCALL. We can recommend replacing this with an operation that is a direct SLOAD. Note, this isn't detect for old versions of solidity like 0.4.25 since they didn't use STATICALL (see is_static_call in HighLevelCall's API).

Replaces https://github.com/trailofbits/slither-private/pull/322


slither/detectors/variables/var_read_using_this.py
```diff
@@ -0,0 +1,56 @@
...
+from slither.core.declarations import Function, SolidityVariable
+from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
+from slither.slithir.operations.high_level_call import HighLevelCall
+
+
+class VarReadUsingThis(AbstractDetector):
+    ARGUMENT = "var-read-using-this"
+    HELP = "Contract reads its own variable using `this`"
+    IMPACT = DetectorClassification.OPTIMIZATION
+    CONFIDENCE = DetectorClassification.MEDIUM
```
montyly, line 12: What is the reason for the medium confidence?

0xalpharush: I didn't give it much thought. I suppose it can be high as I don't think it will be noisy

PR #1498, opened 2022-12-06
Add Codex vuln detector

montyly:
The detector requires:
- The user set the flag `--codex` (meaning that codex is not ran by default)
- `openai` must be installed, and `OPENAI_API_KEY` set

The detector works at the contract level, and send the whole contract body to codex

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -302,7 +302,10 @@ def parse_args(
     parser.add_argument("filename", help=argparse.SUPPRESS)
 
     parser.add_argument(
-        "--codex", help="Enable codex (require an OpenAI API Key)", action="store_true", default=False
+        "--codex",
+        help="Enable codex (require an OpenAI API Key)",
+        action="store_true",
+        default=False,
     )
 
     cryticparser.init(parser)
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -166,7 +166,6 @@ def process_from_asts(
 def get_detectors_and_printers() -> Tuple[
     List[Type[AbstractDetector]], List[Type[AbstractPrinter]]
 ]:
-
     detectors_ = [getattr(all_detectors, name) for name in dir(all_detectors)]
     detectors = [d for d in detectors_ if inspect.isclass(d) and issubclass(d, AbstractDetector)]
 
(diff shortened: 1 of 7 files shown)


slither/detectors/functions/codex.py
```diff
@@ -0,0 +1,80 @@
...
+
+        for contract in self.compilation_unit.contracts:
+            prompt = "Is there a vulnerability in this solidity contracts?\n"
+            src_mapping = contract.source_mapping
+            content = contract.compilation_unit.core.source_code[src_mapping.filename.absolute]
+            start = src_mapping.start
+            end = src_mapping.start + src_mapping.length
+            prompt += content[start:end]
+            answer = openai.Completion.create(  # type: ignore
+                model="text-davinci-003", prompt=prompt, temperature=0, max_tokens=200
```
devtooligan, line 62: `model` - I think we sb using model [code-davinci-002](https://beta.openai.com/docs/models/codex)  (it's also free rn in limited beta)
`temperature` - I think we should bump this up a little, with it set to `0` I was getting a lot of "There are no vulnerabilities in this contract."  
`max_tokens` - `code-davinci-002`  supports max requests (prompt + result) up to 8,000 tokens (compared with 4,000 for the "text-davinci-003") so we can comfortably bump this up by a lot.

I'm working on a pr that parameterizes all of these things we can merge into this one...

slither/detectors/functions/codex.py
```diff
@@ -0,0 +1,80 @@
...
+        api_key = os.getenv("OPENAI_API_KEY")
+        if api_key is None:
+            logging.info(
+                "Please provide an Open API Key in OPENAI_API_KEY (https://beta.openai.com/account/api-keys)"
+            )
+            return []
+        openai.api_key = api_key
+
+        for contract in self.compilation_unit.contracts:
+            prompt = "Is there a vulnerability in this solidity contracts?\n"
```
devtooligan, line 55: Not sure why, but no matter what contract I used, I kept getting "No vulnerabilities found".  When I changed the prompt up to something like "Analyze this contract and report the vulns" It started working for me.

slither/detectors/functions/codex.py
devtooligan, line 62: I've been playing around with these numbers temperature and max_tokens but haven't noticed a big difference.  I think it would be nice if these were all parameterized.  I'll be submitting a pr based on this branch that does that.

slither/__main__.py
```diff
@@ -301,6 +301,13 @@ def parse_args(
 
     parser.add_argument("filename", help=argparse.SUPPRESS)
 
+    parser.add_argument(
+        "--codex",
+        help="Enable codex (require an OpenAI API Key)",
+        action="store_true",
+        default=False,
```
devtooligan, line 308: Should we add this key to the `defaults_flag_in_config` dict?

PR #1499, opened 2022-12-06
feat: Codex vuln detector -  devtooligan updates

devtooligan:
Highlights of this pr:
- Parameterize the inputs to the OpenAI codex call: `model`, `temperature`, and `max_tokens`
- Add `--codex-contracts` argument which takes a comma-delimited list of contracts and, if specified, it will limit the Codex detector to only those contracts.  I found this a great way to save money when working with a large contract with a deeply nested inheritance tree.  Someday, may want to expand this feature to all detectors.
- It also changes the prompt to something I was having more luck with.  It no longer asks a "Yes" or "No" question, so the prompt now also requires the use of a keyword if vulns are found.

I did spend some time implementing a "retries" system.  But for some reason, the retried "queries" were all identical.  I guess I was hoping it would exhibit similar behavior to the chatbot when "retry" is pressed, which results in a different response.  So I've removed that feature but will continue to look into it.

### DIFF
--- a/slither/detectors/functions/codex.py
+++ b/slither/detectors/functions/codex.py
@@ -7,6 +7,7 @@ from slither.utils.output import Output
 
 logger = logging.getLogger("Slither")
 
+VULN_FOUND = "VULN_FOUND"
 
 class Codex(AbstractDetector):
     """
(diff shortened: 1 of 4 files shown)


devtooligan:
Additional pr based on this branch submitted. https://github.com/crytic/slither/pull/1498

slither/detectors/functions/codex.py
```diff
@@ -52,29 +53,63 @@ def _detect(self) -> List[Output]:
         openai.api_key = api_key
 
         for contract in self.compilation_unit.contracts:
-            prompt = "Is there a vulnerability in this solidity contracts?\n"
+            if self.slither.codex_contracts != "all" and contract.name not in self.slither.codex_contracts.split(","):
+                continue
+            prompt = "Analyze this Solidity contract and find the vulnerabilities. If you find any vulnerabilities, begin the response with {}".format(VULN_FOUND)
```
devtooligan, line 58: We could probably parameterize the prompt.

Also, maybe we could just always display the Codex response, even if it doesn't find anything.  This would simplify it so we don't need adjust the prompt or look for a special word.

slither/detectors/functions/codex.py
```diff
@@ -52,29 +53,63 @@ def _detect(self) -> List[Output]:
...
-                                contract,
-                                "\n",
-                                answer["choices"][0]["text"],
-                                "\n",
-                            ]
-
-                            res = self.generate_result(info)
-                            results.append(res)
-
+            logging.info("Querying OpenAI")
```
devtooligan, line 64: I wasn't 100% sure what to do with logging/printing . I noticed in this file we used the logger but it didn't display anything when Slither was run.  And I noticed in other parts of the tool we use `print`.

slither/detectors/functions/codex.py
```diff
@@ -52,29 +53,63 @@ def _detect(self) -> List[Output]:
...
+            print("Querying OpenAI")
+            answer = ""
+            res = {}
+            try:
```
devtooligan, line 68: I added this try/catch because I was hitting the max token limit with a large contract in the prompt.  Another reason to consider running this per-function.

slither/detectors/functions/codex.py
```diff
@@ -52,29 +53,63 @@ def _detect(self) -> List[Output]:
         openai.api_key = api_key
 
         for contract in self.compilation_unit.contracts:
-            prompt = "Is there a vulnerability in this solidity contracts?\n"
```
devtooligan, line 55: Above in line 40, the `import openai` inside of this function was problematic when I tried to extract the OpenAI query logic into a separate function.  I think it scopes the import to the function so I wasn't able to access it.  

Maybe we can move the `import` (wrapped in a try/except) to the top level?  We could still do something here that checks whether it's been installed.

montyly:
I merged the improvements in #1498, some notes:
- The logging was not working because I used `logging` instead of `logger` (...) 
- I refactored part of the codex logic into a `utils.codex`. We can then use the similar logic for https://github.com/crytic/slither/pull/1494. Right now its only for logging/import openai. We might add general parsing of openai queries, depending on how we are going to use openai in slither
- I added a `--codex-log` flag. This will create files in `crytic_export/codex/...`, with all the queries/answers. This should help debugging

PR #1503, opened 2022-12-15
Add `state_variable_slots` property to contract.py

webthethird:
I am using this in my fork, and thought it would be a nice and useful enough addition on its own to make opening a PR worthwhile. I will be using it in a new upgradeability check I am working on, which will detect all of the differences between V1 and V2 of an implementation contract, for informational purposes and to feed into the differential fuzzing test framework I aim to design. Knowing the storage slot for each state variable will be useful when using cheat codes to read from a contract's storage and compare the values between two versions.

### DIFF
--- a/slither/core/declarations/contract.py
+++ b/slither/core/declarations/contract.py
@@ -388,6 +388,23 @@ class Contract(SourceMapping):  # pylint: disable=too-many-public-methods
                 variables_used[variable].add(variable)
             self._state_variables_used_in_reentrant_targets = variables_used
         return self._state_variables_used_in_reentrant_targets
+    
+    @property
+    def state_variable_slots(self) -> Dict["StateVariable", int]:
+        """
+        Returns a dictionary mapping StateVariable objects to their storage slot
+        """
+        current_slot = 0
+        current_bytes = 0
+        variable_slots = {}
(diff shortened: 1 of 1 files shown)


0xalpharush:
Hi @webthethird, this sounds very interesting and could likely reuse the existing work on [slither-read-storage](https://github.com/crytic/slither/tree/master/slither/tools/read_storage) and [slither-check-upgradeability](https://github.com/crytic/slither/tree/master/slither/tools/upgradeability)! Fwiw, you can retrieve the same information from the compilation unit's method, [`storage_layout_of`](https://github.com/crytic/slither/blob/master/slither/core/compilation_unit.py#L276). If you have questions on how to do anything with Slither or run into issues, please reach out on slack (it looks like maybe you're in the Empire Hacking slack channel).

webthethird:
> Fwiw, you can retrieve the same information from the compilation unit's method, [`storage_layout_of`](https://github.com/crytic/slither/blob/master/slither/core/compilation_unit.py#L276). 

Oh interesting, I didn't think to look for that in the compilation unit! I'll look into using that instead then. My intuition was I'd find it in the `Contract` class if it existed. I've been working with slither-check-upgradeability, but didn't realize there's also a slither-read-storage!

PR #1504, opened 2022-12-15
Fix `ExtraVariablesProxy` upgradeability check

webthethird:
... and by inheritance, also fix `ExtraVariablesNewContract`. 
This check was reporting too many variables, because it was starting at `idx = len(order2) - len(order1)`. 
So, if the new contract (or proxy) had 12 variables, and the original contract had 10, it was reporting every variable starting at `idx = 2`, whereas it should have only been reporting those starting at `idx = len(order1)`.


webthethird:
Any idea why the Dapp test check failed here? This PR is a very simple fix, limited to just the one line, so I can't imagine the failed test has anything to do with the small change I made.

0xalpharush:
@webthethird There's a fix in `dev` if you'll merge it into your branch. It is unrelated to your changes

webthethird:
> @webthethird There's a fix in `dev` if you'll merge it into your branch. It is unrelated to your changes

Done! And thanks @elopez for switching the base branch, makes sense to me.

montyly:
great catch. Thanks @webthethird

Issue #1427, opened 2022-10-14
echidna-printer: add result of binary/unary expressions with constant operands to output


0xalpharush:
I think the best way to deal with this is just to precompute the value of binary/unary operations when the operads are constant. This should be done from the slither side, and add the result into `constants_used`. Echidna should be able to pick them without any modification.

_Originally posted by @ggrieco-tob in https://github.com/crytic/echidna/issues/636#issuecomment-1241905188_

PR #1508, opened 2022-12-16
fold binary expressions with constant operands for fuzzing guidance

0xalpharush:
close https://github.com/crytic/slither/issues/1427

### DIFF
--- a/slither/visitors/expression/constants_folding.py
+++ b/slither/visitors/expression/constants_folding.py
@@ -1,4 +1,11 @@
-from slither.core.expressions import BinaryOperationType, Literal, UnaryOperationType
+from slither.core.expressions import (
+    BinaryOperationType,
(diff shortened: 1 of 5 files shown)


slither/visitors/expression/constants_folding.py
```diff
@@ -115,9 +154,13 @@ def _post_tuple_expression(self, expression):
...
                 return
         raise NotConstant
 
     def _post_type_conversion(self, expression):
-        raise NotConstant
+        cf = ConstantFolding(expression.expression, self._type)
+        expr = cf.result()
+        assert isinstance(expr, Literal)
+        set_val(expression, convert_string_to_fraction(expr.converted_value))
+        return
```
0xalpharush, line 166:
```suggestion
```

PR #1513, opened 2022-12-16
Printer outputs sorted

### LINKED ISSUE #1374: Deterministic printer output

When running:
`slither <file> --print function-summary`
and
`slither <file> --print vars-and-auth`

The order or values in the arrays changes between runs. Examples below:
Screen Shot 2022-09-12 at 11 49 10 AM
Screen Shot 2022-09-12 at 11 50 12 AM

The intention was to run slither printers and write to files when updating submodules to check diffs, but since the order of the outputs in the arrays in non-deterministic, the diffs are too large to be usable.

Easiest solution in my opinion would be to simply alphabetically order the array values.

bart1e:
Outputs of several printers sorted so that they print consistent results.
Fixes #1374 
I have tried to find all places where the outputs might not possibly be deterministic, but if you find another one, please let me know.

### DIFF
--- a/slither/printers/functions/authorization.py
+++ b/slither/printers/functions/authorization.py
@@ -54,8 +54,8 @@ class PrinterWrittenVariablesAndAuthorization(AbstractPrinter):
                 table.add_row(
                     [
                         function.name,
-                        str(state_variables_written),
-                        str(msg_sender_condition),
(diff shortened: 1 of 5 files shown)


montyly:
This is great, thanks @bart1e

PR #1516, opened 2022-12-20
In implement of signature, no return check and no use nonce.

godsangsoo:


montyly:
Thanks @godsangsoo for the contribution

I am closing this PR given that we have a new iteration through: https://github.com/crytic/slither/pull/2015

PR #1518, opened 2022-12-20
Begin improvements to upgradeable proxy detection

webthethird:
This is the first of many PRs I plan on submitting, which should improve the detection of upgradeable proxy contracts. @0xalpharush suggested that I do this incrementally, since the fork I created over a year ago (for my survey of proxy patterns) has accumulated a lot of code. 

This first PR basically renames `Contract.is_upgradeable_proxy` as `Contract.is_proxy`, and introduces a new `Contract.is_upgradeable_proxy` which, for now, does little more than confirm that any delegate target variable found in a LowLevelCall is not in fact constant, immutable or a hard-coded literal. In time, `Contract.is_upgradeable_proxy` should always find the delegate target variable (even in assembly), and find its setter/getter functions to confirm upgradeability.

I will add an example or two of proxy contracts that use `delegatecall` but should not be considered upgradeable soon, so it is clear what the problem to be solved is and so it can be tested.


webthethird:
Actually, the Proxy contract in [`/tests/check-upgradeability/proxy.sol`](https://github.com/crytic/slither/blob/dev/tests/check-upgradeability/proxy.sol) is itself a good example of a proxy that should not be considered upgradeable, since there is no setter for updating `address destination`.

In this first PR, the test script [`ci_test_upgradeability`](https://github.com/crytic/slither/blob/dev/scripts/ci_test_upgradability.sh) passes with no need for changes. However, my end goal is to require at least a setter in order for `is_upgradeable_proxy` to be true, so it will become necessary to use a modified version of the proxy currently used in the CI script.

webthethird:
This kind of thing belongs in `slither.utils.upgradeability`, not the core Contract class (see #1757). Though I would still argue that `Contract.is_upgradeable_proxy` should be renamed `is_proxy` and the new `is_upgradeable_proxy` should search for an implementation address setter.

PR #1519, opened 2022-12-21
slot-collision

godsangsoo:


0xalpharush:
This is currently detected by: `slither-check-upgradeability slot_collision.sol impl --proxy-name Proxy`. We've been working on a way to automate this with code comments (https://github.com/crytic/slither/pull/1522). Just need to connect the comments to slither-check-upgradeability if that's something you're interested in working on.

PR #1520, opened 2022-12-21
price-manipulation

godsangsoo:
It is known that using spot price is vulnerable to price manipulation.


0xalpharush:
I think this is something that is better suited for manual review and would result in too many false positives.

PR #1523, opened 2022-12-21
dangerous-external-call

godsangsoo:
Many incidents are caused by arbitary external call.


0xalpharush:
This should be detected and flagged as higher severity, but it's only detected as reentrancy-benign: `slither ftm:0x660184CE8AF80e0B1e5A1172A16168b15f4136bF --detect reentrancy-benign`. We may be able to infer that calls to `transferFrom`/ `balanceOf` is equivalent to a read-after-write by modeling the storage of the other contract.

PR #1525, opened 2022-12-21
arbitary calldata

godsangsoo:
In calling call and delegatecall, calldata should be verified.


0xalpharush:
I think this is highly contextual and as such would result in a large number of false positives. Fwiw, we do have a detector `controlled-delegatecall` which is only concerned with `delegatecall` since using calldata in a `call` is done commonly but is more dangerous for `delegatecall`

PR #1526, opened 2022-12-21
access-control

godsangsoo:
When storage variable is written, access-control could be necessary.


0xalpharush:
I believe this is similar to the existing `protected-vars`. Perhaps they can be combined and the heuristics fine tuned.

0xalpharush:
Closing this as we have a way to check for access control for state writes:
```
    /// @custom:security write-protection="onlyOwner()"
    address owner;
    function set_protected() public onlyOwner(){
        owner = msg.sender;
    }
    function set_not_protected() public{
        owner = msg.sender;
    }
```
If this doesn't cover your use case, would you open an issue with an example bug and describe how you think Slither could improve please?

PR #1527, opened 2022-12-21
voting-amplificaiton

godsangsoo:
In implement of on-chain governance, copy & paste could make it improper delegation of voting power.


0xalpharush:
This looks like it focuses on a specific implementation of a bug versus detecting a bug pattern. We're focused on the latter with detectors considering it's ran on a wide variety of codebases. Feel free to open an issue if you have an idea how to make this more general!

Issue #936, opened 2021-09-07
slither-check-upgradeability: allow to filter checks


montyly:
Add flags to `slither-check-upgradeability` to disable some detectors per name, or per detectors impact, or to enable only some of the detectors (per name/impact)

PostMan56:
Could this be similarly done like how the flags --detect and --exclude are in slither/__main__.py?

webthethird:
I'll likely take this one on, since I'll probably need it for my project

webthethird:
This is done! See #1532 @montyly

PR #1532, opened 2022-12-23
Filter upgradeability checks by name/impact

webthethird:
Implements #936.

Follows the same approach as the equivalent argument options in the main Slither module, with options to selectively enable/disable checks by name using comma separated lists, and to exclude checks according to impact level.

Does not currently provide an option to enable only a given impact level (i.e., the opposite of `--exclude-<IMPACT>`), but implementing this would be trivial if desired.


montyly:
This is great, thanks @webthethird

PR #1533, opened 2022-12-27
Uninitialised local vars bug fixed

### LINKED ISSUE #1505: [Bug-Candidate]: uninitialized-local-variables FP in destructuring assignment
### Describe the issue:

Running slither on the provided code example produces a false positive regarding an uninitialized local variable.

The slither parser renames variables so that multiple declarations in the same function don't have the same name. Printing the IR of the provided code example shows that in the second branch, the returned identifier is renamed to `q_scope_0` while the identifier in the declaration remains `q`. The declaration identifier should have been renamed to `q_scope_0`. I think the solution to this might be to copy the `id` field from the json `ast` [here](https://github.com/crytic/slither/blob/master/slither/solc_parsing/declarations/function.py#L772-L775) so that it can be used for renaming lookup.
 

pragma solidity 0.8.13;

contract Test
{
    function f1() public view returns (int, string memory)  {
        return (0,"hello");
    }

    function f2(bool a) public view returns (string memory)  {
        if (a) {
            (int x, string memory z) = f1();
            return z;
        } else {
            (int x, string memory z) = f1();
            return z;
        }

    }
}

### Version:

0.9.1

### Relevant log output:

```shell
Test.f2(bool).q_scope_0 (test.sol#15) is a local variable never initialized
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#uninitialized-local-variables
```

bart1e:
Fixes #1505.
There is a bug in `_parse_variable_definition` when handling tuples in expressions like `(type1 a, type2 b) = f()`.
It manifests itself when we have two variables named the same in different scopes and when they are defined in an expression like the one above.
The problem is that `slither` renames variables with the same names in different scopes, but while handling such expressions, an additional expression is created, but it contains old variable names. Code responsible for doing this is depicted below:
```
identifier = {
     "nodeType": "Identifier",
     "src": v["src"],
     "name": v["name"],
     "typeDescriptions": {"typeString": v["typeDescriptions"]["typeString"]},
}
```
It causes `slither` to think that the same variables are referenced in all scopes and it makes both `slithir` representation and further variable analysis incorrect.
I have suggested the simplest fix, that is, to modify the `src` variable so that the code executed afterwards (pasted above) creates an `Expression` with correct variable names. But, if we don't want to modify `src`, we may store all new variable names in a list and then use it when initialising the `identifier` variable.


0xalpharush:
Can you add the example from OP's issue as a regression test and document why the "name" is being overwritten in a code comment please?

bart1e:
@0xalpharush I have added both tests as requested. 
I also tried to explain the fix in the comment, but I'm not sure it's clear - I tried to be concise, but it's hard to explain the bug without going deeply into the details. I have linked this PR in the comment just in case.

duckki:
Ooo, I found this change worked, too.

https://github.com/CertiKProject/slither-certik/commit/08faa901fc427ad458e8a382c2ad3ec024697a8f:
```
diff --git a/slither/solc_parsing/declarations/function.py b/slither/solc_parsing/declarations/function.py
index 13037521..f76fb23a 100644
--- a/slither/solc_parsing/declarations/function.py
+++ b/slither/solc_parsing/declarations/function.py
@@ -770,6 +770,7 @@ class FunctionSolc(CallerContextExpression):
                     for v in variables:
                         identifier = {
                             "nodeType": "Identifier",
+                            "referencedDeclaration": v["id"],
                             "src": v["src"],
                             "name": v["name"],
                             "typeDescriptions": {"typeString": v["typeDescriptions"]["typeString"]},
```

0xalpharush:
Replaced by https://github.com/crytic/slither/pull/1912

Issue #1315, opened 2022-07-29
slither-read-storage: allow for storage diff


montyly:
### Describe the desired feature

We could allow the tool to build a diff made by a transaction. The idea is coming from https://github.com/banteg/storage-layout (thanks to @banteg)

PR #1534, opened 2022-12-28
List storage differences between two blocks.

glarregay-tob:
This PR implements logic to calculate the difference between storage slots at two different blocks.

A new command line option (`--diff-block`) is added to `slither-read-storage`. This will allow to compare between storage state at the block specified by `--block` (`latest`, by default) and `--diff-block`.

Using this option will print two tables after execution: the first one shows the values that haven't changed, the second one the modified values.

Should address #1315 when complete.

### DIFF
--- a/slither/tools/read_storage/__main__.py
+++ b/slither/tools/read_storage/__main__.py
@@ -150,10 +150,17 @@ def print_diff_table(orig_block: int, diff_block: int, orig_info: SlotInfo, diff
         else:
             table_equal.add_row(row)
 
-    print("\n\nSlots with same value:\n======================\n")
-    print(table_equal)
-    print("\n\nSlots with different value:\n===========================\n")
-    print(table_different)
+    if len(table_equal._rows) > 0:
+        print("\n\nSlots with same value:\n======================\n")
+        print(table_equal)
+    else:
+        print("\n\nNo slots with same value found.\n===============================\n")
+
+    if len(table_different._rows) > 0:
+        print("\n\nSlots with different value:\n===========================\n")
+        print(table_different)
+    else:
+        print("\n\nNo slots with different value found.\n====================================\n")
 
 
 def main() -> None:
@@ -215,6 +222,7 @@ def main() -> None:
         srs.walk_slot_info(srs.get_slot_values)
 
     if args.diff_block:
+        orig_block = srs.block
         slots_block_orig = copy.deepcopy(srs.slot_info)
 
         srs.block = diff_block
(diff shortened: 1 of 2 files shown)


slither/tools/read_storage/__main__.py
```diff
@@ -157,6 +212,13 @@ def main() -> None:
         assert args.rpc_url
         srs.walk_slot_info(srs.get_slot_values)
 
+    if args.diff_block:
+        slots_block_orig = copy.deepcopy(srs.slot_info)
+        srs.block = diff_block
```
0xalpharush, line 217: This doesn't account for dynamic arrays (that would require calling `get_storage_layout` again) because the old block would have been used https://github.com/crytic/slither/blob/7ea572723b4eddb18e716c4ae94e0d2100791127/slither/tools/read_storage/read_storage.py#L602-L622

glarregay-tob: Thanks for the info. I fixed it assuming that `get_all_storage_variables` should not be called again, as I understand that storage layout won't change. Is that correct?

By chance, do you know of any verified contract that uses dynamic arrays, to test against it?

slither/tools/read_storage/__main__.py
```diff
@@ -133,6 +187,16 @@ def main() -> None:
     except ValueError:
         srs.block = str(args.block or "latest")
 
+    if args.diff_block:
```
0xalpharush, line 190: Is there a way to avoid checking ars.diff_block twice?

slither/tools/read_storage/__main__.py
```diff
@@ -157,6 +221,21 @@ def main() -> None:
...
+        srs.block = diff_block
+        # This takes into account differences in dynamic arrays
+        if args.variable_name:
+            srs.get_target_variables(**vars(args))
+        else:
+            srs.get_storage_layout()
+        srs.walk_slot_info(srs.get_slot_values)
+        slots_block_diff = copy.deepcopy(srs.slot_info)
+
+        print_diff_table(orig_block, diff_block, slots_block_orig, slots_block_diff)
```
0xalpharush, line 237: Can you move this logic out of __main__ and do something like `srs.walk_slot_info(srs.convert_slot_info_to_rows)`

slither/tools/read_storage/__main__.py
```diff
@@ -104,11 +107,62 @@ def parse_args() -> argparse.Namespace:
...
+    )
+
     cryticparser.init(parser)
 
     return parser.parse_args()
 
 
+def print_diff_table(orig_block: int, diff_block: int, orig_info: SlotInfo, diff_info: SlotInfo):
+
+    if orig_block == "latest" or orig_block > diff_block:
```
0xalpharush, line 123: It seems like this function shouldn't sort its arguments (then its misleading that one arg is called `orig_block`).

0xalpharush:
To test this I would add a case here https://github.com/crytic/slither/blob/master/tests/test_read_storage.py, modify https://github.com/crytic/slither/blob/master/tests/storage-layout/storage_layout-0.8.10.sol#L35 to have a setter, and check that the diff shows the update as expected

Issue #185, opened 2019-03-04
slither-check-upgradability: improve support for complex datatype


montyly:
We need to improve the memory layout checks in case of complex data, including nested structures, enum, ...

0xalpharush:
closed by https://github.com/crytic/slither/pull/1535

PR #1535, opened 2022-12-28
slither-check-upgradeability: support complex datatypes

webthethird:
Addresses [#185](https://github.com/crytic/slither/issues/185), an old one, but still an issue!

Resolves false positives when comparing storage layouts that use structure types (test contracts and updated CI test included).

Currently only addresses structures, but I can add more here to resolve comparisons between other complex types, i.e., contract types and enums. (edit: Contract types also supported, at least on a superficial level, i.e., using the current `Contract.__eq__()` method which only works when comparing a contract object to another contract's name)


webthethird:
@montyly @0xalpharush got any advice re: how to fix these failing checks? It seems like the pip-audit may be unrelated to the content of this PR, since I'm suddenly getting the same issue on #1506 (this check was passing before). As for the Features and Printers tests, it's not immediately clear to me why these are failing, though they seem to be possibly related to recursively nested structures. I just don't understand how the `__eq__()` method I implemented would affect those, when there's no equality comparison being done.

slither/core/declarations/structure.py
```diff
@@ -51,3 +51,12 @@ def elems_ordered(self) -> List["StructureVariable"]:
 
     def __str__(self) -> str:
         return self.name
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Structure):
+            return False
+        for idx, elem in enumerate(self.elems_ordered):
+            other_elem = other.elems_ordered[idx]
+            if other_elem.type != elem.type or other_elem.name != elem.name:
```
0xalpharush, line 60: This can cause an infinite loop because the elem can be a user defined type and its __eq__ calls back into structure when comparing its underlying type
https://github.com/crytic/slither/blob/master/tests/function_ids/rec_struct-0.8.sol

```
slither/core/solidity_types/user_defined_type.py:77: in __eq__
    return self.type == other.type
slither/core/declarations/structure.py:60: in __eq__
    if other_elem.type != elem.type or other_elem.name != elem.name:
```

webthethird: Ahh I see that now. Hmmm, I'll have to think up a way to avoid that. Maybe we just skip over any element in the structure which is an instance of the same structure? But that might still lead to an infinite loop if the element is an instance of some other structure, and that structure includes an element which is an instance of the original struct...

webthethird:
Alright, my last commit resolved the infinite loop issue with recursively nested structures, but I still don't get why the printers test is failing. I tried commenting out my `__eq__()` implementation entirely and running `ci_test_printers.sh` without it, but the test still fails with the following error message:
```
Printer failed
functioncall-0.5.3.sol-0.5.17-compact.zip
```

webthethird:
@0xalpharush @montyly my latest commit resolved the printer test issue (which was related to the `human-summary` printer and one of the detectors failing to hash a Structure object). This time the slither-read-storage test inexplicably failed to install Ganache, and of course pip-audit is still failing, though I saw that there's a separate issue now to fix the latter. Anyway, I'm pretty confident that all issues related to my code additions have been resolved.

slither/core/declarations/structure.py
```diff
@@ -51,3 +51,17 @@ def elems_ordered(self) -> List["StructureVariable"]:
...
         return self.name
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Structure):
+            return False
+        if len(self.elems) != len(other.elems):
+            return False
+        for idx, elem in enumerate(self.elems_ordered):
+            other_elem = other.elems_ordered[idx]
+            if str(other_elem.type) != str(elem.type) or other_elem.name != elem.name:
```
montyly, line 62: I think this might create issue where two types share the same name

For example in 
```solidity
struct St{
	uint a;
}

contract A{

	St s;

}
```

```solidity
import {A} from "test.sol";

struct St{
	uint b;
}

contract B is A{
	St s2;
}
```

Maybe we should create a `canonical_name` for type that would include the source mapping definition?  (cc @smonicas  @0xalpharush)

webthethird: Seems like `canonical_name` would be a good idea. But in the example you gave, wouldn't it be correct to say that the two `St` structs are not equal? In the highlighted code, while enumerating the elements of the two structs, the `other_elem.name != elem.name` would cause my `__eq__()` implementation to return `False`, which seems correct at a glance. Granted, comparing canonical names would probably be better still.

webthethird: I suppose if you renamed `St.b` in the second code block to `uint a`, then the resulting equality might be considered a false positive, since it would see that both structs have the same name, and their elements have the same types and names, therefore returning true. But that seems quite picky, since the two structs would be identical except for where they are defined in Solidity.

Issue #1543, opened 2023-01-03
CI: install solc version only when needed


montyly:
Wen can remove `solc-select install all` for

- https://github.com/crytic/slither/blob/master/.github/workflows/detectors.yml#L38
- https://github.com/crytic/slither/blob/master/.github/workflows/features.yml#L38
- https://github.com/crytic/slither/blob/master/.github/workflows/ci.yml#L58
- https://github.com/crytic/slither/blob/master/.github/workflows/IR.yml#L37

So that the CI only install the solc versions that are needed. Related https://github.com/crytic/solc-select/issues/113

Pavan-Nambi:
hey can i work on this? @montyly

PR #1546, opened 2023-01-03
chore(CI):removing solc-select install all lines

Pavan-Nambi:
issue  https://github.com/crytic/slither/issues/1543#issue-1517415925

### DIFF
--- a/.github/workflows/IR.yml
+++ b/.github/workflows/IR.yml
@@ -34,7 +34,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip install ".[dev]"
-          solc-select install all
           solc-select use 0.8.11
 
       - name: Test with pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip install ".[dev]"
-          solc-select install all
           solc-select use 0.5.1
           pip install typing_extensions==4.1.1
           pip install importlib_metadata==4.8.3
--- a/.github/workflows/detectors.yml
+++ b/.github/workflows/detectors.yml
@@ -35,7 +35,6 @@ jobs:
         run: |
           pip install ".[dev]"
 
-          solc-select install all
           solc-select use 0.7.3
       - name: Test with pytest
         run: |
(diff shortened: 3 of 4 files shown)


elopez (commented):
Thanks for the PR!

.github/workflows/IR.yml
```diff
@@ -34,7 +34,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip install ".[dev]"
-          solc-select install all
           solc-select use 0.8.11
```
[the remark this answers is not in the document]
Pavan-Nambi: ohh got it thankyou

.github/workflows/IR.yml
```diff
@@ -34,8 +34,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install ".[dev]"
-          solc-select install all
-          solc-select use 0.8.11
+          solc-select use 0.8.11--always-install
```
elopez, line 37: There should be a space between the version and the `--always-install` flag:

```suggestion
          solc-select use 0.8.11 --always-install
```

Pavan-Nambi: oh ok thankyou! sorry for the mess

Pavan-Nambi:
i am adding few more installations to pass tests.. instaed of specific ver 
correct me if this approach is wrong...

.github/workflows/ci.yml
```diff
@@ -55,6 +55,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install ".[dev]"
+          solc-select use 0.4.25 --always-install
+          solc-select use 0.8.0 --always-install
```
bohendo, line 59: Looks good, almost there. You'll have to do something similar to this to get the IR tests passing eg update it to look something like this:

```
      - name: Install dependencies
        run: |
          pip install ".[dev]"
          solc-select install 0.5.0
          solc-select use 0.8.11 --always-install

      - name: Install old solc
        if: matrix.os == 'ubuntu-latest'
        run: solc-select install 0.4.0
```
IR tests require 0.5.0 and 0.4.0 except not on windows because that binary is not available.
Note: only one solc version can be `use`d at a time, so you can also just `install` multiple solc versions and only `use` the last one

Pavan-Nambi: ok got it thankyou

montyly:
This is great. Thanks @Pavan-Nambi

PR #1552, opened 2023-01-05
Deprecated safeApprove detector

aviggiano:
# Description

This pull request creates a new detector for deprecated [`safeApprove`](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#SafeERC20-safeApprove-contract-IERC20-address-uint256-) usage.

# Motivation

Although `safeApprove` [has been deprecated](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2219), many projects still use the `SafeERC` extension with that function, instead of using the recommended [safeIncreaseAllowance](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#SafeERC20-safeIncreaseAllowance-contract-IERC20-address-uint256-) and [safeDecreaseAllowance](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#SafeERC20-safeDecreaseAllowance-contract-IERC20-address-uint256-), as we can see on reference issues from Code4rena findings on the OpenZeppelin tread.


0xalpharush:
We're working to identify OpenZeppelin libs by their hash. Perhaps when that is done we can match on the hash, too, so that we don't have false positives

0xalpharush:
I think we should build a tool like https://github.com/crytic/slither/issues/1598 to check for outdated dependencies or deprecated functions instead of implementing a detector for each one

aviggiano:
By the way @0xalpharush, I was also thinking about creating a detector for ChainLink's deprecated APIs, as many projects fall into that trap. See here

https://github.com/code-423n4/2021-06-gro-findings/issues/106
https://github.com/code-423n4/2022-12-tigris-findings/issues/655

Could we use the same method for those too?

aviggiano:
By the way @0xalpharush, I was also thinking about creating a detector for ChainLink's deprecated APIs, as many projects fall into that trap. See here

https://github.com/code-423n4/2021-06-gro-findings/issues/106

Could we use the same method for those too?

PR #1564, opened 2023-01-06
tests: source_unit: remove incomplete submodule

elopez:


tests/test_source_unit.py
```diff
@@ -1,5 +1,7 @@
 from slither import Slither
 
+# NB: read tests/source_unit/README.md before using this test
```
montyly, line 3: Could we detect if this was not done in the test itself, and raise a warning in pytest?

elopez: I added a skipif, it will skip the test if deps are not met + log a skip reason

Screenshot 2023-01-06 at 15 35 18

PR #1567, opened 2023-01-07
fix generate_source_to_evm_ins_mapping

A-23187:
In `generate_source_to_evm_ins_mapping`, the statement `offset, _length, file_id, _ = mapping_item` for unpacking the source mapping items will not work in new solidity versions. Because the generated source mapping items are in the format `s:l:f:j:m` since solidity v0.6.0, the new field `m` has been added. See more here [source-mappings](https://docs.soliditylang.org/en/v0.6.0/miscellaneous.html#source-mappings)


A-23187:
@montyly, hi, are there any other problems for this pull request? I see it's still under review.

0xalpharush:
Thanks @A-23187!

PR #1576, opened 2023-01-10
Add type to type_alias

montyly:
Adding `type` will help with directly access, like in https://github.com/crytic/slither/blob/7ea572723b4eddb18e716c4ae94e0d2100791127/slither/core/cfg/node.py#L911


0xalpharush (approved):

0xalpharush:
It'd be nice to have a test for this

PR #1592, opened 2023-01-11
Detector for unused imports

### LINKED ISSUE #1531: Detect unused import
See https://twitter.com/PaulRBerg/status/1605531233265541121

bart1e:
Detector for unused imports implemented. Fixes #1531.
Files with with cyclic imports and files with `import {...} from` directives are not supported.

Apart from telling the user which imports to remove, it may tell to add more specific imports. For instance if we have:
```
C.sol:
    import "./B.sol"; // but uses only things from A.sol
B.sol:
    import "./A.sol"
```
, then it will tell that `import "./B.sol"` should be removed in `C.sol` and `import "./A.sol"` should be added.

For each file, it searches for all items used there and updates the list of needed imports.
Unfortunately, it isn't currently possible to iterate over user defined types in Slither (aliases like `type something is uint`), so I had to handle them "manually". 
Additionally, some items don't have their `references` set up correctly. Some examples of that include:
- references to contract member functions when they are called
- references aren't added to contract when some other contract inherits from it
- references to top level functions when no contract is declared in a source file

Some issues regarding the implementation:
- since custom types (aliases) had to be treated in a specific way, it's possible that the detector will not detect all of their uses (although I've done my best to handle most common use cases)
- sometimes (for instance in OpenZeppelin repository) there are "import containers" - files that simply contain `import` directives with related files. The detector will tell that they contain unused imports, which is correct, but creating such files is intentional. You may see it by running the detector on @openzeppelin/contracts/interfaces.
- I have created many tests for the detector, especially to test it against false positives - I feel that they are important, but if you feel there are too many of them, just let me know.

You are welcome to test the detector, and if you have any suggestion on how to improve it, or any question regarding the implementation, please let me know.


0xalpharush:
Thanks for tackling this! We definitely want to have this, but will likely need to address the issues you outline. 

> Unfortunately, it isn't currently possible to iterate over user defined types in Slither (aliases like type something is uint), so I had to handle them "manually".

You may be able to iterate over the [compilation unit](https://github.com/crytic/slither/blob/master/slither/core/compilation_unit.py#L50) and contract [scope](https://github.com/crytic/slither/blob/master/slither/core/scope/scope.py#L55) to get this info.

Wrt `references`, some of this has to do with the AST references being inaccurate/ missing (see [here](https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/solc_parsing/expressions/find_variable.py#L420-L439)), but we are working to improve the parsing here https://github.com/crytic/slither/pull/1565. More broadly, we will probably overhaul the current design of top level declarations as it will make features like this easier.

bart1e:
@0xalpharush Thanks a lot for the feedback! I will try to make use of these `_user_defined_value_types` you mentioned - this will definitely make the code shorter and easier to read. I will try to change it tomorrow - I'm a little bit tired a the moment. 

An additional problem that I just realised is within the function `_import_to_absolute_path`, where I try to unify imports by converting them into the absolute paths, but it seems to be failing on some tests. Is there any easy way to convert the `Import` class instance to an absolute path? Because my approach doesn't seem to work.

bart1e:
I have just checked both options (using compilation unit and scope) and indeed, I can iterate over user defined types (aliases), but their references aren't updated correctly. 
So there are two options: 
- I may either edit the code assuming that the references are filled correctly and wait until it's fixed
- or just leave it as is for now

I would prefer the first option, because the code will be easier to read and maintain in the future + we will already have tests for the `references` in user defined types. 
The only question is, how long it will take to improve the parsing (to have correct `references`).

slither/detectors/statements/unused_imports.py
```diff
@@ -0,0 +1,455 @@
...
+                v.type.source_mapping.filename.absolute,
+            )
+
+    def _add_item_by_references(self, item: SourceMapping):
+        """
+        Adds all uses of item to self.needed_imports by its references.
+        """
+        for ref in item.references:
+            self._add_import(
+                self.needed_imports, ref.filename.absolute, item.source_mapping.filename.absolute
```
0xalpharush, line 114: You may need to do something like Path(...).absolute() and then use [filename_loop](https://github.com/crytic/crytic-compile/blob/727cd9189fcc3968688461e0e91083972a718327/crytic_compile/crytic_compile.py#L164) in `_import_to_absolute_path`

0xalpharush:
> but their references aren't updated correctly
@bart1e I'm not exactly sure what's incorrect but you're welcome to make changes that fix it and help make your detector more accurate

bart1e:
@0xalpharush, I will try to make the changes as you suggested. Can I do them in this PR, or create a separate one?

And, perhaps, I have expressed myself wrong previously - by  "but their references aren't updated correctly" I meant that `references` list in user defined types (aliases) doesn't reflect all their uses.
Nonetheless, I see that missing references may be added here:
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/solc_parsing/solidity_types/type_parsing.py#L359-L365

0xalpharush:
@bart1e A separate PR would make it a little easier to review/ test in isolation

bart1e:
@0xalpharush OK, I will try to submit a separate PR tomorrow.

slither/detectors/statements/unused_imports.py
```diff
@@ -0,0 +1,481 @@
...
+        imported_but_unneeded = self._get_imported_but_unneeded()
+
+        for k, v in imported_but_unneeded.items():
+            info = []
+            output = ""
+            if k in UnusedImports.analysed_files or len(v) == 0:
+                continue
+            output += (
+                "Unused imports found in "
+                + PurePath(os.path.relpath(k, os.curdir)).as_posix()
```
0xalpharush, line 457: Not 100% sure, but I think we can get rid of the local/CI file path discrepancies by using `PurePath(k).as_posix()` here

bart1e: I have made the change you suggested, however, for some reason, tests weren't run on the code, so I am not sure if it passes the tests that were failing.

bart1e:
@0xalpharush Before this will be merged, I have to know how the detector should deal with the following things:
- as I have mentioned before, sometimes there are files that just aggregate imports (like https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC3156.sol). The question is, how the detector should handle these cases. Right now, for the file I have linked, it will output that 2 files imported there are unneeded. Optionally, I may try to detect (and ignore) such "import containers" by checking whether a file contains only import directives.
- the second question is about which unused imports should be listed by the detector. Consider the following scenario:
```solidity
// file C.sol uses B.sol and A.sol and contains the following imports:
import "./A.sol";
import "./B.sol";
// file B.sol doesn't use anything from A.sol, but contains the import:
import "./A.sol";
```
Currently, the detector will say that `import "./A.sol"` is unneeded in `C.sol` (since `B.sol` that is already imported contains it), and that `import "./A.sol"` is unneeded in `B.sol`. So, if the user removes `A.sol` import from both files at the same time, the project will not compile. Should we leave that behaviour or should the detector's output print that `import "./A.sol"` is unneeded only in `B.sol`?

0xalpharush:
I think it'd make sense to ignore files that only contain imports but if a file imports this "container" we'd still want to warn for unused ones.

I think leaving it as is for right now is fine. If it becomes a nuisance for users down the line, we can improve the heuristic.

bart1e:
I have added a code that will ignore "import containers" while printing the output. 

In the future, I may also handle cyclic imports (right now, if cyclic imports are detected, the detector simply omits the files that contain them) - I think that would be too much for one PR as the code is already complicated enough.

slither/detectors/statements/unused_imports.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+class UnusedImports(AbstractDetector):
+    """
+    Unused imports detector
+    """
+
+    ARGUMENT = "unused-imports"
+    HELP = "Detect unused imports"
+    IMPACT = DetectorClassification.OPTIMIZATION
```
0xalpharush, line 23:
```suggestion
    IMPACT = DetectorClassification.INFORMATIONAL
```

slither/detectors/statements/unused_imports.py
```diff
@@ -0,0 +1,526 @@
...
+            info = []
+            output = ""
+            if k in UnusedImports.analysed_files or len(v) == 0:
+                continue
+            if k in UnusedImports.import_containers:
+                continue
+            output += "Unused imports found in " + PurePath(k).as_posix() + ".\n"
+            output += "Consider removing the following imports:\n"
+            for imp in v:
+                output += PurePath(self.absolute_path_to_imp_filename[imp]).as_posix()
```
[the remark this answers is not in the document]
bart1e: Good catch - I fixed it in the newest commit.

slither/detectors/statements/unused_imports.py
bart1e: I changed it in the latest commit.

slither/detectors/statements/unused_imports.py
```diff
@@ -0,0 +1,543 @@
...
+    WIKI = "https://github.com/crytic/slither/wiki/Detector-Documentation#unused-imports"
+
+    WIKI_TITLE = "Unused imports"
+    WIKI_DESCRIPTION = (
+        "Detects imports that are not used by source files. Currently, files with cyclic imports and files with "
+        "'import {...} from' directives are not supported."
+    )
+    WIKI_RECOMMENDATION = "Remove unused imports"
+
+    analysed_files: Set[str] = set()
```
montyly, line 39: Using class variables for things that are going to be dynamically change by the objects seem really error prone. Can you explain the reasoning here?

montyly (commented):
This is an amazing work, thanks @bart1e for the effort. 

I made minor modifications (mostly python related). My main concern is around the class variable (see the comment)

slither/detectors/statements/unused_imports.py
bart1e: Sure, my reasoning about this was to store some data globally so that it persists through the detector's execution. 

If my understanding is correct (please correct me if I'm wrong), detector will be run on each file individually. However, the detector analyses all the files in the compilation unit and prints recommendations for all of them. So, the `analysed_files` set tells the detector not to print findings for files that were already analysed if they belong to many compilation units.

The reasoning behind the second class level variable, `import_containers`, was to store files that only aggregate imports, so that the detector doesn't have to check if they are "import containers" all over again. However, I forgot to add the check in `_is_import_container`, so it doesn't provide any optimization at the moment and when I think about it right now, such files don't contain too much code, so even if they are analysed once again, it wouldn't bring too much overhead - I should probably change it to instance variable.

bart1e: It makes perfect sense. Thank you for the explanation. I have changed `import_containers` to an instance variable and removed `analysed_files` entirely. 

I was running `slither . --detect unused-imports` in the test directory and it was reporting each file individually and hence my confusion. But when I tried it on a hardhat project, it doesn't duplicate outputs.

0xalpharush:
We determined that we don't want to identify "redundant" imports i.e. where a dependency can be satisfied transitively and instead only recommend removing absolutely unused imports

bart1e:
I've removed the code responsible for detecting "redundant" imports. 

Now, only imports whose own import graph doesn't contain any needed import will be listed.

So, for example, if we have `A.sol` importing `X_1, X_2, ..., X_10` and `X_i` imports `B.sol` and `A.sol` uses something from `B.sol`, but nothing from `X_i`, then nothing will be printed by the detector.

So, the detector will now print something a lot less frequently than with the previous implementation.

mds1:
Just wanted to bump this and see if there are plans to get this merged anytime soon :)

0xalpharush:
I will try to get this updated and ready for the next release (probably a couple weeks)

PR #1599, opened 2023-01-13
Clarify requirement of installing solc

CodeSandwich:
Maybe other toolkits don't require `solc` either, I didn't have a chance to find out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,8 +175,9 @@ See the [Tool documentation](https://github.com/crytic/slither/wiki/Tool-Documen
 
 ## How to install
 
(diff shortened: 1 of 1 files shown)


montyly:
Thanks CodeSandwich for the contribution. Actually this statement is true for *any* compilation platform, not just foundry (including hardhat, truffle, brownie, ...).

Maybe we could say something like `[solc](https://github.com/ethereum/solidity/), the Solidity compiler, or any [supported compilation framework](https://github.com/crytic/crytic-compile))` ?

CodeSandwich:
I didn't know that, that's pretty cool! Added.

PR #1601, opened 2023-01-13
Minor API improvements

montyly:
- Add `Source.content` - the same code was re-use in different parts
- Add `Source.content_hash` to standardize tests on source code
- Add `unchecked_arithemtic_usage`, which returns function with some unchecked arithmetic operations


0xalpharush (approved):

0xalpharush:
Long term, this does seem like something that belongs in crytic-compile. For instance, it would allow us to have a caching layer and clean compilation artifacts if changes in the source files are detected.

PR #1604, opened 2023-01-14
Missing references fix

bart1e:
There is a problem with user defined aliases (using `type [...] is [...]` statement) - their `references` list is not updated during parsing.
Also, user defined types (contracts, interfaces, etc.) have missing references when casting (`TypeConversion` expressions).
This PR fixes these issues and is needed to simplify code in #1592.

### DIFF
--- a/slither/solc_parsing/expressions/expression_parsing.py
+++ b/slither/solc_parsing/expressions/expression_parsing.py
@@ -2,7 +2,6 @@ import logging
 import re
 from typing import Dict, TYPE_CHECKING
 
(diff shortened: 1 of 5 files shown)


0xalpharush:
@bart1e If possible, would you mind adding a test here please? https://github.com/crytic/slither/blob/master/tests/test_source_mapping.py

bart1e:
@0xalpharush Of course, I will try to add them tomorrow.

montyly:
That's great, thanks @bart1e

Issue #1593, opened 2023-01-11
only warn once per compilation unit for `solc-version`  detector


0xalpharush:
### Describe the desired feature

Currently a result is created for every file but we could only warn once for each version that is considered too recent

EnbangWu:
Can I take on this issue? I think it's `incorrect_solc.py` detector right?

0xalpharush:
@EnbangWu That would be great! Yes, it's `incorrect_solc.py`. I think a good way to fix this would be to make `disallowed_pragmas` a set instead of list so that only unique results are kept.
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/detectors/attributes/incorrect_solc.py#L128-L141
Then, you'll may need to update the detector's tests using `python ./tests/test_detectors.py --overwrite` as explained [here](https://github.com/crytic/slither/blob/master/CONTRIBUTING.md#detectors-tests). Lmk if you need help or have questions!

EnbangWu:
That's a great help! Thanks a lot, @0xalpharush !

EnbangWu:
OK just wrote two test solidity files using version 0.8.17 ( as how the too_recent_versions are defined in incorrect.sol) Running `slither .` would give us 
`Pragma version^0.8.17 (test1.sol#2) necessitates a version too recent to be trusted. Consider deploying with 0.6.12/0.7.6/0.8.16 solc-0.8.17 is not recommended for deployment Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity`
`Pragma version^0.8.17 (test2.sol#2) necessitates a version too recent to be trusted. Consider deploying with 0.6.12/0.7.6/0.8.16 solc-0.8.17 is not recommended for deployment Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity`

would like to see something like this:
`Pragma version^0.8.17 (test1.sol#2), Pragma version^0.8.17 (test2.sol#2) necessitates a version too recent to be trusted. Consider deploying with 0.6.12/0.7.6/0.8.16 solc-0.8.17 is not recommended for deployment Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity`

EnbangWu:
[...]

PR #1605, opened 2023-01-15
only warn once per compilation unit for incorrect_solc.py detector

EnbangWu:
In order to solve this [issue](https://github.com/crytic/slither/issues/1593), disallowed_pragmas is replaced by a set and can only contain unique `(reason,p)` tuples. So that we could only warn once for each version that is considered too recent.

`pytest ./tests/test_detectors.py` ran with no failed tests.

### DIFF
--- a/slither/detectors/attributes/incorrect_solc.py
+++ b/slither/detectors/attributes/incorrect_solc.py
@@ -125,7 +125,7 @@ Consider using the latest version of Solidity for testing."""
         # Detect all version related pragmas and check if they are disallowed.
         results = []
         pragma = self.compilation_unit.pragma_directives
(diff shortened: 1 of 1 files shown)


0xalpharush:
I don't think this works since the reason is what needs to be unique for every result. It may make sense to use `disallowed_reasons: Dict[str, List[Pragma]]`
and if the reason is in the dict append the pragma.  Then, the detector will just list all of the pragmas which share the same reason.

EnbangWu:
got it, will take a further look and test it then!

0xalpharush:
Closing this for now. Feel free to reopen once you've made the requested changes!

PR #1617, opened 2023-01-21
Bugs fixed in strongly connected components and cyclomatic complexity algorithms

bart1e:
File `code_complexity.py` had two bugs in it. This PR fixes both of them.

The first bug was related to strongly connected components algorithm: the final DFS (or `assign` in this case) should be performed on reversed postorder order instead on just postorder.

The second issue concerned calculating cyclomatic complexity: `P` in the formula should be a number of connected components (always `1` in case of solidity functions) instead of strongly connected components.

### DIFF
--- a/slither/utils/code_complexity.py
+++ b/slither/utils/code_complexity.py
@@ -52,7 +52,7 @@ def compute_strongly_connected_components(function: "Function") -> List[List["No
             for father in node.fathers:
                 assign(father, root)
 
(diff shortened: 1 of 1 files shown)


0xalpharush:
I think may be useful to create helper methods on functions to get nodes in postorder and reverse postorder

montyly:
great catch @https://github.com/bart1e ;)

Issue #1591, opened 2023-01-11
Add informational detector for high cyclomatic complexity


montyly:
We should probably flag anything above 11 as risky for smart contracts (see https://en.wikipedia.org/wiki/Cyclomatic_complexity#Interpretation)

We already have a function that computes it: https://github.com/crytic/slither/blob/master/slither/utils/code_complexity.py#L64.

0xalpharush:
closed by https://github.com/crytic/slither/pull/1618

PR #1618, opened 2023-01-21
High cyclomatic complexity detector added

bart1e:
Detector for high cyclomatic complexity (`> 11`) added, as requested in #1591.
Needs #1617 to be merged before due to a bug in the calculation of cyclomatic complexity.


slither/detectors/functions/cyclomatic_complexity.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    def _detect(self):
+        results = []
+
+        for f in self.compilation_unit.functions:
+            cc = compute_cyclomatic_complexity(f)
+            if cc > 11:
+                info = (
+                    "Function "
```
montyly, line 31: We should use `[ f, f" has a high cyclomatic complexity ({cc})"]` here.

`generate_results` has inbuilt that will transform slither object (i.e. `f` here) in a human readable format (ex: adding the source mapping directly)

slither/detectors/functions/cyclomatic_complexity.py
```diff
@@ -0,0 +1,41 @@
...
+    WIKI_DESCRIPTION = "Detects functions with high (> 11) cyclomatic complexity."
+    WIKI_EXPLOIT_SCENARIO = ""
+    WIKI_RECOMMENDATION = (
+        "Reduce cyclomatic complexity by splitting the function into several smaller subroutines."
+    )
+
+    def _detect(self):
+        results = []
+
+        for f in self.compilation_unit.functions:
```
montyly, line 27: I think we should iterate over `contracts` and then `contract.functions_declared` here. The reason is that the same function at the solidity source code level might appear multiple times in `compilation_unit.functions`.

This is because we create one function object per contract's instance to ensure proper IR conversion (ex: if a function uses `super`, the function that will be actually called can depend on the inheritance order).

I will put a note to improve our documentation there

montyly:
This is nice, thanks @bart1e

slither/detectors/functions/cyclomatic_complexity.py
bart1e: Thank you for the feedback. I will change it in a while. But, iterating only over contract functions will miss top level functions. Should I also iterate over `self.compilation_unit.functions_top_level`? I don't think they are widely used, but solidity compiler allows using top level functions, so it may be a good idea to take them into account as well.

bart1e:
@montyly I have added the changes you requested. I am also iterating over top level functions - if we don't need it, just let me know and I will remove it.

montyly:
I merged the PR with #1660. thanks again for the contribution @bart1e. If you are curious, I made minor changes: https://github.com/crytic/slither/pull/1660/commits/d0c0a178231e69315ee10f55859be68ade7c2241

Issue #1607, opened 2023-01-17
type propagation for library calls on type aliases isn't correct


0xalpharush:
Thanks for implementing this feature, guys!

Unfortunately, we might have to re-open this issue. Slither doesn't work with chained function calls on value types:

```solidity
pragma solidity >=0.8.17;

import { UD60x18 } from "prb-math/UD60x18.sol";

contract Foo {
    function tryUnwrap(UD60x18 a, UD60x18 b) external pure returns (uint256 c, uint256 d) {
        // This is ok
        c = a.unwrap();
        // This triggers a bug
        d = a.mul(b).unwrap();
    }
}
```

Defining `c` works, but `d` doesn't. That is, I am getting the following error:

```text
Error:
Function not found on IR: TMP_3575(None) = HIGH_LEVEL_CALL, dest:TMP_3574(None), function:unwrap, arguments:[]  .
Node: EXPRESSION d = a.mul(b).unwrap() (src/Foo.sol#9)
Function: tryUnwrap
Please try compiling with a recent Solidity version. 'NoneType' object has no attribute 'type'
```

I can attach the full stack trace but you should be able to replicate this by spinning up a Foundry project and installing [PRBMath](https://github.com/paulrberg/prb-math) (the latest commit from `main`).

_Originally posted by @PaulRBerg in https://github.com/crytic/slither/issues/1352#issuecomment-1384226469_

PaulRBerg:
Sorry to follow up on this - I wonder if there is an ETA for the fix? Or is there any way I could help here (besides making a PR, as I am not familiar with the source code of Slither)?

Running a Slither check over our contracts is one of the last items on the todo list - we're prepping up for an audit.

0xalpharush:
I'm having some trouble replicating this. Would you create a repo where I can reproduce this issue please? I'm getting this error: 
```
Compiler run failed
error[9582]: TypeError: Member "unwrap" not found or not visible after argument-dependent lookup in UD60x18.
  --> src/Test.sol:10:13:
   |
10 |         d = a.mul(b).unwrap();
   |             ^^^^^^^^^^^^^^^

```

PaulRBerg:
Forge installs the latest release tag (`v3.2.0` in the case of PRBMath) - you just have to edit the `branch` config option in your `.gitmodules` to point to `main`, and then update your git submodules.

I will go open an issue in the Foundry repo about this because tags cannot be used in the `branch` option.

0xalpharush:
[...]

PR #1625, opened 2023-01-24
Fix using for global function name collision

0xalpharush:
Both `SD59x18` and `UD60x18` defined a `mul` function and globally bind the function to the type which wasn't being handled by `_analyze_top_level_function`. I don't think overloading can be done in a using for but that should be investigated.

Fix https://github.com/crytic/slither/issues/1607

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,46 @@
+name: docs
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
+jobs:
+  # Single deploy job since we're just deploying
+  build:
+    environment:
+      name: github-pages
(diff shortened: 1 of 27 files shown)


PaulRBerg:
Nope, Solidity doesn't support overloading functions in the `using for ... global` directive. I have actually [requested this as a feature](https://forum.soliditylang.org/t/permit-overloaded-methods-for-user-defined-value-types/987/3) on the Solidity forum.

Taking this code snippet as an example:

```solidity
pragma solidity >=0.8.17;

type UD60x18 is uint256;

using { foo } for UD60x18;

function foo(UD60x18 x) pure returns (UD60x18 result) {
    result = x;
}

function foo(UD60x18 x, UD60x18 y) pure returns (UD60x18 result) {
    x;
    result = y;
}
```

Will produce this error:

```text
from solidity:
DeclarationError: Identifier not found or not unique.
 --> contracts/Foo.sol:6:9:
  |
6 | using { foo } for UD60x18;
```

montyly:
Can we add a test for this one?

montyly:
Can we add the test to the CI? Do we actually need foundry to run this example?

PaulRBerg:
I'm sorry to chase, but are there any updates here? Is this fix still slated to be included in v0.9.3?

0xalpharush:
@montyly I added a test. @PaulRBerg This will be included in the upcoming release

PaulRBerg:
Amazing news thank so much ser @0xalpharush!! Given that our audits are slated to start on March 20, I'm glad that we will be able to run Slither until then.

PR #1637, opened 2023-01-31
run tests in parallel

0xalpharush:


elopez:
looks like some of the tests are not "thread safe" 😅

.github/workflows/detectors.yml
```diff
@@ -38,4 +38,4 @@ jobs:
           solc-select use 0.7.3 --always-install
       - name: Test with pytest
         run: |
-          pytest tests/test_detectors.py
+          pytest tests/test_detectors.py -n auto
```
0xalpharush, line 41:
```suggestion
          pytest tests/test_detectors.py 
```

montyly:
Note: the detectors are not thread safe, because they run solc-select. However in the long term we could compile the detectors tests files, so that we only run on the crytic-compile output.
It will also save the compilation time

Issue #1595, opened 2023-01-12
[Bug]: immutable-state detector incorrectly claims strings can be made immutable


area:
### Describe the issue:

According to the [solidity docs](https://docs.soliditylang.org/en/latest/contracts.html#constant-and-immutable-state-variables) while strings can be declared `constant`, they can not be declared `immutable` at this time. This detector, however, asks for strings to be declared `immutable` as if they were supported. Doing so causes the solidity compilation to fail.

### Code example to reproduce the issue:

```
pragma solidity 0.8.17;

contract X {
  bool public immutable boolVar;
  string public stringVar;


  constructor(bool _boolVar, string memory _stringVar) {
    boolVar = _boolVar;
    stringVar = _stringVar;
  }
}
```

### Version:

0.9.2

0xalpharush:
This wasn't accounted for when reusing the `_valid_type` helper function from the constant detector. Just need to filter out string types here https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/detectors/variables/unchanged_state_variables.py#L120-L123

PR #1639, opened 2023-02-01
do not recommend to make string immutable 

0xalpharush:
close https://github.com/crytic/slither/issues/1595

### DIFF
--- a/slither/detectors/variables/unchanged_state_variables.py
+++ b/slither/detectors/variables/unchanged_state_variables.py
@@ -118,7 +118,7 @@ class UnchangedStateVariables:
                         self.constant_candidates.append(v)
 
                     elif (
-                        v.type != ElementaryType("string")
+                        not v.is_dynamic
                         and version.parse(self.compilation_unit.solc_version)
                         >= version.parse("0.6.5")
                         and (v in constructor_variables_written or v in variables_initialized)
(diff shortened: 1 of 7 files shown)


slither/detectors/variables/unchanged_state_variables.py
```diff
@@ -118,8 +118,9 @@ def detect(self):
                         self.constant_candidates.append(v)
 
                     elif (
-                        v in constructor_variables_written or v in variables_initialized
-                    ) and version.parse(self.compilation_unit.solc_version) >= version.parse(
-                        "0.6.5"
+                        v.type != ElementaryType("string")
```
0xalpharush, line 121: We need to consider bytes, too

0xalpharush, line 121:
```suggestion
                        not v.is_dynamic
```

slither/detectors/variables/unchanged_state_variables.py
```diff
@@ -118,8 +118,9 @@ def detect(self):
                         self.constant_candidates.append(v)
 
                     elif (
-                        v in constructor_variables_written or v in variables_initialized
-                    ) and version.parse(self.compilation_unit.solc_version) >= version.parse(
-                        "0.6.5"
+                        not v.is_dynamic
```
0xalpharush, line 121:
```suggestion
                        not v.type.is_dynamic
```

Issue #435, opened 2020-04-09
Incorrect handling of continue statement in Slither CFG


aaronyoo:
Consider the following function snippet (a contrived example, but it continue is a relatively common pattern I think):
```solidity
function f(uint input) public {
    for (uint i = 0; i < input; i++) {
        if (i == 0) {
            continue;
        }
        a = a + 1;
    }
}
```

Slither produces the following CFG:
![image](https://user-images.githubusercontent.com/11811606/78927043-069c4400-7a53-11ea-9bdf-1c2ed8a005e3.png)
I am pretty sure this is wrong because of the edge from `CONTINUE --> BEGIN_LOOP`. I think it should be `CONTINUE --> EXPRESSION i++`.  Otherwise, this graph implies that taking continue would lead to an infinite loop, which it does not. I think the problem is that we are jumping to the beginning of the loop and not the increment.

Furthermore I have found that the following different program gives the exact same CFG:
```solidity
 function f(uint input) public {
    for (uint i = 0; i < input;) {
        if (i == 0) {
            continue;
        }
        a = a + 1;
        i++;
    }
}
```

In this case, Slither is correct because taking the continue would imply an infinite loop (because the increment would be skipped).

---

I have looked around at Slither's internal representation but there appears to be nothing that indicates whether the statement is part of the loop header or not. This means that the problem is that Slither cannot distinguish between the above two programs because it does not know whether the i++ is part of the loop header or not. I looked at the solc-ast for this program briefly but I couldn't find anything that would solve this.

duckki:
As I read the code, in most cases, a loop jumps back to the immediate son of the STARTLOOP node. It should be a sole son, which is either a loop condition or do-while body.
The exception is for-loops without conditions, in which case the loop jumps back to the STARTLOOP node itself.

I think there are two possible solutions:
1) When there is no condition in a for-loop, add an no-op EXPRESSION node as a son of STARTLOOP node before the for-loop body.
2) Record the for-loop-no-condition flag in the STARTLOOP node somehow.

Either way, `_fix_continue_node` should be able to patch the back-edge correctly.

PR #629, opened 2020-09-09
cleanup break/continue fixup

samczsun:
we know which loop we're in when we're constructing the cfg, so let's link break/continue properly during cfg construction instead of doing the hacky "find which loop we're in after the fact" thing

my very basic testcase:
```solidity
contract C {
    function f() public {
        for (uint i = 0; i < 10; i++) {
            if (i > 100) {
                break;
            }
            if (i < 3) {
                continue;
            }

            for (uint j = 0; j < 10; j++) {
                if (j > 10) {
                    continue;
                }
                if (j < 3) {
                    break;
                }

                j -= 1;
            }
        }
    }
}
```

PR #1651, opened 2023-02-09
AST parser fix: find the correct jump destinations for `continue` statements

duckki:
This PR fixes https://github.com/crytic/slither/issues/435.

### Issue
`_fix_continue_node` doesn't always find the correct jump destination for `continue` statement.

### Proposed Solution
In this PR, I propose the record the continue destination node in each `STARTLOOP` node.

### Testing
The existing `continue-all.sol` AST test shows the desired changes.
Also, I added while..continue and do..continue..while tests. Since do-while was added in Solidity 0.4.5 and its semantics changed in 0.5. I split the `continue-all.sol` AST test into three versions.


slither/core/cfg/node.py
```diff
@@ -230,6 +230,10 @@ def __init__(
         self.scope: Union["Scope", "Function"] = scope
         self.file_scope: "FileScope" = file_scope
 
+        # Record the continue destination in the STARTLOOP node, so `continue` node can
+        # find the correct jump destination.
+        self.continue_destination: Node = None
```
0xalpharush, line 235: Would it be able to pass down the break and continue context in the parsing rather than store the data on the node?

duckki: Yes, once https://github.com/crytic/slither/pull/629 is merged, I can use `_loops`.
Maybe I can create a new PR on top of `feature/cleanup-fixup` branch.

Issue #1652, opened 2023-02-09
Create contract level `UsingFor` so that source mapping can be tracked


0xalpharush:
We need to create a contract level `UsingFor` class so that we have access to the source mapping for the directive. 
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/core/declarations/using_for_top_level.py#L16-L18
This would require creating the object here
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/solc_parsing/declarations/contract.py#L586
This requires updating the top level object and parsing as well
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/core/declarations/using_for_top_level.py#L10

PR #1653, opened 2023-02-09
Detector for invalid using for at the top level

bart1e:
This PR implements the detector for the incorrect usage of the `using-for` statement: in `Solidity` it is possible to write `using <library> for <type>` even when `<library>` does not contain any function that has `<type>` as its first argument. Such a code will compile even though the statement has no effect.

This detector detects such incorrect usages of `using-for` statements, currently only for top level `using-for` statements (#1652 has to be solved in order to implement detector for contract level `using-for` statements).

It contains several utility functions checking if a certain type may be implicitly convertible to some other one - we can consider moving them to relevant classes (for instance these related to `ElementaryType` may be added there).

There is one issue though - I wasn't able to get information about the `address` type: if it's `payable` or not ("normal" `address` is not implicitly convertible to `address payable`, but I am unable to impose this rule right now).  Detector now always says that `address` types are implicitly convertible to each others - it won't bring any false positives, but we may get some false negatives.


montyly:
This is great work, thanks @bart1e

PR #1655, opened 2023-02-11
add github pages docs

0xalpharush:
https://pdoc.dev/docs/pdoc.html#deploying-to-github-pages


elopez (changes requested):
The Actions part looks good to me, but can the `docs/` folder with autogenerated files be removed from the PR? Adding it to gitignore would be nice as well, so we don't end up committing them by mistake going forward.

elopez (approved):

PR #1665, opened 2023-02-15
Add enums to echidna printer's list of constants

samalws:
Fixes https://github.com/crytic/echidna/issues/918.

This PR makes the echidna printer count enum values as uint256 constants, including them in the `"constants_used"` list so that echidna is aware of them. Currently, when a function has enum arguments, echidna will try all numbers, not privileging in-range enum values over out-of-range values.

Example:
```solidity
contract Example {
  bool flag = false;
  function setFlag(ExampleEnum a, ExampleEnum b, ExampleEnum c) public {
    if (a == ExampleEnum.E28 && b == ExampleEnum.E29 && c == ExampleEnum.E30)
      flag = true;
  }
  function echidna_flag() public view returns (bool) {
    return !flag;
  }
  enum ExampleEnum {
    E0,
    E1,
    // ...
    E30
  }
}
```

Before this PR, echidna takes a very long time to break the contract (didn't break it in 500000 iterations when I ran it). Afterwards, it breaks the contract nearly instantly.

I chose to make enum values be considered uint256 constants rather than uint8 constants since some versions of solidity allow for enums with more than 256 values.

### DIFF
--- a/slither/printers/guidance/echidna.py
+++ b/slither/printers/guidance/echidna.py
@@ -4,7 +4,7 @@ from typing import Dict, List, Set, Tuple, NamedTuple, Union
 
 from slither.analyses.data_dependency.data_dependency import is_dependent
 from slither.core.cfg.node import Node
-from slither.core.declarations import Function
+from slither.core.declarations import Enum, Function
 from slither.core.declarations.solidity_variables import (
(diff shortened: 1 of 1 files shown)


montyly:
great idea, thanks @ samalws

PR #1666, opened 2023-02-16
Add more types hints

montyly:
+ adding TODO on areas that require further investigation

A general issue is related to the usage of the `type`, as it sometimes return a list of types, which might need a bit of refactoring. Moreover `ExpressionTyped` has derived class that redefines `type` 

Another source of issues that will require some refactoring is the handling of top level function in the data dependencies. The analysis was not built with top level function in mind, and needs to be improved


montyly:
Note: this PR includes #1672, #1673, #1674

montyly:
This PR contains several minor bug fixes in the API. They will be documented before the PR is merged

montyly:
In addition to the types, this fixes:
- Remove `process_from_asts`/`--splitted`, which became deadcode 
- Add `SUPPORTED_TYPES` / `Context_types_API` for the data dependencies. This allows now to give a node to data dependencies, to better support top level function. This is a quick fix, and requires a larger refactoring of the data deps. A separate issue must be created
- Remove most of the `core.children`. These ended up making the API more difficult to maintain without providing enough benefit. The relevant property were added when needed in the derived objects
- Create `ContractLevel` to works with `TopLevel`. This replaces `ChildContract`
- Remove `ExpressionTyped`. This is legacy code, the `type` property ended up colliding with properties in derived objects. The relevant `type` properties were added when needed
- Detector can now use `DETECTOR_INFO = List[Union[str, SupportedOutput]]`, to type the `info` object. All the detectors were updated to use this. The reason for that is that mypy is a bit lost when deduced the type, as the list is a union of str + a lot of different objects.
- Remove `visitors.expression.right_value` this is legacy code

slither/analyses/data_dependency/data_dependency.py
```diff
@@ -26,39 +30,51 @@
...
 ###################################################################################
 # region User APIs
 ###################################################################################
 ###################################################################################
 
 
-Variable_types = Union[Variable, SolidityVariable]
+SUPPORTED_TYPES = Union[Variable, SolidityVariable]
+
+# TODO refactor the data deps to be better suited for top level function object
```
0xalpharush, line 47: Can we open an issue outlining what is currently wrong and the desired behavior?

slither/analyses/write/are_variables_written.py
```diff
@@ -70,17 +72,20 @@ def _visit(
         if ir.lvalue and not isinstance(ir.lvalue, (TemporaryVariable, ReferenceVariable)):
             variables_written.add(ir.lvalue)
 
-        lvalue = ir.lvalue
+        lvalue: Any = ir.lvalue
```
0xalpharush, line 75: Not blocking but we should type ir.lvalue so this can be `Optional[...]`

slither/analyses/data_dependency/data_dependency.py
montyly: Yes, see the PR message: `This is a quick fix, and requires a larger refactoring of the data deps. A separate issue must be created`

PR #1669, opened 2023-02-18
add trunk linter

0xalpharush:


woodruffw:
The last two commits run `trunk check -a` and `trunk fmt -a` -- they produce a massive changeset, which makes me think that we need to disable some of the lints here (e.g., we probably shouldn't be auto-formatting JSON test artifacts?)

CC @0xalpharush for opinions.

woodruffw:
Yep, plus they broke the tests. Maybe `ruff` support in Trunk is still pretty green.

0xalpharush:
@woodruffw I don't think we need to enable prettier or we should ignore the JSON test artifacts. Wrt `ruff`, it looks like a circular import was introduced during the change and might be unrelated to ruff itself.

PR #1679, opened 2023-02-22
Abstract contract property

bsamuels453:
This PR adds an "is_abstract" property to the contract class. Without this class property, detectors can only check for abstractness by verifying all of the functions have implementations.

### DIFF
--- a/slither/core/declarations/contract.py
+++ b/slither/core/declarations/contract.py
@@ -85,7 +85,7 @@ class Contract(SourceMapping):  # pylint: disable=too-many-public-methods
         self._kind: Optional[str] = None
         self._is_interface: bool = False
         self._is_library: bool = False
-        self._is_abstract: bool = False
+        self._is_fully_implemented: bool = False
 
         self._signatures: Optional[List[str]] = None
         self._signatures_declared: Optional[List[str]] = None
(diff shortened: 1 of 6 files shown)


slither/core/declarations/contract.py
```diff
@@ -165,6 +166,14 @@ def is_library(self) -> bool:
     def is_library(self, is_library: bool):
         self._is_library = is_library
 
+    @property
+    def is_abstract(self) -> bool:
```
montyly, line 170: I think we should either:
- Document that this supports only contracts where `abstract` is available
- Or implement an abstract contract check, to detect abstract contract before `abstract` was available 

Otherwise users might think that this will return true for any abstract contract, while it will only work for some

montyly (changes requested):

slither/core/declarations/contract.py
bsamuels453: Would it be a better idea to rename this property to "is_fully_implemented"?

The detector that will consume this property does not necessarily care about the `abstract` keyword; it just needs to be run on the "fully implemented" contract.

montyly: yes I like this name. It matches `function.is_implemented`

Not that for the check, we need to filter out the functions that are shadowed  (`function. is_shadowed`) as the contract object has access to all the functions

bsamuels453: for the check, I am just using the `is_fully_implemented` attribute from the solidity AST, so I don't think we have to check for shadowed functions: https://github.com/ethereum/solidity/blob/7dd6d404815651b2341ecae220709a88aaed4038/libsolidity/ast/ASTJsonExporter.cpp#L295-L296

0xalpharush:
It'd be nice to have test cases to known when this will return `True` both for the compact and legacy AST.

montyly:
Note: this is currently failing because of https://github.com/crytic/slither/pull/1741

PR #1694, opened 2023-02-26
Detector for array length caching added

bart1e:
This detector checks for loops that compare loop index against `array.length` and doesn't modify it (where `array` is a state array). Caching `array.length` in a local variable will save gas in such cases.

It is done by parsing each loop individually using function nodes. For each loop, it is checked whether it modifies `array.length` in some way (it can be done using `push`, `pop` or `delete` on the entire array).

For simplicity, the check for `push` and `pop` is done in one step, by iterating over `irs` of each node in the loop. I only check whether `array.length` is referenced there - the reason for that is explained in the code.


slither/detectors/operations/cache_array_length.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _detect(self):
+        results = []
+
+        non_optimal_array_len_usages = CacheArrayLength._get_non_optimal_array_len_usages(
+            self.compilation_unit.functions
+        )
+        for usage in non_optimal_array_len_usages:
+            info = [
+                f"Loop condition at {usage.source_mapping} should use cached array length instead of referencing "
```
0xalpharush, line 210:
```suggestion
                "Loop condition at", usage, "should use cached array length instead of referencing "
```
This will display more info context

bart1e: When I make that change, I'm getting `slither.exceptions.SlitherError: <class 'slither.core.expressions.binary_operation.BinaryOperation'> cannot be converted (no name, or canonical_name`. It seems that `BinaryOperation` isn't yet supported in `_convert_to_description`.

slither/detectors/operations/cache_array_length.py
```diff
@@ -0,0 +1,215 @@
...
+                        exp.expression_left
+                    )
+                if array is None:
+                    continue
+
+                visited: Set[Node] = set()
+                if not CacheArrayLength._is_loop_referencing_array_length(
+                    if_node, visited, array, 1
+                ):
+                    non_optimal_array_len_usages.append(if_node.expression)
```
0xalpharush, line 176: Instead of tracking the expression, if we track the node (`Node`), it will render properly

slither/detectors/operations/cache_array_length.py
0xalpharush: https://github.com/crytic/slither/pull/1694#discussion_r1194131267

slither/detectors/operations/cache_array_length.py
```diff
@@ -0,0 +1,215 @@
...
+            depth += 1
+        if node.type == NodeType.ENDLOOP:
+            depth -= 1
+        if depth == 0:
+            return False
+
+        # Array length may change in the following situations:
+        # - when `push` is called
+        # - when `pop` is called
+        # - when `delete` is called on the entire array
```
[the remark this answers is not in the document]
bart1e: I may handle that. Should we also consider internal calls in such a scenario?

slither/detectors/operations/cache_array_length.py
bart1e: Will change it in the next PR. 

However, the small issue here is that Slither will output the `Node` using internal variable names. So, for example, it may output something like: `Loop condition at i_scope_4 < array.length [...]`, for the variable `i`, which may be a bit misleading for users.

slither/detectors/operations/cache_array_length.py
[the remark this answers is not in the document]
0xalpharush: Good point. If you use `node.all_slithir_operations` instead of `node.irs`, it will include the operations of internal calls (https://crytic.github.io/slither/slither/core/cfg/node.html#Node.all_slithir_operations)

slither/detectors/operations/cache_array_length.py
0xalpharush: I think that's acceptable although not ideal. Other detectors do this as well

bart1e:
@0xalpharush I have made the changes you requested. 

In case of external function calls, I am also checking if the function called is `pure` or `internal` and if this is the case, I'm assuming that they don't modify an array. I have also added several new test cases to cover it.

0xalpharush:
@montyly I think we can merge and follow up in a separate PR

PR #1695, opened 2023-02-26
WIKI URL fixed

bart1e:
Quote (") sign removed from the end of detector's WIKI URL.


0xalpharush (approved):

montyly:
great catch @bart1e, thanks

PR #1699, opened 2023-02-27
Compare upgrades in `slither.utils.upgradeability`

webthethird:
Based on @montyly's recommendation re: adding what I had been doing in my [diff_with_v2.py](https://github.com/webthethird/slither/blob/dev-upgradeable-proxy-cleanup/slither/tools/upgradeability/checks/diff_with_v2.py) directly within Slither, instead of as a new upgradeability check.

This adds a new `slither.utils.upgradeability` containing a function `compare(v1: Contract, v2: Contract)`, which returns a dictionary of lists of Variables/Functions, and a helper function `is_function_modified(f1: Function, f2: Function)`, which returns True if the function is modified or False otherwise.

The dictionary returned by the compare function has the following lists:
```python
{
    "missing-vars-in-v2": list[Variable],
    "new-variables": list[Variable],
    "tainted-variables": list[Variable],
    "new-functions": list[Function],
    "modified-functions": list[Function],
    "tainted-functions": list[Function]
}
```
In this context, a tainted variable is a preexisting variable which is read or written by a new or modified function, and similarly a tainted function is an unmodified function that calls a modified function.

I made sure to address the feedback he gave me:
- [x] You need to check of immutable [here](https://github.com/webthethird/slither/blob/dev-upgradeable-proxy-cleanup/slither/tools/upgradeability/checks/diff_with_v2.py#L46-L47)
- [x] An issue with the comparison with get_summary, is that it might miss a change in the logic of the function. Like if a condition is inverted for example. One thing that we could use is [function.source_mapping.content_hash](https://github.com/crytic/slither/blob/dev/slither/core/source_mapping/source_mapping.py#L89) , this returns an hash for the function source code. So if the source code differ you can detect it. It’s fast, however a downside is that if there is a change in a name, or in a comment, it will lead to a false positive. A better approach would be to compare the CFG of the two function; because we are looking for strict equivalence here, it should not be too difficult to do. Even a naive approach to walk through the CFG and compare the IR opcode should give good result I think.


webthethird:
This was blocked by a bug in the CI tests, but all checks are passing now

PR #1707, opened 2023-03-04
Gas optimizations

demis1997:
### Description
Added 19 gas inefficiency detectors, I can create a separate tool that checks only the gas inefficiencies instead of having slither check vulnerabilities and gas inefficiencies. I can also create multiple branches and only have 1 detector per branch if that would make things easier for you. I figured many audit teams are checking for gas optimisations in certain contracts and can make a big difference depending on the type of contract being audited.

### Build worked on
I started working on this around December and tested today to see if any errors were coming up with the updated version of slither. Other than some minor mistakes on my end, it did not seem to mess up the rest of Slither's functionality
### Some changes that need to happen
I will be making changes to the code as some of it is not 100% accurate when detecting and needs more restrictions so I will make further pull requests after an initial review of what to keep, what to ditch and what to change.
One change that needs to happen for each file is the wiki address which currently points to my github and will need to add the wiki info in the official slither repo.
The code is not even throughout and there are random lines of code that should have been written differently to follow the same format as the rest of the files.

### Gas inefficiency detector list
The checks I am doing are the following:

- if array length is being read in every iteration of a for loop
- if the function is external and read only, to use calldata as storage location rather than memory
- if the increments within a for loops condition can be made unchecked
- checking if the contract can be deployed as clone
- checking to see if you can change keccak variables from constant to immutable
- check if they are using revert string and instead suggest to use custom errors
- checking if variables are explicitly set with their default values  (address with 0,  0 for uint, false for bool etc)
- checking if you can use double require rather than && in require statements
- checks if i variables are initialised within loops
- checking if state variables can be changed to immutable
- checking to use >= and <= rather than > and < to save gas
- check if internal view functions can be used in a modifier for require statements
- checks if it is reading from storage more than once and suggests to cache the variables
- checking if modifiers can be used rather than functions
- checks if !=0 can be used instead of >0
- checks if increments are done after the variable or before so i++ or ++i
- checks if shift right and shift left can be used instead of division and multiplication
- checks if it is fetching the reference in a map or array or uses a storage variable 
- checks if it is using ! = 0 within require statements rather than > 0, and > 0 outside require statements rather than ! = 0

### DIFF
--- a/slither/detectors/functions/inequality.py
+++ b/slither/detectors/functions/inequality.py
@@ -1,0 +1,11 @@
+from slither import Slither
+from slither import InefficientInequalityDetector
+
+slither = Slither('')
+detector = InefficientInequalityDetector(slither.contracts)
+detector.analyze()
+results = detector.get_results()
+for contract_name in results:
+    print(f"Results for {contract_name}:")
+    for result in results[contract_name]:
+        print(f"\t{result.description}")
--- a/slither/detectors/gas-inefficiencies/array_length_cache.py
+++ b/slither/detectors/gas-inefficiencies/array_length_cache.py
@@ -2,14 +2,9 @@
 Gas: Module detecting Array length inside of loop
 
 """
-from collections import defaultdict
-from slither.core.solidity_types.elementary_type import ElementaryType
 
+from slither import solc_parsing
 from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
-from slither.detectors.abstract_detector import Detection
-from slither import CFG
-from slither.core.solidity_types import ArrayType
-from slither.slithir.operations import Load
 
 class GasInefficientLoopLength(AbstractDetector):
     """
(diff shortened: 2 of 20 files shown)


montyly:
Wow this is huge and awesome, thanks @demis1997

montyly:
@demis1997: the detectors are using API that do not exist (`Load` / `CFG`/...). Are you using a separate fork?

Additionally, the detector object should use `_detect`, see the doc: https://github.com/crytic/slither/wiki/Adding-a-new-detector

demis1997:
> @demis1997: the detectors are using API that do not exist (`Load` / `CFG`/...). Are you using a separate fork?
> 
> Additionally, the detector object should use `_detect`, see the doc: https://github.com/crytic/slither/wiki/Adding-a-new-detector
I will change the objects to use _detect also and update the build with api

demis1997:
> @demis1997: the detectors are using API that do not exist (`Load` / `CFG`/...). Are you using a separate fork?
> 
> Additionally, the detector object should use `_detect`, see the doc: https://github.com/crytic/slither/wiki/Adding-a-new-detector I will change the objects to use _detect also and update the build with api

Additionally I am looking into the 3 tests that failed:
  Lint Code Base / Lint Code Base (pull_request)
  Lint Code Base / Lint Code Base (pull_request)
  CI / tests (ubuntu-latest, dapp) (pull_request)

demis1997:
@montyly let me know what other changes you need, like i said earlier it's getting the descriptions for each gas inefficiency from my account so I will have to add into crytics wiki each description and point there instead

0xalpharush:
It would be great if you could provide tests (https://github.com/crytic/slither/blob/dev/CONTRIBUTING.md#adding-detector-tests) and for each detector document how much gas savings there are for the latest version of solc with optimizations enabled. If you're willing to do so, please open an issue with this info and we can discuss which detectors to upstream

PR #1710, opened 2023-03-05
replace decode/encode_single with decode/encode | formatting

8times4:
decode/encode_single was removed from eth-abi, so this change replaces it with the regular decode/encode func. 
https://github.com/ethereum/eth-abi/commit/b03717187800e1bd0f556a70987becb9acf478a0

move eth stuff out from the try-except, since no matter the error, it throws pip3 install web3 for everything


0xalpharush:
Replaced by https://github.com/crytic/slither/pull/1743

Issue #1700, opened 2023-02-28
"Set up cachix" action is failing on PRs


webthethird:
Action failed with error: Error: Unable to locate executable file: cachix. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.

![image](https://user-images.githubusercontent.com/6509209/221960938-9bfc7547-6124-4b24-b241-dea5b747fb00.png)

This doesn't seem to have anything to do with the PR itself, though I could be wrong.

PR #1711, opened 2023-03-06
upgrade nix installation to fix CI installation

0xalpharush:
closes https://github.com/crytic/slither/issues/1700

ref https://github.com/cachix/cachix-action/issues/144


elopez (approved):
Thanks! Can you also update `cachix/cachix-action` to the latest version? Going forward, we should probably enable dependabot action updates on this repo.

Issue #1645, opened 2023-02-02
fix compact tests that are mistakenly legacy


0xalpharush:
Some compact tests were overwritten as legacy in https://github.com/crytic/slither/pull/1118
E.g. this file is in the legacy AST format https://github.com/crytic/slither/blob/master/tests/ast-parsing/compile/continue-all.sol-0.4.0-compact.zip
We need to fix this and perform some validation to prevent regressions

PR #1712, opened 2023-03-06
remove mislabeled tests, update 0.4.12 tests to be compact

0xalpharush:
closes https://github.com/crytic/slither/issues/1645

Some tests with solc versions prior to 0.4.12 are labeled as compact despite the compact format not being available and actually being in the legacy format (duplicated). In addition, the 0.4.12 tests labeled as compact were legacy because of a bug in crytic-compile that's fixed here https://github.com/crytic/crytic-compile/pull/398. This PR adds validation to the AST tests to prevent regressions, removes duplicated artifacts, and corrects the 0.4.12 compact tests.


0xalpharush:
Rinkeby is deprecated so the etherscan integration tests return 404

PR #1713, opened 2023-03-06
incorrect-equality: do not check addresses

ydm:
This commit adds a simple additional check to make sure no strict equality of addresses is flagged at all.

### DIFF
--- a/slither/detectors/statements/incorrect_strict_equality.py
+++ b/slither/detectors/statements/incorrect_strict_equality.py
@@ -73,6 +73,14 @@ contract Crowdsale{
         return isinstance(ir, Binary) and ir.type == BinaryType.EQUAL
 
     @staticmethod
+    def is_not_comparing_addresses(ir: Binary) -> bool:
+        """
+        Comparing addresses strictly should not be flagged.
+        """
+        addr = ElementaryType("address")
+        return ir.variable_left.type != addr or ir.variable_right.type != addr
+
+    @staticmethod
     def is_any_tainted(
         variables: List[
             Union[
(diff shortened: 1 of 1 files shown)


ydm:
i really don't think these errors have anything to do with my changes, can you please rerun the ci?

elopez:
Hi @ydm ! Those CI failures were fixed in #1723 and #1711. Can you merge the `master` branch and push to resolve the issue?

ydm:
should be good now?

slither/detectors/statements/incorrect_strict_equality.py
```diff
@@ -72,6 +72,14 @@ class IncorrectStrictEquality(AbstractDetector):
     def is_direct_comparison(ir: Operation) -> bool:
         return isinstance(ir, Binary) and ir.type == BinaryType.EQUAL
 
+    @staticmethod
+    def is_not_comparing_addresses(ir: Binary) -> bool:
+        """
+        Comparing addresses strictly should not be flagged.
+        """
+        addr = ElementaryType("address")
+        return ir.variable_left.type != addr or ir.variable_right.type != addr
```
0xalpharush, line 81: If something like `Contract(address(0x0)) == Contract(address(0x0))` is used, `is_not_comparing_addresses` won't filter it out. We would need to add something like `isinstance(ir.variable_left.type, UserDefinedType) and isinstance(ir.variable_left.type.type, Contract)`. It'd be helpful to add a test case for both the address and contract case so that we can be sure of the modifications and prevent regressions in the future.

ydm: alright, will do that

PaulRBerg:
Casting my support of this PR.

The [wiki](https://github.com/crytic/slither/wiki/Detector-Documentation#dangerous-strict-equalities) about this detector mentions a use case that should never be relevant for address checks.

montyly:
Thanks @ydm for the contribution. We merged this PR through #1890

ydm:
> Thanks @ydm for the contribution. We merged this PR through #1890

I'm truly sorry for taking so long to fix this PR!

Issue #1683, opened 2023-02-23
SSA: function parameter use index 1 instead of 0


montyly:
```solidity
contract A{
	function f(int a, int b) public returns(int){
		return a+b;
	}
}
```
$ `slither test.sol --print slithir-ssa`

	Function A.f(int256,int256)
		Expression: a + b
		IRs:
			TMP_0(int256) = a_1 (c)+ b_1
			RETURN TMP_0
```

Here `a_1` should be `a_0`. In particular this will break the taint analysis on the SSA API

Troublor:
Can this bug be avoided temporarily by rolling back to old slither versions?

Issue #473, opened 2020-04-29
Incorrect SSA for conditional after previous dominating conditional


aaronyoo:
Slither generates incorrect SSA for the following program:
```solidity
pragma solidity >=0.4.16 <0.7.0;
contract Contract {
    function f() public returns (int) {
        int a = 1;
        if (a > 0) {
            a = 2;
        }
        if (a == 3) {
            a = 6;
        }
        return a;
    }
}
```
```
f:
	a_1(int256) := 1(int256)

	TMP_0(bool) = a_1 > 0
	CONDITION TMP_0
		a_2(int256) := 2(int256)
	a_3(int256) := ϕ(['a_2', 'a_1'])

	TMP_1(bool) = a_3 == 3
	CONDITION TMP_1
		a_4(int256) := 6(int256)
	a_5(int256) := ϕ(['a_4', 'a_1'])    // ERROR

	RETURN a_5
```
The ERROR line should be replaced with `a_5(int256) := ϕ(['a_4', 'a_3'])`.

PR #1715, opened 2023-03-06
[Draft] SSA IR Fix Phi Handling

bsamuels453:
Derived from #1102 

This PR changes Phi handling so that Phis are injected at dominance frontiers. 

The current implementation adds empty Phi nodes which are later removed and cause variable instances to be missing a definition. 

Linked issues: 
- #434
- #473
- #1683 

These changes also improves the ArbitrarySendErc20 detector, which can now detect test case `bad2`: https://github.com/crytic/slither/blob/aa3171ee15b8be88ba6a24f494465ece183f19f8/tests/detectors/arbitrary-send-erc20/0.4.25/arbitrary_send_erc20.sol#L44-L46

This PR also fixes a previously unknown issue where the existence of a Phi node is incorrectly interpreted as a variable write when `variables_written` is constructed. This bug introduced some small false positives, so their unit tests needed to be updated.


0xalpharush:
> These changes also improves the ArbitrarySendErc20 detector, which can now detect test case bad2:
I'm not sure it fixed this as the data dependency tests are failing.

bsamuels453:
I'm about go to on a project for a month, so if anyone wants to adopt this PR in the meantime, feel free.

Right now it's failing the data dependency tests; specifically the one for tracking data dependencies through function args. 
```
contract PropagateThroughArguments {
    uint var_tainted;
    uint var_not_tainted;
    uint var_dependant;

    function f(uint user_input) public {
        f2(user_input, 4);
        var_dependant = var_tainted;
    }

    function f2(uint x, uint y) internal {
        uint z = x + 1;
        var_tainted = z;
        var_not_tainted = y;
    }
}
```
generates:
```
        Function PropagateThroughArguments.f2(uint256,uint256)
                Expression: z = x + 1
                IRs:
                        TMP_1(uint256) = x_0 + 1
                        z_1(uint256) := TMP_1(uint256)
```
The ssa IR is missing a phi statement for `x_0 = phi(user_input_0)`

Normally I believe fix_phi  would create these phi statements, so I was trying to port that over for the new SSA generation but running into infinite loop issues.

PR #1722, opened 2023-03-08
deregister_detector

sidarth16:
This function would be helpful to **_de-register_** a detector class from the list of detectors already registered with the slither object.

### DIFF
--- a/slither/slither.py
+++ b/slither/slither.py
@@ -194,7 +194,7 @@ class Slither(SlitherCore):  # pylint: disable=too-many-instance-attributes
         """
 
         for obj in self._detectors:
-            if type(obj) == detector_class :
+            if type(obj) == detector_class :   # pylint: disable=unidiomatic-typecheck
                 self._detectors.remove(obj)
                 return
 
--- a/slither/slither.py
+++ b/slither/slither.py
@@ -187,14 +187,14 @@ class Slither(SlitherCore):  # pylint: disable=too-many-instance-attributes
         for compilation_unit in self.compilation_units:
             instance = detector_class(compilation_unit, self, logger_detector)
             self._detectors.append(instance)
-            
+
     def deregister_detector(self, detector_class: Type[AbstractDetector]) -> None:
         """
         :param detector_class: Class inheriting from `AbstractDetector`.
         """
 
         for obj in self._detectors:
-            if type(obj) == detector_class :   # pylint: disable=unidiomatic-typecheck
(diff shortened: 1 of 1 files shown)


sidarth16:
The major issue it solves is the reusing of the same slither Object for multiple detector classes.
Example: 
```
from slither.detectors.all_detectors import *

sl = Slither("<contract_address>")

detectors = [
    TxOrigin, ReentrancyBenign, ReentrancyEth, 
    UncheckedLowLevel, UncheckedTransfer, UncheckedSend
]
for d in detectors:
    sl.register_detector(d)  
r1 = sl.run_detectors()

# Now I no more want TxOrigin in the context slither object detectors

sl.deregister_detector(TxOrigin)

r2 = sl.run_detectors()

```

Thus, We are now able to re-run detectors , reusing the same slither object. We don't need to declare new slither object for each combination of detector classes.

Here r1, contains results of TxOrigin , while r2 does not contain results of TxOrigin .

slither/slither.py
```diff
@@ -187,14 +187,14 @@ def register_detector(self, detector_class: Type[AbstractDetector]) -> None:
         for compilation_unit in self.compilation_units:
             instance = detector_class(compilation_unit, self, logger_detector)
             self._detectors.append(instance)
-            
+
     def deregister_detector(self, detector_class: Type[AbstractDetector]) -> None:
```
0xalpharush, line 191:
```suggestion
    def unregister_detector(self, detector_class: Type[AbstractDetector]) -> None:
```

slither/slither.py
```diff
@@ -187,14 +187,14 @@ def register_detector(self, detector_class: Type[AbstractDetector]) -> None:
...
         """
         :param detector_class: Class inheriting from `AbstractDetector`.
         """
 
         for obj in self._detectors:
-            if type(obj) == detector_class :   # pylint: disable=unidiomatic-typecheck
+            if type(obj) == detector_class:  # pylint: disable=unidiomatic-typecheck
```
0xalpharush, line 197: Can you use `isinstance`?

0xalpharush:
It would be great to have a test case. I haven't looked into it but this might have weird effects on detector result filtering.

slither/slither.py
sidarth16: Yaa sure

montyly:
Note that some detectors cache the results (ex: the reentrancies detectors), so running multiple times will not lead to a fresh run.

I am not sure what is your goal with this API, but you should be aware that detectors are not side-effects free ;)

PR #1724, opened 2023-03-08
deregister_printer

sidarth16:
This function would be helpful to _**de-register a printer class**_ from the list of _printers already registered with the slither object.

### DIFF
--- a/slither/slither.py
+++ b/slither/slither.py
@@ -196,7 +196,17 @@ class Slither(SlitherCore):  # pylint: disable=too-many-instance-attributes
 
         instance = printer_class(self, logger_printer)
         self._printers.append(instance)
-
+    
+    def deregister_printer(self, printer_class: Type[AbstractPrinter]) -> None:
+        """
+        :param printer_class: Class inheriting from `AbstractPrinter`.
+        """
+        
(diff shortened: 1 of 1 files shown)


0xalpharush:
What is the use case? Can't you only register specific detectors and leave out the one's that are not desired?

sidarth16:
> What is the use case? Can't you only register specific detectors and leave out the ones that are not desired?

The major issue it solves is the reusing of the same slither Object for multiple printer and detector classes.
Example: 
```
from slither.detectors.all_detectors import *

sl = Slither("<contract_address>")

detectors = [
    TxOrigin, ReentrancyBenign, ReentrancyEth, 
    UncheckedLowLevel, UncheckedTransfer, UncheckedSend
]
for d in detectors:
    sl.register_detector(d)  
r1 = sl.run_detectors()

# Now I no more want TxOrigin in the context slither object detectors

sl.deregister_detector(TxOrigin)

r2 = sl.run_detectors()

```

Thus, We are now able to re-run detectors and printers , reusing the same slither object. We don't need to declare new slither object for each combination of detector / printer classes.

Here r1, contains results of TxOrigin , while r2 does not contain results of TxOrigin

PR #1730, opened 2023-03-08
Generate interface code in new `slither.utils.code_generation`

webthethird:
Adds `Function.interface_signature_str`, `Structure.interface_def_str()` and `Contract.generate_interface()`.

The new Contract method uses the other two to generate Solidity code for an interface to the contract.

Still isn't quite as clean as the code @glarregay-tob wrote outside of Slither, which I think handles struct return types better.


0xalpharush:
I think this should not be on the core object and should instead be a utility like `generate_interface(Contract)`

webthethird:
> I think this should not be on the core object and should instead be a utility like `generate_interface(Contract)`

That sounds reasonable enough. Do you think I should start a new util in `slither.utils`? Perhaps one for everything related to code generation?

0xalpharush:
@webthethird That sounds like a good idea

webthethird:
Looks like the failing CI checks related to linting are unrelated to this PR:
`Pull down action image 'ghcr.io/github/super-linter:slim-v4.9.2'`
`Error: Docker pull failed with exit code 1`

slither/utils/code_generation.py
```diff
@@ -0,0 +1,96 @@
...
+        func: A Function object
+
+    Returns:
+        The function interface as a str (contains the return values).
+        Returns None if the function is private or internal, or is a constructor/fallback/receive.
+    """
+
+    name, parameters, return_vars = func.signature
+    visibility = func.visibility
+    if (
```
montyly, line 54: Maybe `contract.functions_entry_points` instead of the visibility check?  https://github.com/crytic/slither/blob/master/slither/core/declarations/contract.py

It will also remove the duplicate functions due to shadowing

montyly (commented):
can we add tests? (cc @0xalpharush )

webthethird:
> can we add tests? (cc @0xalpharush )

Would we just test this by having an input .sol file and an expected output .sol file, and doing a diff between the expected and actual output?

PR #1743, opened 2023-03-10
slither-read-storage explicit dependency group

0xalpharush:


setup.py
```diff
@@ -30,7 +30,8 @@
             "solc-select>=v1.0.0b1",
             "openai",
             "pdoc",
-        ]
+        ],
+        "read-storage": ["solc-select>=v1.0.0b1", "web3>=6.0.0b"],
```
elopez, line 34: Is solc-select a dependency? (I was thinking this might be useful to install a "fully featured" Slither with brew, but solc-select so far is not a sliither dependency)

0xalpharush: To use slither with Etherscan, I think solc-select is required

PR #1761, opened 2023-03-16
Fix IR operation when initializing array with one-element array literal 

### LINKED ISSUE #1748: [Bug]: Assigning to array literal (len==1) to state variable results in invalid IR operation
### Describe the issue:

When assigning an array literal to a state variable works fine when the array literal has at least two elements. 
For example:
`arr = [1,2]` results in the IR operation `arr(uint256[]) =  ['1', '2']`.

However, when the array literal only has one element, the `rvalue` of the IR operation becomes a single element instead of an array:
`arr = [1]` results in `arr(uint256[]) := 1(uint256)`
This is buggy since the type of `lvalue` and `rvalue` mismatch. 
`arr(uint256[]) = ['1']` is expected in this case.

It seems like when the array literal's length is 1, Slither will consider this as an `Assignment` operation instead of `InitArray` operation, while the latter one is expected. 

```solidity
contract A {
    uint[] arr;
    function foo(uint i) public returns (uint) {
        arr = [1];
        return arr[i];
    }
}
```

### Version:

0.9.2

### Relevant log output:

```shell
Contract A
        Function A.foo(uint256) (*)
                Expression: arr = (1)
                IRs:
                        arr(uint256[]) := 1(uint256) # Buggy IR Operation
                Expression: arr[i]
                IRs:
                        REF_0(uint256) -> arr[i]
                        RETURN REF_0
```

Troublor:
Fix #1748 

I added one test to reveal the bug, but I am not sure whether `tests/test_ssa_generation.py` is the right place to put it. Correct me if it is not the right place. 

As I mentioned in the comments of #1748, the root cause is that the AST visitor unpacks the right-hand side value of `arr = [1]` when visiting the AST tree. However, it is not possible to decide whether the tuple should be unpacked or not when visiting `TupleExpression` AST node. 
So, the workaround I made in this PR is to recover the array literal when we later find (at L223 of file expression_to_slithir.py) that an array is being initialized.

### DIFF
--- a/slither/visitors/slithir/expression_to_slithir.py
+++ b/slither/visitors/slithir/expression_to_slithir.py
@@ -220,6 +220,12 @@ class ExpressionToSlithIR(ExpressionVisitor):
                 operation.set_expression(expression)
                 self._result.append(operation)
                 set_val(expression, left)
+            elif isinstance(left.type, ArrayType):
+                # Special case for init of array, when the right has only one element
+                operation = InitArray([right], left)
+                operation.set_expression(expression)
+                self._result.append(operation)
+                set_val(expression, left)
             else:
                 operation = convert_assignment(
                     left, right, expression.type, expression.expression_return_type
--- a/tests/test_ssa_generation.py
+++ b/tests/test_ssa_generation.py
@@ -27,6 +27,7 @@ from slither.slithir.operations import (
     BinaryType,
     InternalCall,
     Index,
+    InitArray,
 )
 from slither.slithir.utils.ssa import is_used_later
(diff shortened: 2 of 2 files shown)


0xalpharush:
The failure for slither-read-storage is unrelated and fixed in https://github.com/crytic/slither/pull/1743

Troublor:
I found the error messages of `Lint Code Base / Pylint` are related to `ERRORS FOUND in MARKDOWN:[1]` and `Failed to call GitHub Status API!`. 
They are not related to my commit I think.

PR #1766, opened 2023-03-17
upgrade-prettytable, use colored table

0xalpharush:


montyly:
Should we add a config option to change the default theme?

Issue #1251, opened 2022-06-20
Detector Documentation: Solidity Compiler Versions


pcaversaccio:
Based on a recent exchange on [Twitter](https://twitter.com/pcaversaccio/status/1538889927622008832), I think it would be a good idea to clarify why the specific Solidity compiler versions are chosen as recommendations [here](https://github.com/crytic/slither/wiki/Detector-Documentation#recommendation-99): 
![image](https://user-images.githubusercontent.com/25297591/174627915-5ee14036-ecfa-4e3d-a057-0fda14d1dc39.png)

montyly:
Hi @pcaversaccio, thanks for asking, this is a great question.

For 0.5/0.6/0.7 the recomended versions are the latest ones.

For 0.8, the recommendations are currently outdated, and will be update before the next release.

Regarding our process, we take into consideration:
- The known bugs
- The risk of new features
- The time since the release - the more time has passed, the less likely the version is to have an unknown bug. 

We do not recommend to use the latest version for deployment due to the risks related to recent compiler's updates. However we do recommend to use the latest version for testing (as new version of the compiler might integrate new checks).

We also run fuzzing campaign against solc from time to time to help us in our evaluation (see https://blog.trailofbits.com/2020/06/05/breaking-the-solidity-compiler-with-a-fuzzer/ and https://blog.trailofbits.com/2021/03/23/a-year-in-the-life-of-a-compiler-fuzzing-campaign/)

pcaversaccio:
@montyly thanks for the swift reply! I do understand your process taken and can follow the arguments (in particular regarding the time since release as we've seen just recently with Solidity version `0.8.15` which fixed an optimizer bug for `0.8.13` and `0.8.14`). What I'm asking you guys however is to provide more technical details on the selection in order to understand what exactly are the key drivers behind e.g. not recommending `0.6.10` but recommending `0.6.11`. This might be a highly technical task but is still important due to the sake of transparency. It's important to follow the mantra: verify what you actually run without trusting the compiler! And I want to verify Slither's output now ;-)

pcaversaccio:
for the sake of transparency, ToB will write a FAQ: https://twitter.com/dguido/status/1538911430921142277

lawrencexia:
@montyly do you know what the 0.8 recommended version will be for the next release?

pcaversaccio:
Any update on this @montyly?

montyly:
[...]

PR #1773, opened 2023-03-17
Update solc version recommendations

montyly:
- Recommend 0.8.18 (0.8.19 being less than 1 month old)
- Remove recommendations for 0.5/0.6/0.7 branches. 0.8 has been around for long enough that everyone should use it.


montyly:
Close https://github.com/crytic/slither/issues/1251

PR #1777, opened 2023-03-17
Fix incorrect temporary variable type in NewArray operation

### LINKED ISSUE #1776: [Bug]: Incorrect variable type for NewArray operation
### Describe the issue:

When using `new` expression to initialize a dynamic array, e.g., `uint[] memory arr = new uint[](2);`, slither convert it to a `NewArray` IR operation. 
However, the type of `lvalue`, `arr` variable in the example, is incorrectly set as `uint256` instead of `uint256[]`. 

```solidity
contract A {
    function foo() public returns (uint) {
        uint[] memory arr = new uint[](2);
        return 0;
    }
}
```

### Version:

0.9.2

### Relevant log output:

```shell
Contract A
        Function A.foo() (*)
                Expression: arr = new uint256[](2)
                IRs:
                        TMP_1 = new uint256[](2)
                        arr(uint256[]) := TMP_1(uint256) # TMP_1 has wrong type
                Expression: 0
                IRs:
                        RETURN 0
```

Troublor:
Fix #1776 

When setting the type of the temporary variable from `NewArray` expression, the original code directly use the array base type instead of the array type. 
This PR re-construct the `ArrayType` using the base type and the depth at L792-797 of file slither/slithir/convert.py. 

One imperfect point is that length of nested array is lost since it is not stored in the `NewArray` expression, in other words, `new uint[10][](2);` expression will become `new uint[][](2)`. 
Fixing this will result in a great modification in the `NewArray` class, which may induce breaking change. So, I just leave it for future work.

### DIFF
--- a/slither/slithir/convert.py
+++ b/slither/slithir/convert.py
@@ -789,7 +789,13 @@ def propagate_types(ir: Operation, node: "Node"):  # pylint: disable=too-many-lo
                                 if v:
                                     ir.lvalue.set_type(v.type)
             elif isinstance(ir, NewArray):
-                ir.lvalue.set_type(ir.array_type)
+                # Reconstruct the array type from base type and depth
+                # Note that the length of the array type is lost since NewArray expression does not store it.
+                # So, here we over-approximately set the length to None (dynamic array).
+                typ = ir.array_type
+                for _ in range(ir.depth):
+                    typ = ArrayType(typ, None)
+                ir.lvalue.set_type(typ)
             elif isinstance(ir, NewContract):
                 contract = node.file_scope.get_contract_from_name(ir.contract_name)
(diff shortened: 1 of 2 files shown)


0xalpharush:
Thanks for identifying this issue! While it's reasonable to make incremental changes, it may be possible to fix this in one go. I haven't tested it a ton but I think something like this [branch](https://github.com/crytic/slither/tree/array-type) would simplify and make the IR more accurate if you'd like to work of it and add additional tests like for the case you mention (`new uint[10][](2);`).

Troublor:
@0xalpharush Sure. I would like to work on that enhancement. Please assign the issue to me.

Troublor:
Closed in favor of #1784

PR #1784, opened 2023-03-21
Make type information of NewArray more precise

### LINKED ISSUE #1776: [Bug]: Incorrect variable type for NewArray operation

Troublor:
Fix #1776, i.e., incorrect `array_type` in `NewArray` operation. 
This PR also makes the `array_type` field more precise by including the length of the fixed-sized nested array.


tests/test_ssa_generation.py
```diff
@@ -1077,3 +1079,28 @@ def test_issue_1748():
...
+        new_op = operations[0]
+        lvalue = new_op.lvalue
+        lvalue_type = lvalue.type
+        assert isinstance(lvalue_type, ArrayType)
+        lvalue_type1 = lvalue_type.type
+        assert isinstance(lvalue_type1, ArrayType)
+        assert lvalue_type1.length_value.value == "10"
+        lvalue_type2 = lvalue_type1.type
+        assert isinstance(lvalue_type2, ArrayType)
+        assert lvalue_type2.length_value.value == "5"
```
0xalpharush, line 1106: This seems backwards?

Troublor: What do you mean `backwards`? 
Here, I am checking the nested array type and length should be parsed and converted to IR correctly.

0xalpharush: Nvm, I misunderstood which length value corresponded to which type. This seems correct

tests/test_ssa_generation.py
```diff
@@ -1077,3 +1079,28 @@ def test_issue_1748():
...
+    """
+    with slither_from_source(source) as slither:
+        c = slither.get_contract_from_name("Contract")[0]
+        f = c.functions[0]
+        operations = f.slithir_operations
+        new_op = operations[0]
+        lvalue = new_op.lvalue
+        lvalue_type = lvalue.type
+        assert isinstance(lvalue_type, ArrayType)
+        lvalue_type1 = lvalue_type.type
```
0xalpharush, line 1101:
```suggestion
        lvalue_type1 = lvalue_type.type
        assert lvalue_type1.is_dynamic
```
We can add the assertion that this array is dynamically sized and perhaps the inverse for those statically sized

Troublor: It seems to me that only lvalue_type should be dynamic.
lvalue_type1 and lvalue_type2 are statically sized.
I have updated the test according to your suggestion in 4768a08.

0xalpharush:
I think this makes sense but it will need @montyly's review too. Would you make this change against `dev` if possible please?

Troublor:
> I think this makes sense but it will need @montyly's review too. Would you make this change against `dev` if possible please?

The base branch of this PR has been made to `dev` branch.

0xalpharush (approved):

0xalpharush:
@Troublor Do you mind resolving the merge conflicts please?

montyly:
This is a great improvement, thanks @Troublor

Issue #1783, opened 2023-03-20
[Bug]: AssertionError in `UserDefinedType.__init__()`


webthethird:
### Describe the issue:

I've got a fuzz testing contract that I generated using my differential fuzzing test generator, and it compiles, but I was getting an error when trying to test it with Echidna. So I ran `slither --print echidna ./src/test/compound/DiffFuzzCustomInitialization.sol` and saw the AssertionError below. 

It appears to be asserting that `None` is either a Contract, Structure or Enum.

### Code example to reproduce the issue:

https://github.com/webthethird/solidity-diff-fuzz-upgrades/blob/main/src/test/compound/DiffFuzzCustomInitialization.sol
https://github.com/webthethird/solidity-diff-fuzz-upgrades/blob/main/src/test/compound/DiffFuzzUpgrades.sol

### Version:

0.9.2

### Relevant log output:

```shell
Error in /home/webthethird/Ethereum/solidity-diff-fuzz-upgrades/src/test/compound/DiffFuzzCustomInitialization.sol
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/__main__.py", line 834, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/__main__.py", line 98, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/__main__.py", line 76, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/slither.py", line 133, in __init__
    self._init_parsing_and_analyses(kwargs.get("skip_analyze", False))
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/slither.py", line 153, in _init_parsing_and_analyses
    raise e
  File "/usr/local/lib/python3.10/dist-packages/slither_analyzer-0.9.2-py3.10.egg/slither/slither.py", line 149, in _init_parsing_and_analyses
    parser.analyze_contracts()
[...]

PR #1539, opened 2022-12-31
FileScope: apply renaming in get_contract_from_name()

SheldonHolmgren:
This fixes a bug that was making slither crash for https://github.com/makerdao/dss.

Before this patch:
```bash
makerdao/dss $ slither . --dapp-ignore-compile
Function not found suck
Impossible to generate IR for PublicClip.kick:
 'NoneType' object has no attribute 'type'
Function not found suck
Impossible to generate IR for PublicClip.redo:
 'NoneType' object has no attribute 'type'
Function not found flux
Impossible to generate IR for PublicClip.take:
 'NoneType' object has no attribute 'type'
Function not found flux
Impossible to generate IR for PublicClip.yank:
 'NoneType' object has no attribute 'type'
Function not found nope
Impossible to generate IR for Cat_.file:
 'NoneType' object has no attribute 'type'
Function not found ilks
Impossible to generate IR for Cat_.bite:
 'NoneType' object has no attribute 'type'
Function not found hope
Impossible to generate IR for TestVow.constructor:
 'NoneType' object has no attribute 'type'
Function not found nope
Impossible to generate IR for TestVow.file:
 'NoneType' object has no attribute 'type'
Function not found dai
Impossible to generate IR for TestVow.heal:
 'NoneType' object has no attribute 'type'
Function not found dai
Impossible to generate IR for TestVow.kiss:
 'NoneType' object has no attribute 'type'
Function not found sin
Impossible to generate IR for TestVow.flop:
 'NoneType' object has no attribute 'type'
Function not found dai
Impossible to generate IR for TestVow.flap:
 'NoneType' object has no attribute 'type'
Function not found dai
Impossible to generate IR for TestVow.cage:
 'NoneType' object has no attribute 'type'
Function not found sin
Impossible to generate IR for TestVow.Awe:
 'NoneType' object has no attribute 'type'
Function not found dai
Impossible to generate IR for TestVow.Joy:
 'NoneType' object has no attribute 'type'
Traceback (most recent call last):
  File "/home/holmgren/dev/slither/slither/__main__.py", line 867, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/home/holmgren/dev/slither/slither/__main__.py", line 97, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/home/holmgren/dev/slither/slither/__main__.py", line 75, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
[...]

PR #1786, opened 2023-03-21
Fix missing contract in `slithir.convert.propagate_types` due to import renaming

webthethird:
For NewContract ir, check for renaming in import if contract is not found using local name.

Addresses #1783


montyly:
Looks like there is a conflict with the `dev` branch. Can we also add a unit test?

0xalpharush:
I think this solution handles less issues than https://github.com/crytic/slither/pull/1539. However, I am not 100% sure either will work considering issues like https://github.com/crytic/slither/issues/1809 where collisions in the namespace occur due to merging scopes. That is, a renaming may existing in two scopes and be incorrectly overwritten with the wrong identifier

webthethird:
I'll go ahead and close this, since the issue is more likely to be fixed elsewhere.

PR #1792, opened 2023-03-22
coverage workflow

0xalpharush:
experiment


montyly:
I love the coverage report, but I am a bit unsure about doing things in sequential instead of parallel.

Right now the action takes 7 min to complete. I think it would make sense to keep separate actions for
- end to end detectors (which we can optimize a lot by pre-compiling the contracts)
- end to end parsing
- unit tests (features + IR tests + storage layout)

PR #1796, opened 2023-03-23
fold constant of hash expressions

0xalpharush:
Evaluate keccak on constants to enable folding of expressions like `bytes32 constant IMPLEMENTATION_SLOT = bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)`. Also updated the `Literal` type of rational constants to be `uint256`.


webthethird:
I'll be keeping an eye on this, as it will be very useful for cleaning up #1752 :eyes:

PR #1799, opened 2023-03-23
testing strategy + coverage workflow

0xalpharush:


montyly:
This is great.
- I think we should revert merging all the pytest run into one github action, because this is making the github action slow (see https://github.com/crytic/slither/pull/1792#issuecomment-1481854226).
- Can we improve/automatize the `TEST_DATA_DIR`? The current `Slither(Path(TEST_DATA_DIR, "test.sol").as_posix())` does not seem really easy to manipulate
- The `test_data` structure is a bit confusing
   - For example, we have now `unit/core/test_data`, which can contain a sub directory per file `test_X.py`, but there are also artefacts in `unit/core/test_data` directly.
   - What will happen if we have a `unit/core/contracts` ? should we then have `unit/core/contracts/test_data`?
- We need to create guidelines and rules for the tests
   - They should explain the structure of the tests folder
   - They should discuss the usage of solc-select
   - The guidelines should answer "I want to add a new tests for my new component, where/how to do it?"

0xalpharush:
@montyly I've created a matrix divided into `integration`, `unit`, and `tools` and disabled coverage on Windows (the coverage report was already only generated on Ubuntu to begin with). This allows the tests to run in parallel and has significantly increased the speed. 

Wrt the other points, there's a lot of refactoring to be done and things to make the tests easier to update, but I figure that can be done in a follow up PR so the changes are easier to review (the diff is already fairly large). I plan on update the CONTRIBUTING.md and adding a Makefile to make running tests locally straightforward imminently.

.github/workflows/test.yml
```diff
@@ -0,0 +1,106 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "windows-2022"]
+        type: ["unit", "integration", "tool"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v3
```
elopez, line 31:
```suggestion
        uses: actions/setup-python@v4
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,106 @@
...
+            # Setup Ganache for slither-read-storage tests.
+            npm install --global ganache
+          elif [ ${{ matrix.type }} = "integration" ]; then
+            # Setup Hardhat for compilation tests.
+            pushd tests/e2e/compilation/test_data/test_node_modules/ || exit
+            npm install hardhat
+            popd || exit
+          fi
+
+      - name: ${{ matrix.type }}
```
elopez, line 62:
```suggestion
      - name: Run ${{ matrix.type }} tests
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          if [ ${{ matrix.os }} = "ubuntu-latest" ]; then
+            TEST_ARGS="--cov=slither --cov-append"
+          elif [ ${{ matrix.os }} = "windows-2022" ]; then
+            TEST_ARGS=""
+          fi
+          bash "./.github/scripts/${TEST_TYPE}_test_runner.sh" $TEST_ARGS
+
+
+      - uses: ./.github/actions/upload-coverage
```
elopez, line 75:
```suggestion
      - name: Upload coverage
        uses: ./.github/actions/upload-coverage
```

elopez (commented):
From a quick look (15k modified files 😮), this is looking good 👍  I left a few nits below

0xalpharush:
We can probably squash and merge this if it's okay for me to follow up with improvements. I've addressed the main concern of parallelizing the runs while preserving the ability to aggregate coverage.

montyly:
Note: I will squash the PR when merging, and I agree that we can work on the other parts in different PRs

0xalpharush:
> Should slither/.github/scripts/ be moved to slither/scripts ? I am assuming that we might re-used them for the makefile?
I was only planning on using them in the CI, but if it is necessary to use them in the makefile, I will move them.

Issue #1648, opened 2023-02-04
function signature string does not contain data location for dynamic types


0xalpharush:
Strings/bytes etc are indicated as memory/calldata in the function signature
https://github.com/crytic/slither/blob/4c976d5af56219eeef079e03a35009af3e03644d/slither/core/declarations/function.py#L996

PR #1804, opened 2023-03-27
Include data location in function signatures

webthethird:
Include data location in function signature parameters and returns for dynamic types. 
Modifies `Function.signature` and `Function.solidity_signature`.
Addresses #1648


webthethird:
Looks like this wouldn't be entirely accurate, since Solidity omits the data location when computing the function selector from its signature. Better to handle this elsewhere for interface code generation.

PR #1808, opened 2023-03-28
run tests in parallel locally with makefile

0xalpharush:
This PR adds a Makefile that creates a developer env (`make dev`), run linting *validation* (`make lint`), reformats and *modifies* the code (`make reformat`), runs tests (`make test`). The guidance in CONTRIBUTING.md has been updated to reflect these and the modifications to the test directory structure. In addition, the usage of solc-select was modified to use a fixture (`solc_binary_path`) which takes a lock on the binary while the test runs, allowing the tests to be run in parallel


tests/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+@pytest.fixture(scope="session")
+def solc_versions_installed():
+    """List of solc versions available in the test environment."""
+    return []
+
+@pytest.fixture(scope="session", autouse=True)
+def register_solc_versions_installed(solc_versions_installed):
+    solc_versions_installed.extend(solc_select.installed_versions())
+
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
0xalpharush: Doesn't look like it works well for windows https://github.com/crytic/slither/actions/runs/4545625995/jobs/8013219900?pr=1808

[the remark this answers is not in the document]
elopez: I think we could simplify this quite a bit. We can rely on the filesystem instead of carrying a list. What do you think of something like the following instead?

```python
@pytest.fixture(scope="session")
def solc_binary_path():
    def inner(version):
        lock = FileLock(f"{version}.lock")
        with lock.acquire(timeout=60):
            if not solc_select.artifact_path(version).exists()
                print("Installing solc version", version)
                solc_select.install_artifacts([version])
        return solc_select.artifact_path(version)
    return inner
```
and then on the tests you can do

```python
def test_contract_name_collision(solc_binary_path) -> None:
    solc_path = solc_binary_path("0.8.0")
    ...
```

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+import pytest
+from filelock import FileLock
+from solc_select import solc_select
+from slither import Slither
+
+
+@pytest.fixture()
+def solc_binary_path():
+    def inner(version):
+        lock = FileLock(f"{version}.lock", timeout=60)
```
0xalpharush, line 14: This file sticks around which is a little annoying

elopez: Maybe you can do something like the following to have a global temp dir across all tests, and then create filelocks there: https://github.com/pytest-dev/pytest/issues/1402#issuecomment-186299177

0xalpharush: Awesome, thanks! I think this is ready to go now that I've made this change https://github.com/crytic/slither/pull/1808/commits/6451285f8f62d7fb24d79d52fa47e7039a2f2b0c

CONTRIBUTING.md
```diff
@@ -39,50 +44,72 @@ Below is a rough outline of slither's design:
 A code walkthrough is available [here](https://www.youtube.com/watch?v=EUl3UlYSluU).
 
 ## Development Environment
+
 Instructions for installing a development version of Slither can be found in our [wiki](https://github.com/crytic/slither/wiki/Developer-installation).
 
-To run the unit tests, you need to clone this repository and run `pip install ".[dev]"`.
+To run the unit tests, you need to clone this repository and run `make test`. Run a specific test with `make test TESTS=$test_name`.
```
montyly, line 50: We should list the test_name available I think

CONTRIBUTING.md
```diff
@@ -39,50 +44,72 @@ Below is a rough outline of slither's design:
...
-### Detectors tests
+### Testing
+
+Slither's test suite is divided into three categories end-to-end (`tests/e2e`), unit (`tests/unit`), and tools (`tests/tools/`).
+
+How do I know what kind of test(s) to write?
+
+- End-to-end: functionality that requires invoking `Slither` and inspecting some output such as printers and detectors.
+- Unit: additions and modifications to objects should be accompanied by a unit test that defines the expected behavior. Aim to write functions in as pure a way as possible such that they are easier to test.
+- Tools: tools built on top of Slither (`slither/tools) but not apart of its core functionality
```
montyly, line 76: Missing " ` "

tests/unit/core/test_source_mapping.py
```diff
@@ -102,26 +100,16 @@ def _test_references_user_defined_aliases():
...
+    slither = Slither(file, solc=solc_path)
 
     contracts = slither.compilation_units[0].contracts
     a = contracts[0] if contracts[0].is_interface else contracts[1]
     assert len(a.references) == 2
     lines = _sort_references_lines(a.references)
     assert lines == [12, 18]
-
-
-def test_references():
```
montyly, line 120: was this removed?

0xalpharush: No, the test was calling two nested tests and I split them to be their own test. https://github.com/crytic/slither/pull/1808/files/d1b9e8288900d6b7d75025b3f589adec0b68a045#diff-fee079d1aa52bfd03e4808f11f095beadfb074ecb24718f690004bb271bd4754R80-R103

CONTRIBUTING.md
0xalpharush: https://github.com/crytic/slither/pull/1808/commits/0a6197560156e6b926eb480bc8e943e56b8da76e

PR #1818, opened 2023-03-31
Add failing test of `utils.upgradeability.get_implementation_var(proxy)` 

webthethird:
The `get_implementation_var(proxy)` function in `slither.utils.upgradeability` is failing to find all the dependencies of the local `implementation` address variable in the fallback of [TransparentUpgradeableProxy.sol](https://github.com/webthethird/slither/blob/dev-upgradeability-utils/tests/unit/utils/test_data/upgradeability_util/src/TransparentUpgradeableProxy.sol).

[Line 412 of upgradeability.py](https://github.com/webthethird/slither/blob/dev-upgradeability-utils/slither/utils/upgradeability.py#L412) tries to find the data source of the `implementation` LocalVariable, which is declared in the signature of `_delegate(address implementation)` of the base Proxy contract, using `data_dependency.get_dependencies`, but only a TemporaryVariable is returned, representing the return value of `_implementation()` in the expression `_delegate(_implementation())`. The data dependency tool should recognize that `_implementation()` is overridden in the `UpgradeableProxy` contract, which returns a local variable called `impl`, which in turn depends on the local variable `slot` which equals the state variable `_IMPLEMENTATION_SLOT`.

This PR adds a failing condition to `tests.unit.utils.test_upgradeability_util` for reproducing the issue. By the time this PR is ready to merge, the failure should be resolved such that the test passes.


0xalpharush:
Since we identified the bug https://github.com/crytic/slither/issues/1819, can this be closed or it wasn't just to demonstrate a bug?

webthethird:
> Since we identified the bug #1819, can this be closed or it wasn't just to demonstrate a bug?

Ideally this would actually become part of the test, without needing to modify the Solidity code to avoid using a named return variable. It's sort of an unhappy path at the moment, even though it is a relatively common example -- that version of the TransparentUpgradeableProxy contract is probably used all over the place.

PR #1843, opened 2023-04-17
slither-read-storage native POA support

webthethird:
Adds a new `RpcInfo` class that holds `self.rpc`, `self.web3` and `self.block`, which are removed from `SlitherReadStorage`. `SlitherReadStorage` gets new `self._rpc_info: Optional[RpcInfo]` which is required for using Web3.

In `RpcInfo`, automatically detect if the network is POA when setting `self._block`. `web3.eth.get_block` raises an ExtraDataLengthError when using an RPC for a POA network, so if we catch one of those after trying to set `self._block`, we know it is a POA network and therefore inject the appropriate middleware.


slither/tools/read_storage/read_storage.py
```diff
@@ -73,6 +75,26 @@ def log(self) -> str:
...
+        if not self._block:
+            try:
+                self._block = self.web3.eth.get_block("latest")["number"]
+            except ExtraDataLengthError:
+                self.web3.middleware_onion.inject(geth_poa_middleware, layer=0)
+                self._block = self.web3.eth.get_block("latest")["number"]
+        return self._block
+
+    @block.setter
+    def block(self, block) -> None:
```
0xalpharush, line 90:
```suggestion
    def block(self, block: Union[str, int]) -> None:
```

slither/tools/read_storage/__main__.py
```diff
@@ -143,6 +138,11 @@ def main() -> None:
 
         srs.rpc = args.rpc_url
 
+    try:
+        srs.block = int(args.block)
+    except ValueError:
+        srs.block = str(args.block or "latest")
```
0xalpharush, line 144: Should this call the setter with "latest" as the parameter?

slither/tools/read_storage/read_storage.py
```diff
@@ -73,6 +75,26 @@ def log(self) -> str:
...
 
+    @property
+    def block(self) -> int:
+        """If the RPC is for a POA network, the first call to get_block fails, so we inject geth_poa_middleware"""
+        if not self._block:
+            try:
+                self._block = self.web3.eth.get_block("latest")["number"]
+            except ExtraDataLengthError:
+                self.web3.middleware_onion.inject(geth_poa_middleware, layer=0)
+                self._block = self.web3.eth.get_block("latest")["number"]
```
0xalpharush, line 86: Since we're calling the setter, should this return None and be `Optional[int]` if it hasn't been set?

webthethird: I was just thinking about this. We might want to default to "latest", since that was the default value before we added the property getter and setter.

slither/tools/read_storage/__main__.py
```diff
@@ -128,21 +128,20 @@ def main() -> None:
...
-    except ValueError:
-        srs.block = str(args.block or "latest")
-
     if args.rpc_url:
+        try:
+            block = int(args.block)
+        except ValueError:
+            block = "latest"
+        rpc_info = RpcInfo(args.rpc_url, block)
+        srs = SlitherReadStorage(contracts, args.max_depth, rpc_info)
```
0xalpharush, line 137: I think we could something like the following to avoid instantiating `SlitherReadStorage` twice
 ```
rpc_info = None
if args.rpc_url:
        try:
            block = int(args.block)
        except ValueError:
            block = "latest"
        rpc_info = RpcInfo(args.rpc_url, block)
srs  = SlitherReadStorage(contracts, args.max_depth, rpc_info)

slither/tools/read_storage/read_storage.py
```diff
@@ -42,18 +44,36 @@ class SlitherReadStorageException(Exception):
     pass
 
 
+class RpcInfo:
+    def __init__(self, rpc_url: str, block: Union[int, str] = "latest") -> None:
+        assert isinstance(block, int) or block.isnumeric() or block == "latest"
```
0xalpharush, line 49: What is `block.isnumeric()`

slither/tools/read_storage/read_storage.py
```diff
@@ -42,18 +44,36 @@ class SlitherReadStorageException(Exception):
...
+        self.rpc: str = rpc_url
+        self._web3: Web3 = Web3(Web3.HTTPProvider(self.rpc))
+        try:
```
0xalpharush, line 52: Can you add the comment about POA back please?

slither/tools/read_storage/read_storage.py
webthethird: Since `block` can be a string, I figured I'd play it safe in case `block == "12345678"` or something like that. Not necessary?

0xalpharush: I'm not sure. It's probably fine. I think there's also arguments besides "latest" like "finalized"

webthethird: Ahh that's a good point, I'll make sure "latest", "earliest", "pending", "safe" and "finalized" are all acceptable

slither/tools/read_storage/read_storage.py
```diff
@@ -42,18 +44,41 @@ class SlitherReadStorageException(Exception):
     pass
 
 
+class RpcInfo:
+    def __init__(self, rpc_url: str, block: Union[int, str] = "latest") -> None:
+        assert (
+            isinstance(block, int)
+            or block.isnumeric()
+            or block in ["latest", "earliest", "pending", "safe", "finalized"]
```
0xalpharush, line 52: Since we cast block to an `int` if possible perhaps it'd be best to reject `"1234"` (remove `isnumeric`) and create and Enum for the valid `str` args

slither/tools/read_storage/__main__.py
```diff
@@ -126,22 +126,21 @@ def main() -> None:
...
-        # Default to implementation address unless a storage address is given.
-        if not args.storage_address:
-            args.storage_address = address
-        srs.storage_address = args.storage_address
-
-        srs.rpc = args.rpc_url
+        try:
+            block = int(args.block)
+        except ValueError:
+            block = "latest"
```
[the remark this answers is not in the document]
0xalpharush: See also https://github.com/crytic/slither/pull/1843/files#r1185253313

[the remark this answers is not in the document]
webthethird: `block` is no longer a member of `SlitherReadStorage`, since it was moved to the new `RpcInfo` class. So how about this?
```python
    if args.rpc_url:
        if args.block in ["latest", "earliest", "pending", "safe", "finalized"]:
            block = args.block
        else:
            try:
                block = int(args.block)
            except ValueError:
                block = "latest"
        rpc_info = RpcInfo(args.rpc_url, block)
```

slither/tools/read_storage/read_storage.py
webthethird: Do you know of an example elsewhere in the codebase that uses an enum like this?

slither/tools/read_storage/__main__.py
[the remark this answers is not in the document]
0xalpharush:
```
    if args.rpc_url:
            try:
                block = int(args.block)
            except ValueError:
                block = BlockTag(args.block) or "latest"
        rpc_info = RpcInfo(args.rpc_url, block)
```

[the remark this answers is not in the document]
webthethird: Done!

slither/tools/read_storage/read_storage.py
```diff
@@ -42,18 +45,55 @@ class SlitherReadStorageException(Exception):
...
+class BlockTag(Enum, metaclass=MetaEnum):
+    LATEST = "latest"
+    EARLIEST = "earliest"
+    PENDING = "pending"
+    SAFE = "safe"
+    FINALIZED = "finalized"
+
+
+class RpcInfo:
+    def __init__(self, rpc_url: str, block: Union[int, str] = "latest") -> None:
```
0xalpharush, line 67:
```suggestion
    def __init__(self, rpc_url: str, block: Union[int, BlockTag] = "latest") -> None:
```

slither/tools/read_storage/read_storage.py
```diff
@@ -42,18 +45,55 @@ class SlitherReadStorageException(Exception):
...
+        assert isinstance(block, int) or block in BlockTag
```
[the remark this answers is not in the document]
webthethird: I just tried that, but got an assertion error. 

I got the MetaEnum solution from [here](https://stackoverflow.com/questions/63335753/how-to-check-if-string-exists-in-enum-of-strings).

slither/tools/read_storage/__main__.py
```diff
@@ -126,22 +126,21 @@ def main() -> None:
...
-        # Default to implementation address unless a storage address is given.
-        if not args.storage_address:
-            args.storage_address = address
-        srs.storage_address = args.storage_address
-
-        srs.rpc = args.rpc_url
+        try:
+            block = int(args.block)
+        except ValueError:
+            block = BlockTag(args.block) or "latest"
```
0xalpharush, line 134:
```suggestion
            block = BlockTag(args.block)
```
"latest" is the default arg and not being converted to `BlockTag`

webthethird:
I'm running into an error in `web3.eth.get_block` when passing in an enum element, even though its value is a string.

```
    def test_read_storage(web3, ganache) -> None:
        assert web3.is_connected()
        bin_path = Path(TEST_DATA_DIR, "StorageLayout.bin").as_posix()
        abi_path = Path(TEST_DATA_DIR, "StorageLayout.abi").as_posix()
        bytecode = get_source_file(bin_path)
        abi = get_source_file(abi_path)
        contract = deploy_contract(web3, ganache, bytecode, abi)
        contract.functions.store().transact({"from": ganache.eth_address})
        address = contract.address
    
        sl = Slither(Path(TEST_DATA_DIR, "storage_layout-0.8.10.sol").as_posix())
        contracts = sl.contracts
    
>       rpc_info: RpcInfo = RpcInfo(ganache.provider, BlockTag.LATEST)

test_read_storage.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../slither/tools/read_storage/read_storage.py:73: in __init__
    self._block: int = self.web3.eth.get_block(block)["number"]
../../../../../.local/lib/python3.10/site-packages/web3/eth/eth.py:389: in get_block
    return self._get_block(block_identifier, full_transactions)
../../../../../.local/lib/python3.10/site-packages/web3/module.py:58: in caller
    (method_str, params), response_formatters = method.process_params(
../../../../../.local/lib/python3.10/site-packages/web3/method.py:214: in process_params
    self.json_rpc_method = self.method_choice_depends_on_args(value=params[0])
cytoolz/functoolz.pyx:253: in cytoolz.functoolz.curry.__call__
    ???
cytoolz/functoolz.pyx:249: in cytoolz.functoolz.curry.__call__
    ???
../../../../../.local/lib/python3.10/site-packages/web3/_utils/blocks.py:62: in select_method_for_block_identifier
    if is_predefined_block_number(value):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = <BlockTag.LATEST: 'latest'>

    def is_predefined_block_number(value: Any) -> bool:
        if is_text(value):
            value_text = value
        elif is_bytes(value):
            # `value` could either be random bytes or the utf-8 encoding of
            # one of the words in: {"latest", "pending", "earliest", "safe", "finalized"}
            # We cannot decode the bytes as utf8, because random bytes likely won't be
            # valid. So we speculatively decode as 'latin-1', which cannot fail.
            value_text = value.decode("latin-1")
        elif is_integer(value):
            return False
        else:
>           raise TypeError(f"unrecognized block reference: {value!r}")
E           TypeError: unrecognized block reference: <BlockTag.LATEST: 'latest'>
```
I'm thinking it might be better to use the pre-existing `web3.types.BlockParams` instead of the new BlockTag enum:
`BlockParams = Literal["latest", "earliest", "pending", "safe", "finalized"]`

webthethird:
test_read_storage is passing on my machine, but for some reason the `assert web3.is_connected()` is failing here
```
=========================== short test summary info ===========================
FAILED tests/tools/read-storage/test_read_storage.py::test_read_storage - assert False
 +  where False = <bound method Web3.is_connected of <web3.main.Web3 object at 0x0000020F37819AF0>>()
 +    where <bound method Web3.is_connected of <web3.main.Web3 object at 0x0000020F37819AF0>> = <web3.main.Web3 object at 0x0000020F37819AF0>.is_connected
 ```

**Comment by webthethird** [comment]
The more I look at it, the more it seems like introducing the `BlockTag` enum just over-complicates the code. 
Looking at web3._utils, this is similar to what we were doing before introducing that enum:
```python
def parse_block_identifier(
    w3: "Web3", block_identifier: BlockIdentifier
) -> BlockIdentifier:
    if block_identifier is None:
        return w3.eth.default_block
    if isinstance(block_identifier, int):
        return parse_block_identifier_int(w3, block_identifier)
    elif block_identifier in ["latest", "earliest", "pending", "safe", "finalized"]:
        return block_identifier
    elif isinstance(block_identifier, bytes) or is_hex_encoded_block_hash(
        block_identifier
    ):
        return w3.eth.get_block(block_identifier)["number"]
    else:
        raise BlockNumberOutofRange
```
I am leaning toward going back to this and removing `BlockTag`:
```python
class RpcInfo:
    def __init__(self, rpc_url: str, block: BlockIdentifier = "latest") -> None:
        assert isinstance(block, int) or block in ["latest", "earliest", "pending", "safe", "finalized"]
        ...
```

webthethird:
>I am leaning toward going back to this and removing BlockTag

@0xalpharush I went ahead and changed it back

slither/tools/read_storage/__main__.py
```diff
@@ -126,22 +126,22 @@ def main() -> None:
...
-        if not args.storage_address:
-            args.storage_address = address
-        srs.storage_address = args.storage_address
-
-        srs.rpc = args.rpc_url
+        try:
+            block = int(args.block)
+        except ValueError:
+            valid = ["latest", "earliest", "pending", "safe", "finalized"]
+            block = next((v for v in valid if v == args.block), "latest")
```
0xalpharush, line 135:
```suggestion
            block = args.block if args.block in valid else "latest"
```

0xalpharush: We already default to latest for `args.block` so maybe this should throw an error as the user passed in an invalid parameter explicitly

webthethird: > We already default to latest for `args.block` so maybe this should throw an error as the user passed in an invalid parameter explicitly

Something like this?
```python
    if args.rpc_url:
        try:
            valid = ["latest", "earliest", "pending", "safe", "finalized"]
            block = args.block if args.block in valid else int(args.block)
        except ValueError as err:
            raise SlitherError(str(err)) from err
        rpc_info = RpcInfo(args.rpc_url, block)
```

0xalpharush: I would leave off ```
except ValueError as err:
            raise SlitherError(str(err)) from err
```

webthethird: Ok, then I'll drop the try/except and let it fail with the ValueError

0xalpharush (approved):

0xalpharush:
I think we can squash and merge this

PR #1848, opened 2023-04-18
Bug Fix: Contract obj is_fully_implemented

### LINKED ISSUE #1847: [Bug]: is_fully_implemented property of Contract objects always False
### Describe the issue:

All `Contract` objects' `is_fully_implemented` property seems to never get updated, and is always False (since initialisation).
In previous versions of Slither, the equivalent property was `fullyImplemented`, and was working well at least up till v0.6.8

Here is an example with 2 contracts, Loan (not fully implemented) and LoanDirectory (fully implemented).

https://etherscan.io/address/0xc477042db387DD59C038Ca4b829a07964b674347#code

### Version:

0.9.3

### Relevant log output:

_No response_

DarrenChangJR:
Previously was not updating the is_fully_implemented property due to being incorrectly included in the `if` statement.
Fixes #1847

### DIFF
--- a/slither/solc_parsing/declarations/contract.py
+++ b/slither/solc_parsing/declarations/contract.py
@@ -173,7 +173,6 @@ class ContractSolc(CallerContextExpression):
 
         self._contract.is_fully_implemented = attributes["fullyImplemented"]
         self._linearized_base_contracts = attributes["linearizedBaseContracts"]
-        # self._contract.fullyImplemented = attributes["fullyImplemented"]
 
         # Parse base contract information
         self._parse_base_contract_info()
--- a/slither/solc_parsing/declarations/contract.py
+++ b/slither/solc_parsing/declarations/contract.py
@@ -170,8 +170,8 @@ class ContractSolc(CallerContextExpression):
             elif attributes["contractKind"] == "library":
                 self._contract.is_library = True
             self._contract.contract_kind = attributes["contractKind"]
-            self._contract.is_fully_implemented = attributes["fullyImplemented"]
 
+        self._contract.is_fully_implemented = attributes["fullyImplemented"]
         self._linearized_base_contracts = attributes["linearizedBaseContracts"]
         # self._contract.fullyImplemented = attributes["fullyImplemented"]
 
--- a/tests/unit/core/test_contract_declaration.py
+++ b/tests/unit/core/test_contract_declaration.py
@@ -32,6 +32,7 @@ def test_concrete_contract(solc_binary_path) -> None:
     slither = Slither(Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol").as_posix(), solc=solc_path)
     assert slither.contracts[0].is_fully_implemented
 
+    solc_path = solc_binary_path("0.5.0")
     slither = Slither(
         Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol").as_posix(),
         solc_force_legacy_json=True,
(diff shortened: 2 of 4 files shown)


0xalpharush:
Thanks for the PR! Would you mind adding a regression test case that expect `is_fully_implemented` to return True for a non-abstract contract here please https://github.com/crytic/slither/blob/3fb564fbc3c9fd6fb144f6a6e02bef3733d3c8a0/tests/unit/core/test_contract_declaration.py#L11-L27

slither/solc_parsing/declarations/contract.py
```diff
@@ -170,8 +170,8 @@ def _parse_contract_info(self) -> None:
             elif attributes["contractKind"] == "library":
                 self._contract.is_library = True
             self._contract.contract_kind = attributes["contractKind"]
-            self._contract.is_fully_implemented = attributes["fullyImplemented"]
 
+        self._contract.is_fully_implemented = attributes["fullyImplemented"]
         self._linearized_base_contracts = attributes["linearizedBaseContracts"]
         # self._contract.fullyImplemented = attributes["fullyImplemented"]
```
0xalpharush, line 176:
```suggestion
```

tests/unit/core/test_contract_declaration.py
```diff
@@ -27,6 +27,25 @@ def test_abstract_contract(solc_binary_path) -> None:
...
+    assert slither.contracts[0].is_fully_implemented
+
+    slither = Slither(
+        Path(CONTRACT_DECL_TEST_ROOT, "implicit_abstract.sol").as_posix(),
+        solc_force_legacy_json=True,
+        solc=solc_path,
+    )
+    assert slither.contracts[0].is_fully_implemented
+
+
```
0xalpharush, line 48: It seems like a concrete contract should be tested here instead of abstract

DarrenChangJR: Sorry, this is my first open source pull request. Am I supposed to edit the checks or do anything else?

0xalpharush: No worries. Right now, we're asserting that this [abstract contract](https://github.com/crytic/slither/blob/dev/tests/unit/core/test_data/contract_declaration/abstract.sol) (`Path(CONTRACT_DECL_TEST_ROOT, "abstract.sol")`) is_fully_implemented. However, it will return `False` as it's abstract.

Since we want `is_fully_implemented` to return `True`, we could add a new solidity file [here](https://github.com/crytic/slither/blob/dev/tests/unit/core/test_data/contract_declaration/) called `concrete.sol` and then use `Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol")`. Note, this contract should implement all of its functions and not use the `abstract` keyword.

Let me know if I can further help as we welcome contributions and appreciate it!

tests/unit/core/test_contract_declaration.py
```diff
@@ -27,6 +27,19 @@ def test_abstract_contract(solc_binary_path) -> None:
...
+    solc_path = solc_binary_path("0.8.0")
+    slither = Slither(Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol").as_posix(), solc=solc_path)
+    assert slither.contracts[0].is_fully_implemented
+
+    slither = Slither(
+        Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol").as_posix(),
+        solc_force_legacy_json=True,
+        solc=solc_path,
+    )
+    assert slither.contracts[0].is_fully_implemented
```
0xalpharush, line 40:
```suggestion
    solc_path = solc_binary_path("0.5.0")
    slither = Slither(
        Path(CONTRACT_DECL_TEST_ROOT, "concrete.sol").as_posix(),
        solc_force_legacy_json=True,
        solc=solc_path,
    )
    assert slither.contracts[0].is_fully_implemented
```

0xalpharush (approved):

montyly:
Great fix, thanks @DarrenChangJR

PR #1856, opened 2023-04-22
Remove assertion in unary operation

smonicas:
Remove an assertion that is not needed and prevent valid use cases such as negating an element of an array.
```
pragma solidity 0.8.19;

contract D {
 function a(int256[] memory data) public pure {
    -data[0];
 }
}
```


montyly:
can we add this example as a test?

0xalpharush (approved):

PR #1861, opened 2023-04-24
Improve tuple analysis for unused-return detector

smonicas:
Now if a function call returns multiple values all must be used. Before it was enough to use only 1 value and this detector wouldn't detect the other unused.
```
        // Detected as unused return
        (uint e,) = t.g();
```


0xalpharush:
There's not really a great way to fix this detector because if the variable is not needed, adding it will create an unused local.

tests/e2e/detectors/snapshots/detectors__detector_UnusedReturnValues_0_6_11_unused_return_sol__0.txt
```diff
@@ -1,4 +1,8 @@
...
+User.test(Target) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#18-37) ignores return value by t.f() (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#19)
 
-User.test(Target) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#17-29) ignores return value by t.f() (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#18)
+User.test(Target) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#18-37) ignores return value by a.add(0) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#23)
+
+User.test(Target) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#18-37) ignores return value by t.g() (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#31)
+
+User.test(Target) (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#18-37) ignores return value by (e) = t.g() (tests/e2e/detectors/test_data/unused-return/0.6.11/unused_return.sol#36)
```
0xalpharush, line 7: It would be good if we could get this to display as `(e, ) = t.g()` so it's clear what is being ignored

montyly: I agree, but this should be done in a separate PR

Issue #1450, opened 2022-11-03
implicit, named return does not have RETURN node


0xalpharush:
```
function test() external view returns (uint named) {
     named = 1;
     // name in return parameter list indicates implicit return
}
```
The AST does not contain a `ReturnStatement` so we never create a `RETURN` node. We can enhance slithIR with this implicit return and change the node's type.

webthethird:
> We can enhance slithIR with this implicit return and change the node's type.

You mean change the type of the node containing the AssignmentOperation from `EXPRESSION`  to `RETURN`, right? 
Would this be done after parsing the function's CFG? And is it necessary that the node with the AssignmentOperation have no children?

0xalpharush:
closed by https://github.com/crytic/slither/pull/2326

Issue #1719, opened 2023-03-07
data dependency of named return SSA variables


Troublor:
### Describe the desired feature

Given a simple function with named return variable like this:
```
contract A {
    uint a = 1;
    function foo(uint x) public returns (uint y) {
        if (x > 0) 
            y = x;
        else
            y = x + a;
    }
}
```
In the SSA SlithIR, the return variable (i.e., `returns_ssa` field of the function object) is `y_0`, and the variable being assigned in the statement `y=x` and `y=x+a` is `y_1` and `y_2`. 
There is no assignment for the return variable `y_0`, resulting that there is no data dependency of the return variable `y_0`. 
This poses a difficulty in analyzing the data dependencies of function returns using SSA. 

One possible enhancement is do a control flow analysis and find last assignment to variable `y` in all possible paths and add a PHI operation at the end of the function, e.g., `y_0 = phi[y_1, y_2]`.

0xalpharush:
I believe this is fixed in https://github.com/crytic/slither/pull/1102 but it requires more testing. Something helpful to know: there's also a limitation that a return node is not created for `y` https://github.com/crytic/slither/issues/1450.

bsamuels453:
#1715 related

0xalpharush:
closed by https://github.com/crytic/slither/pull/2326

PR #1850, opened 2023-04-20
Preserve the order of sons when splitting ternary node

Troublor:
Fix #1846 

The order of sons of node matters. 
Node with `NodeType.IF` relies on the order of the sons to determine which son is the true branch and which is the false branch. 

However, the `_split_ternary_node` function during AST parsing does not preserve the son order of the father node. 
https://github.com/crytic/slither/blob/0ec487460690482c72cacdea6705e2c51bb3981e/slither/solc_parsing/declarations/function.py#L1397-L1400
When the father node is a node with multiple sons, and the index of the son to remove is not `-1`, the order of sones in the father node will be changed after `remove_son` and `add_son`. 
As a result, the issue in #1846 occurs. 

This PR adds a new utility method, `replace_son`, in `Node` class to preserve the order of sones when replacing a son. 
So that #1846 can be fixed. 

Two bug-revealing test cases are also added.

PR #1880, opened 2023-05-02
Create return nodes for named implicit returns

webthethird:
Resolves issues #1450 and #1719, and includes the previously failing test from #1818, which now passes.

Creates an artificial return node iff a function has a named return variable declared in its signature. Finds all leaf nodes in the CFG which are not return nodes, and links them to the artificial return node. The return node is given an Identifier expression with a `referencedDeclaration` pointing to the variable declaration in the signature.

Once this return node is in place and correctly linked, subsequent analyses resolve the data dependency issue for named variables without any further changes, i.e., to Slither's IR generation. The goal being to implement the simplest solution to that problem.

### DIFF
--- a/.github/DISCUSSION_TEMPLATE/installation.yml
+++ b/.github/DISCUSSION_TEMPLATE/installation.yml
@@ -2,12 +2,6 @@
 body: 
   - 
     attributes: 
-      value: |
-          Please check the issues tab to avoid duplicates.
-          Thanks for taking the time to fill out this bug report!
-    type: markdown
-  - 
-    attributes: 
       label: "What operating system are you using?"
     id: os
     type: textarea
@@ -31,14 +25,6 @@ body:
         - "Yes"
         - "No"
         - "Not sure"
-  - type: dropdown
-    id: solc
-    attributes:
-      label: Do you have solc-select installed?
-      multiple: true
-      options:
-        - "Yes"
-        - "No"
   - 
     attributes: 
       description: |
@@ -47,9 +33,6 @@ body:
       render: shell
       label: "Output of running `slither-doctor .`:"
     id: logs
-    type: textarea
-description: "Get help troubleshooting slither installation"
 labels: 
   - installation-help
-name: "Trouble with Installing Slither"
 title: "[Installation-Help]: "
--- a/.github/ISSUE_TEMPLATE/false_negative.yml
+++ b/.github/ISSUE_TEMPLATE/false_negative.yml
@@ -57,5 +57,5 @@ body:
 description: "Slither missed a bug it should find."
 labels: 
   - false-negative
-name: False Negative"
+name: False Negative
 title: "[False Negative]: "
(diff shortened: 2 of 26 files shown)


0xalpharush:
The printers are failing because of an assertion failure:
```
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/slither/solc_parsing/declarations/function.py", line 321, in analyze_content
    self._fix_implicit_return(body)
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/slither/solc_parsing/declarations/function.py", line 1301, in _fix_implicit_return
    return_node.add_unparsed_expression(
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/slither/solc_parsing/cfg/node.py", line 30, in add_unparsed_expression
    assert self._unparsed_expression is None
AssertionError

Printer failed
yul-0.4.11.sol-0.5.17-compact.zip
```

I would check locally that the updates are correct, but you can run `python tests/e2e/solc_parsing/test_ast_parsing.py --overwrite` to update the tests.

webthethird:
Since we decided on closing #1886 in favor of leaving the empty functions flagged as implemented and just adding an implicit `RETURN` node after the `ENTRY_POINT`, we will need to update the expected result files for the CFG printer test to account for the new nodes.

Also, @montyly you mentioned:
>[However if we do so, we should add the information that the node was not explicit in Solidity, as this information can be useful for further analyses](https://github.com/crytic/slither/pull/1886#issuecomment-1547744033)

Do you have any thoughts on what the best way to capture this information? I imagine we wouldn't want to create a new `NodeType`, since there must be lots of places where `if(node.type == NodeType.RETURN)` is used, and we want implicit returns to still register as returns.

webthethird:
@0xalpharush could you take a look at [this test failure](https://github.com/crytic/slither/actions/runs/4984745250/jobs/8923482979?pr=1880#step:7:151) and let me know if you can tell why it's failing? I think practically all of the assertion failures in the 2 failing checks below are due to the new return nodes, and can be resolved by updating the expected CFGs. But I can't think of any reason why a reentrancy detector would have failed.

webthethird:
It may have something to do with the fact that `DAO.splitDAO` has a named return variable but never uses it.

webthethird:
I'm having a hard time figuring out how to resolve the failing tests for `conditional-all.sol`. It seems like the issue should be resolved by #1850.
I did a diff of the expected and actual results:
```
33,34c33,34
<   14->16[label="True"];
<   14->26[label="False"];
---
>   14->26[label="True"];
>   14->16[label="False"];
```
It seems the links are getting flipped, even though the order of the sons of node 14 should be preserved now.

webthethird:
All checks are passing, including the `implicit-returns` parser tests I added. This one is ready for review!

0xalpharush:
replaced by https://github.com/crytic/slither/pull/2326

PR #1881, opened 2023-05-02
chore: run black

devtooligan:


devtooligan:
i was using a more recent version of black. closing this

PR #1882, opened 2023-05-02
feat: loc printer

devtooligan:
- This pr adds a new printer: `loc` for metrics about lines of code. 
```bash

INFO:Printers:Lines of Code
+------+-----+-----+------+
|      | src | dep | test |
+------+-----+-----+------+
| loc  |  21 |  0  |  0   |
| sloc |  11 |  0  |  0   |
| cloc |  5  |  0  |  0   |
+------+-----+-----+------+
```

- The Human summary printer was also updated.  Previously, the Human summary only reported total lines of code (LOC) but now it only reports SLOC. Also, it only displays information about Test and Dependency files if those type of files were found.

Before:
```bash
INFO:Printers:
Compiled with solc
Number of lines: 21 (+ 0 in dependencies, + 0 in tests)
```

After
```bash
INFO:Printers:
Compiled with solc
Source lines of code (SLOC) in source files: 11
```

- A couple of utility files were added to easily convert `dict`s to `MyPrettyTable`


montyly:
For visibility, I made the following changes:
- Move LoC features into utils.loc as they are now used by two printers
- Remove the usage of dict and create LoC/LoCInfo dataclass to reduce complexity
- Add to_pretty_table in the LoC object to reduce complexity
- Use proper type (remove PEP 585 for python 3.8 support)

0xalpharush (approved):

Issue #1893, opened 2023-05-05
Create slither-interface


montyly:
See https://github.com/crytic/slither/pull/1802#issuecomment-1525717360

0xGusMcCrae:
I'll take a look at this over the weekend and see if I can get it done

montyly:
That's great @0xGusMcCrae, let me know if you need guidance or help. 

Some notes:
- We can build `slither-interface` following the tool's demo structure: https://github.com/crytic/slither/tree/master/slither/tools/demo
- Most the logic might already be implemented in `slither/utils/code_generation.py`
- We recently made a significant refactor of our testing strategies. Be sure to work from the `dev` branch, and read https://github.com/crytic/slither/blob/dev/CONTRIBUTING.md

0xGusMcCrae:
I've got the basic functionality down.  I just need to write some tests and I should have a PR ready tomorrow.

For the output, do we want to print the interface in the console and allow users to write to file manually? 

i.e. `slither-interface WETH9 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 >> IWeth9.sol`

Or should it write to file by default? Maybe a flag so there's an option between the two?

montyly:
Awesome. I think a file is probably more appropriate. 

Most of our tools use `crytic-export/name_tool/contract`, maybe we can do `crytic-export/interfaces/ContractA.sol`, `crytic-export/interfaces/ContractB.sol`, ... by default?

PR #1898, opened 2023-05-08
add tool slither-interface

0xGusMcCrae:
As per issue #1893 

This tool outputs an interface file for a given contract input. The input can be a local source file or a deployment address if the contract is verified on etherscan.

Usage:
`slither-interface <ContractName> <Source File or Address>`

For example:
`slither-interface WETH9 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2`
will output the interface file IWETH9.sol in crytic-export/interfaces.

There are optional flags for unrolling structs or excluding structs/enums/errors/events.

Tests can be found in `scripts/ci_test_interface.sh`

I believe I did testing/linting correctly, let me know if I need to fix anything.

### DIFF
--- a/scripts/ci_test_interface.sh
+++ b/scripts/ci_test_interface.sh
@@ -12,7 +12,7 @@ DIFF=$(diff crytic-export/interfaces/IWETH9.sol "$DIR_TESTS/test_1.sol")
 if [  "$DIFF" != "" ]
 then
     echo "slither-interface test 1 failed"
-    cat test_1.sol
+    cat "crytic-export/interfaces/IWETH9.sol"
     echo ""
     cat "$DIR_TESTS/test_1.sol"
     exit 255
@@ -25,7 +25,7 @@ DIFF=$(diff crytic-export/interfaces/IMock.sol "$DIR_TESTS/test_2.sol")
 if [  "$DIFF" != "" ]
 then
     echo "slither-interface test 2 failed"
-    cat test_2.sol
+    cat "crytic-export/interfaces/IMock.sol"
     echo ""
     cat "$DIR_TESTS/test_2.sol"
     exit 255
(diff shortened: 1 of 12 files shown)


0xalpharush:
We need to add "interface" to this list so the test is run in CI https://github.com/crytic/slither/blob/0ec487460690482c72cacdea6705e2c51bb3981e/.github/workflows/ci.yml#L45

0xGusMcCrae:
I'm not sure why the tests are failing on Windows but not Ubuntu. Any ideas?

Is there a different directory structure that's causing it to be unable to find the test contract?

scripts/ci_test_interface.sh
```diff
@@ -0,0 +1,95 @@
...
+
+solc-select use 0.8.19 --always-install
+
+#Test 1 - Etherscan target
+slither-interface WETH9 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2
+DIFF=$(diff crytic-export/interfaces/IWETH9.sol "$DIR_TESTS/test_1.sol")
+if [  "$DIFF" != "" ]
+then
+    echo "slither-interface test 1 failed"
+    cat test_1.sol
```
elopez, line 15: Should this be `crytic-export/interfaces/IWETH9.sol`? ditto for the other occurrences below.

elopez:
As for the test failures on Windows, it might be caused by `diff` picking up on the different line endings. The `test_N.sol` files commited in the repo have `\n` endings, and [autocrlf is disabled for solidity code in tests/](https://github.com/crytic/slither/blob/master/tests/.gitattributes). The files created by slither-interface will have `\r\n` line endings on Windows -- files are created with `open(....'w')` i.e. in text mode, which adjusts line endings to the platform specific variant.

You can pass `--strip-trailing-cr` to diff, which should make it skip the `\r` when comparing lines and hopefully fix the issue 👍

montyly:
This is great work, thanks @0xGusMcCrae

0xGusMcCrae:
Thanks, glad I could help out!

Issue #1849, opened 2023-04-19
[Bug-Candidate]: type_str(self) property in slithir/operations/unary.py not returning str


DarrenChangJR:
### Describe the issue:

```
    @property
    def type_str(self):
        return self._type.value

    def __str__(self):
        return f"{self.lvalue} = {self.type_str} {self.rvalue} "
```

`type_str` always returns 0 and 1 in place of "!" and "~" respectively. Is this intended behaviour?
If not, I will issue a pull request with below changes

```
    @property
    def type_str(self) -> str:
        return str(self._type)
```

### Code example to reproduce the issue:

contract UnaryOperations {
    uint256 public a = 10;
    bool public b = true;

    function executeUnaryOperations() public {
        a = -a; // Negation
        b = !b; // Logical negation
        a++; // Post-increment
        --a; // Pre-decrement
        a = ~a; // Bitwise negation
        a = +a; // Unary plus
    }
}

### Version:

0.9.3

0xalpharush:
@DarrenChangJR You're welcome to submit this patch as it should return a string.

PR #1914, opened 2023-05-15
Fix issue #1849: type_str not returning str

DarrenChangJR:


montyly:
Great fix, thanks @DarrenChangJR

Issue #1904, opened 2023-05-11
[Bug]: Ill-typed IR for pop


kevinclancy:
### Describe the issue:

The IR for pop obtains a reference variable to the final array element and deletes it. The type of this reference variable is always uint256, but it should be equal to the type of the array's elements. 

### Code example to reproduce the issue:
```
contract Poptest {
    struct S {
        int x;
        int y;
    }
    S[] a;
    function poptest() internal {
        a.pop();
    }
}
```

### Version:

0.9.3

### Relevant log output:

```shell
INFO:Printers:Contract Poptest
        Function Poptest.poptest() (*)
                Expression: a.pop()
                IRs:
                        REF_1 -> LENGTH a
                        TMP_1(uint256) = REF_1 (c)- 1
                        REF_2(uint256) -> a[TMP_1]
                        a = delete REF_2 
                        REF_3 -> LENGTH a
                        REF_3 (->a) := TMP_1(uint256)
```

kevinclancy:
I noticed that a fix for this was recently developed, but I don't think it handles popping from `bytes` correctly. I think the type of an element of `bytes` is `bytes1`, but the type that the fix gives the reference variable is `bytes`.

PR #1926, opened 2023-05-18
Fix bytes pop ir

smonicas:
Popping from `bytes` should give a `bytes1` type not `bytes` https://github.com/crytic/slither/issues/1904#issuecomment-1552251332


0xalpharush:
Let's add a test

0xalpharush:
I'm wondering if we should consider `bytes` an `ArrayType` and with type `bytes1`

PR #1933, opened 2023-05-22
Better struct handling in code generation util

webthethird:
Generates better code, which will compile without manual editing, when dealing with structs. 
- Avoids using a struct's canonical name, which will not match the generated interface once an `I` is prepended to the contract name.
- Handles arrays of structs in function parameters and returns
- Handles nested mappings inside and outside of structs
- Handles structs defined in other contracts/libraries
- Converts the type of enums used in structs, in case `include_enums` is false


webthethird:
@montyly @0xalpharush pip-audit may need to be updated again

webthethird:
One unresolved issue is that structs defined in a library and referenced in a using-for statement do not get included in the interface, though they may be referenced in a function's arguments or returns. For example:
The contract at [0xaec84eac74981ab22905919cb282b78c7ca782df](https://optimistic.etherscan.io/address/0xaec84eac74981ab22905919cb282b78c7ca782df#code#F14#L21) uses `FixedLib.Position` in a couple of public mappings: 
```solidity
contract Market is Initializable, AccessControlUpgradeable, PausableUpgradeable, ERC4626 {
  using FixedLib for FixedLib.Pool;
  using FixedLib for FixedLib.Position;

  /// @notice Tracks account's fixed deposit positions by maturity, account and position.
  mapping(uint256 => mapping(address => FixedLib.Position)) public fixedDepositPositions;
  /// @notice Tracks account's fixed borrow positions by maturity, account and position.
  mapping(uint256 => mapping(address => FixedLib.Position)) public fixedBorrowPositions;
  ...
```
for which `generate_interface` creates the following function signatures:
```solidity
    function fixedDepositPositions(uint256,address) external returns (Position memory);
    function fixedBorrowPositions(uint256,address) external returns (Position memory);
```
This causes the compilation error below:
```solidity
Error: Identifier not found or not unique.
--> DiffFuzzUpgrades.sol:257:71:
|
257 |     function fixedDepositPositions(uint256,address) external returns (Position memory);
|                                                                           ^^^^^^^^
```

Issue #1918, opened 2023-05-15
[Bug]: struct-destructuring assignments generate tuple variable whose "type" is a list of strings


kevinclancy:
### Describe the issue:

The "types" of the tuple variable generated from a struct-destructuring assignment are actually strings.

### Code example to reproduce the issue:

Modify the `set_type` method in `variable.py` to assert that if the argument is a list then its elements are types:
```
    def set_type(self, t: Optional[Union[List, Type, str]]) -> None:
        if isinstance(t, str):
            self._type = ElementaryType(t)
            return
        assert isinstance(t, (Type, list)) or t is None
        if isinstance(t, list):
            assert all(map(lambda x: isinstance(x, Type), t))
        self._type = t
```

Then run slither on the following file.   
```
pragma solidity ^0.8.19;

contract OtherContract {
    struct S {
        uint x;
        uint y;
    }

    mapping(uint => S) public m;
}

contract Test {
    OtherContract c;
    function test() external {
        (uint x, uint y) = c.m(0);
    }
}
```

The assertion fails.

### Version:

0.9.3

kevinclancy:
BTW. I believe this issue is caused by the `_convert_to_structure_to_list` function in `convert.py`, which should return `[return_type]` instead of `[return_type.type]`. I'll submit a PR with this fix soon.

PR #1935, opened 2023-05-23
fix: make _convert_to_structure_to_list return a type instead of an ElementaryType's `type` field

kevinclancy:
Fix https://github.com/crytic/slither/issues/1918

### DIFF
--- a/slither/slithir/convert.py
+++ b/slither/slithir/convert.py
@@ -1583,11 +1583,9 @@ def _convert_to_structure_to_list(return_type: Type) -> List[Type]:
     # }
     if isinstance(return_type, (MappingType, ArrayType)):
         return []
-    if isinstance(return_type, (ElementaryType, UserDefinedType, TypeAlias)):
-        return [return_type]
 
-    assert False
-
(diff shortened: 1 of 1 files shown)


0xalpharush:
Would you mind making this against the dev branch please?

slither/slithir/convert.py
```diff
@@ -1583,7 +1583,10 @@ def _convert_to_structure_to_list(return_type: Type) -> List[Type]:
     # }
     if isinstance(return_type, (MappingType, ArrayType)):
         return []
-    return [return_type.type]
+    if isinstance(return_type, (ElementaryType, UserDefinedType, TypeAlias)):
+        return [return_type]
+
+    assert False
```
0xalpharush, line 1589: Could we change the isinstance check to an assert instead of asserting False?

0xalpharush, line 1589:
```suggestion
    assert isinstance(return_type, (ElementaryType, UserDefinedType, TypeAlias)):
    return [return_type]
```

kevinclancy:
Yes. Done.

PR #1942, opened 2023-05-29
local variable location fix

Tiko7454:
When we had a function f() which returns a pair, the result passed to local variables didn't initialize the location attribute. For example: contract A {
	function f() public returns (int, bytes memory) {
		return (1, "asdf");
	}
	function g() public {
		(int x, bytes memory y) = f();
	}
}
the location of the local variable x as well as the location of the local variable y was None


smonicas:
Hi thanks for your pr. You are correct that this is an issue. Your fix would probably work however a clearer way is to override `_analyze_variable_attributes` in local_variable_init_from_tuple.py like it's done in local_variable.py (you can copy paste and should just work).

Tiko7454:
Thank you for your prompt response. I have done the corrections you suggested.

Tiko7454:
Seems like this 
`Collecting requests<2.30 (from -r /home/runner/work/_actions/pypa/gh-action-pip-audit/v1.0.7/requirements.txt (line 4))`
 should be upgraded to 2.31 or something like that
(its from the log of pip-audit)

smonicas:
Don't worry about it, we will update it in another pr

PR #1945, opened 2023-06-03
Optimizations for similar_variables.py

### LINKED ISSUE #1630: similar variables detector is extremely slow
This detector makes up the majority of runtime on large codebases
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   41.286   41.286 /Users/alpharush/tob/slither/slither/__main__.py:81(process_all)
        1    0.000    0.000   39.059   39.059 /Users/alpharush/tob/slither/slither/__main__.py:58(process_single)
        1    0.000    0.000   33.319   33.319 /Users/alpharush/tob/slither/slither/__main__.py:111(_process)
        1    0.000    0.000   33.317   33.317 /Users/alpharush/tob/slither/slither/slither.py:198(run_detectors)
        1    0.000    0.000   33.317   33.317 /Users/alpharush/tob/slither/slither/slither.py:204(<listcomp>)
       84    0.001    0.000   33.317    0.397 /Users/alpharush/tob/slither/slither/detectors/abstract_detector.py:176(detect)
        1    0.000    0.000   31.215   31.215 /Users/alpharush/tob/slither/slither/detectors/variables/similar_variables.py:72(_detect)

```
https://github.com/crytic/slither/blob/master/slither/detectors/variables/similar_variables.py#L63-L66

0xGusMcCrae:
Fixes #1630

Optimizations to the algorithm for detecting similar variable names in similar_variables.py. 

Runtime for slither/dev on current C4 competiton MAIA codebase:

    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    
    40      0.005    0.000    39.760   0.994   .../slither/detectors/variables/similar_variables.py:76(_detect)

Runtime for optimized version on same codebase:

    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    
    40      0.004    0.000    17.769   0.444   .../slither/detectors/variables/similar_variables.py:83(_detect)

Runtime is reduced by >50% here, and the issues detected are identical.

Changes (in `detect_sim`):

- Inner loop reworked so that it doesn't repeat comparisons. No need to compare the current outer loop's variable to anything with a lower index, as it's already been compared.
- caching v1.name.lower() in the outer loop so it is only repeated once per outer loop rather than once per inner loop

### DIFF
--- a/slither/detectors/variables/similar_variables.py
+++ b/slither/detectors/variables/similar_variables.py
@@ -47,9 +47,7 @@ class SimilarVarsDetection(AbstractDetector):
         Returns:
             bool: true if names are similar
         """
-        if len(seq1) != len(seq2):
-            return False
-        val = difflib.SequenceMatcher(a=seq1.lower(), b=seq2.lower()).ratio()
+        val = difflib.SequenceMatcher(a=seq1, b=seq2).ratio()
         ret = val > 0.90
         return ret
 
@@ -69,11 +67,15 @@ class SimilarVarsDetection(AbstractDetector):
 
         ret = []
         for v1 in all_var:
+            _len_v1 = len(v1.name) 
             for v2 in all_var:
-                if v1.name.lower() != v2.name.lower():
-                    if SimilarVarsDetection.similar(v1.name, v2.name):
-                        if (v2, v1) not in ret:
-                            ret.append((v1, v2))
+                if _len_v1 != len(v2.name):
+                    continue
+                _v1_name_lower = v1.name.lower()
+                _v2_name_lower = v2.name.lower()
+                if _v1_name_lower != _v2_name_lower:
+                    if SimilarVarsDetection.similar(_v1_name_lower, _v2_name_lower):
+                        ret.append((v1, v2))
 
         return set(ret)
 
--- a/slither/detectors/variables/similar_variables.py
+++ b/slither/detectors/variables/similar_variables.py
@@ -47,7 +47,9 @@ class SimilarVarsDetection(AbstractDetector):
         Returns:
             bool: true if names are similar
         """
-        val = difflib.SequenceMatcher(a=seq1, b=seq2).ratio()
+        if len(seq1) != len(seq2):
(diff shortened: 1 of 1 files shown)


0xGusMcCrae:
There were some further optimizations (down to ~14 seconds for the MAIA codebase) that I didn't include in this PR since I wasn't 100% sure that they wouldn't break anything if anything external depended on the `similar` function having its exact current behavior.

It's possible to feed already lowercase variable names into `similar` and bring the length equality check from `similar` out into `detect_sim`

And AFAIK this line can be deleted:

https://github.com/crytic/slither/blob/79ff12a5e290a0695826f276bf848b1165e5d57a/slither/detectors/variables/similar_variables.py#L75 

since duplicates will be eliminated when the list is converted to a set in the return statement.

Something like this:

    @staticmethod
    def similar(seq1: str, seq2: str) -> bool:
        ...
        Returns:
            bool: true if names are similar
        """
        val = difflib.SequenceMatcher(a=seq1, b=seq2).ratio()
        ret = val > 0.90
        return ret

    @staticmethod
    def detect_sim(contract: Contract) -> Set[Tuple[LocalVariable, LocalVariable]]:
        ...
        all_var = list(set(all_var + contract_var))

        ret = []
        for i in range(len(all_var)):
            v1 = all_var[i]
            _v1_name_lower = v1.name.lower()
            for j in range(i,len(all_var)): 
                v2 = all_var[j]
                if len(v1.name) != len(v2.name):
                    continue
                _v2_name_lower = v2.name.lower()
                if _v1_name_lower != _v2_name_lower:
                    if SimilarVarsDetection.similar(_v1_name_lower, _v2_name_lower):
                        ret.append((v1, v2))

        return set(ret)

So if further optimization is desired, these are possibilities.

0xalpharush (approved):

montyly:
I think we can reformat `similar` to follow your recommendation, it's only used in this detector as far as I can tell

Issue #1906, opened 2023-05-11
[Bug-Candidate]: Arguments for named struct constructor are not reordered


kevinclancy:
### Describe the issue:

Types `struct S { ... }` allow "named constructors" that look like `S({ field1: val1, field2: val2, ... })`. The fields may be given in any permutation, even permutations distinct from the one shown in the type declaration. But the `NewStruct` IR generated by Slither does not use names and does not reorder the arguments, so `S({field2: val2, field1: val1})` would incorrectly get translated to `TMP = new S(val2, val1)` instead of `TMP = new S(val1, val2)`.
  

### Code example to reproduce the issue:

contract ReorderStruct {
    struct S {
        int[] a;
        uint x;
    }

    S[] public s;
    function test() external {
        s.push(S({x: 2, a: new int[](2)}));
    }
}

### Version:

0.9.3

### Relevant log output:

```shell
INFO:Printers:Contract ReorderStruct
        Function ReorderStruct.test() (*)
                Expression: s.push(S(2,new int256[](2)))
                IRs:
                        TMP_1 = new int256[](2)
                        TMP_2 = new S(2,TMP_1)
                        REF_1 -> LENGTH s
                        TMP_4(uint256) := REF_1(uint256)
                        TMP_5(uint256) = TMP_4 (c)+ 1
                        REF_1 (->s) := TMP_5(uint256)
                        REF_2(ReorderStruct.S) -> s[TMP_4]
                        REF_2 (->s) := TMP_2(ReorderStruct.S)
```

PR #1949, opened 2023-06-06
fix: reorder named arguments to match declaration order

kevinclancy:
Fixes https://github.com/crytic/slither/issues/1906

A function call with explicitly named parameters, of the form `f({argName1: arg1, ...})`, may list the parameters in an order different from that of the function's declaration. In this case, the IR should reorder the arguments to match the function declaration order before generating IR.  

In practice, I have only seen named argument calls for struct constructors, so I suspect they are the only form that is documented. For calls to built-in functions, contract constructors, and dynamic internal function calls, named arguments produce confusing type checking errors. For all other forms of function calls, named arguments work, but I've never seen them used in a real project.

This PR implements named argument reordering for the following forms of function calls:
* struct constructors (most important)
* static internal calls 

For the following forms of function calls, the order of the declaration arguments is not directly available during IR generation, so the PR did not implement reordering for them:
* external calls (HighLevelCall)
* library calls
* event calls

The `get_declared_param_names` function returns `None` for these unimplemented call forms.

#### Example

```
contract ReorderStruct {
    struct S {
        int[] a;
        uint x;
    }

    S[] public s;
    function test() external {
        S memory s = S({x: 2, a: new int[](2)});
    }
}
```

Output Before:
```
INFO:Printers:Contract ReorderStruct
        Function ReorderStruct.test() (*)
                Expression: s = S({x:2,a:new int256[](2)})
                IRs:
                        TMP_1 = new int256[](2)
                        TMP_2 = new S(2,TMP_1)
                        s(ReorderStruct.S) := TMP_2(ReorderStruct.S)
```

Output After
```
INFO:Printers:Contract ReorderStruct
        Function ReorderStruct.test() (*)
                Expression: s = S({x:2,a:new int256[](2)})
                IRs:
                        TMP_1 = new int256[](2)
                        TMP_2 = new S(TMP_1,2)
                        s(ReorderStruct.S) := TMP_2(ReorderStruct.S)
```


0xalpharush:
@kevinclancy Unless I'm mistaken, the output before/after look the same in the PR description but maybe it's a mistake?

montyly:
This is a great addition, thanks @kevinclancy 

Could we add a few unit tests?

kevinclancy:
Sure, @montyly . I just added a couple of tests.

slither/slithir/convert.py
```diff
@@ -385,6 +385,70 @@ def integrate_value_gas(result: List[Operation]) -> List[Operation]:
...
+    names -
+        Parameter names in call order
+    decl_names -
+        Parameter names in declaration order
+    #### Returns
+    Reordered arguments to constructor call, now in declaration order
+    """
+    assert isinstance(args, list)
+    assert isinstance(call_names, list)
+    assert isinstance(decl_names, list)
```
0xalpharush, line 440: I think these can be removed since the function signature is typed and indicates this

slither/slithir/convert.py
```diff
@@ -385,6 +385,70 @@ def integrate_value_gas(result: List[Operation]) -> List[Operation]:
...
+    if isinstance(ins, HighLevelCall) and isinstance(ins.function, str):
+        return None
+    if isinstance(ins, (InternalCall, LibraryCall, HighLevelCall)):
+        if isinstance(ins.function, Function):
+            return [p.name for p in ins.function.parameters]
+        return None
+    if isinstance(ins, InternalDynamicCall):
+        return [p.name for p in ins.function_type.params]
+
+    assert isinstance(ins, EventCall)
```
0xalpharush, line 417: What is this assertion for? Should `EventCall` return an order?

slither/slithir/convert.py
```diff
@@ -385,6 +385,70 @@ def integrate_value_gas(result: List[Operation]) -> List[Operation]:
...
+    None -
+        Workaround: Unable to obtain list of parameters in declaration order
+    """
+    if isinstance(ins, NewStructure):
+        return [x.name for x in ins.structure.elems_ordered if not isinstance(x.type, MappingType)]
+    if isinstance(ins, NewContract):
+        return [p.name for p in ins.contract.constructor.parameters]
+    if isinstance(ins, (LowLevelCall, NewElementaryType, NewArray)):
+        # named arguments are incompatible with these call forms
+        assert False
```
0xalpharush, line 407: Can we update the function signature to only be subclasses of call which are accepted and remove this?  We can add an assertion like `assert isinstance(ins, NewStructure, NewContract... )`, too

kevinclancy: It might be better to add `assert not isinstance(LowLevelCall, NewElementaryType, NewArray)` at the top of the `get_declared_param_names` function.

We could also make the `assert False` case return `None` instead. Of course, it would be a hassle to get test coverage for this.

Issue #1758, opened 2023-03-14
[Bug]: Infinite loop in SlitherCompilationUnitSolc._analyze_all_enums


romcyncynatus:
### Describe the issue:

When analyzing certain contracts, for example, 0x757bc268bd50da88b2d0cf1966652b18e56ca803, an infinite loop in `SlitherCompilationUnitSolc._analyze_all_enums` is encountered. The `contracts_to_be_analyzed` list seems to contain only a single contract that is being added and removed repeatedly.

### Code example to reproduce the issue:

```
# https://etherscan.io/address/0x757bc268bd50da88b2d0cf1966652b18e56ca803
import slither
s = slither.Slither('0x757bc268bd50da88b2d0cf1966652b18e56ca803')
```

### Version:

0.9.2

### Relevant log output:

```shell
>>> import slither
>>> s = slither.Slither('0x757bc268bd50da88b2d0cf1966652b18e56ca803')
WARNING:CryticCompile:Warning: @uniswap/v2-core/contracts/interfaces/IUniswapV2Factory.sol: Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.

Warning: contracts/core/work/ContributionBoard.sol:390:5: Warning: Function state mutability can be restricted to view
    function _beforeTokenTransfer(
    ^ (Relevant source part starts here and spans across multiple lines).


^CTraceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rom/projects/smart-contracts-similarity/venv/lib/python3.10/site-packages/slither/slither.py", line 133, in __init__
    self._init_parsing_and_analyses(kwargs.get("skip_analyze", False))
  File "/home/rom/projects/smart-contracts-similarity/venv/lib/python3.10/site-packages/slither/slither.py", line 139, in _init_parsing_and_analyses
    parser.parse_contracts()
  File "/home/rom/projects/smart-contracts-similarity/venv/lib/python3.10/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 496, in parse_contracts
    self._analyze_all_enums(contracts_to_be_analyzed)
[...]

PR #1956, opened 2023-06-08
inform user if inheritance cannot be resolved

0xalpharush:
Add a fix suggestion for cases like https://github.com/crytic/slither/issues/1758 and prevent infinite loop

### DIFF
--- a/slither/solc_parsing/slither_compilation_unit_solc.py
+++ b/slither/solc_parsing/slither_compilation_unit_solc.py
@@ -438,7 +438,7 @@ Please rename it, this name is reserved for Slither's internals"""
                     )
                     if target == contract_parser.underlying_contract:
                         raise InheritanceResolutionError(
-                            "Could not resolve contract inheritance. This is likely caused by contract name reuse (see https://github.com/crytic/slither/issues/1758)."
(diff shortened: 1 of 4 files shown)


slither/solc_parsing/slither_compilation_unit_solc.py
```diff
@@ -432,7 +436,12 @@ def parse_contracts(self) -> None:  # pylint: disable=too-many-statements,too-ma
                     target = contract_parser.underlying_contract.file_scope.get_contract_from_name(
                         contract_name
                     )
-                    assert target
+                    if target == contract_parser.underlying_contract:
+                        raise InheritanceResolutionError(
+                            "Could not resolve contract inheritance. This is likely caused by contract name reuse (see https://github.com/crytic/slither/issues/1758)."
```
montyly, line 441: I think the issue is more related to import rename that lead to confusion, rather that contract name reuse. Maybe we can word it as "This is likely caused by an import renaming that collide with existing names", or something like that
What do you think?

0xalpharush, line 441:
```suggestion
                            "Could not resolve contract inheritance. This is likely caused by an import renaming that collides with existing names (see https://github.com/crytic/slither/issues/1758)."
```

PR #1968, opened 2023-06-16
Fixed issue which disallowed using operator[] with TopLevelVariables

Tiko7454:
```solidity
bytes16 constant HEX = '0123456789ABCDEF';

contract C {
    function f(uint i) public {
        HEX[1];
    }
}
```
This code was throwing an assertion error due to HEX was expected to be an lvalue, but it doesn't need to be


0xalpharush:
Would you mind making this PR against the `dev` branch please?

montyly:
Nice catch, thanks @Tiko7454

PR #1974, opened 2023-06-20
Fixed canonical_name of event 

0xxfu:


0xalpharush:
It looks like the detector tests need to be updated since this affects the format. Can you run `pytest tests/e2e/detectors/test_detectors.py --insta update` on your branch of slither and commit the updates`?

montyly:
Great catch, thanks @0xxfu. This was merged with #1976

PR #1977, opened 2023-06-21
allow toplevel var to be an lvalue

0xalpharush:
replaces https://github.com/crytic/slither/pull/1968


montyly:
I am a bit unsure about this. Top level variable can't be a lvalue for most operation, as they need to be constant (so no assignment etc), so specifying it like in https://github.com/crytic/slither/pull/1968 might be more explicit

PR #1980, opened 2023-06-21
Additional optimizations for similar_variables.py

0xGusMcCrae:
Issue #1630 

Adding in the additional optimizations for the similar variable name detector as described in https://github.com/crytic/slither/pull/1945#issuecomment-1575060352 

Changes:

- length check brought out of `similar` into `detect_sim`. If the lengths aren't the same, no need to run `v2.name.lower()`
- `v1.name.lower()` and `v2.name.lower()` are cached in `detect_sim` and given as inputs for `similar` so that .lower() doesn't have to be run an extra 2 times in the `SequenceMatcher` call.

Again running on the C4 contest Maia codebase:

Current slither/dev:
   
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    
    40      0.004    0.000    17.769   0.444   .../slither/detectors/variables/similar_variables.py:83(_detect)

With new optimizations:

    ncalls  tottime  percall  cumtime  percall filename:lineno(function)

    40      0.004    0.000    14.867   0.372   .../slither/detectors/variables/similar_variables.py:84(_detect) 

Detected issues unchanged, ~16% reduction in runtime.

### DIFF
--- a/slither/detectors/variables/similar_variables.py
+++ b/slither/detectors/variables/similar_variables.py
@@ -47,9 +47,7 @@ class SimilarVarsDetection(AbstractDetector):
         Returns:
             bool: true if names are similar
         """
-        if len(seq1) != len(seq2):
-            return False
-        val = difflib.SequenceMatcher(a=seq1.lower(), b=seq2.lower()).ratio()
+        val = difflib.SequenceMatcher(a=seq1, b=seq2).ratio()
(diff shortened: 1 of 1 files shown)


montyly:
This is great, thanks @0xGusMcCrae for all these optimizations

PR #1981, opened 2023-06-21
refactor: introduce lowering context to simplify cfg construction

0xalpharush:
This PR introduces a `LoweringContext` which stores whether the current scope is checked and the loop destination for break/continue.
Depends on https://github.com/crytic/slither/pull/1951
Replaces https://github.com/crytic/slither/pull/629

This will unblock fixes like https://github.com/crytic/slither/pull/1651


montyly:
Can we add unit tests, in particular with nested loop and break/continue, to ensure this does not break the cfg construction?

Issue #1973, opened 2023-06-20
[Bug]: Dominator-deadcode problem


Tiko7454:
### Describe the issue:

The i++ node has no father though it's not  an ENTRY_POINT, so the only dominator of the node i++ is i++ itself. So when slither wants to find dominators for IF_LOOP node, it intersects the dominators of BEGIN_LOOP and i++, and gets nothing. Thus IF_LOOP has no immediate dominator which throws an assertion error

### Code example to reproduce the issue:

```solidity
contract C {
    function f() public {
        for (uint256 i = 1; i < 10; i++) {
            break;  // continue
        }
    }
}
```

### Version:

0.9.3

### Relevant log output:

```shell
'solc --version' running
'solc a.sol --combined-json abi,ast,bin,bin-runtime,srcmap,srcmap-runtime,userdoc,devdoc,hashes,compact-format --allow-paths . a.sol' running
Compilation warnings/errors on a.sol:
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> a.sol

Warning: Source file does not specify required compiler version! Consider adding "pragma solidity ^0.7.5;"
--> a.sol

Warning: Unreachable code.
 --> a.sol:3:37:
  |
3 |         for (uint256 i = 1; i < 10; i++) {
  |                                     ^^^

Warning: Function state mutability can be restricted to pure
 --> a.sol:2:5:
  |
2 |     function f() public {
  |     ^ (Relevant source part starts here and spans across multiple lines).


ERROR:SlitherSolcParsing:
Failed to convert IR to SSA for C contract. Please open an issue https://github.com/crytic/slither/issues.
 
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/slither/__main__.py", line 810, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/usr/lib/python3.10/site-packages/slither/__main__.py", line 101, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/usr/lib/python3.10/site-packages/slither/__main__.py", line 79, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/usr/lib/python3.10/site-packages/slither/slither.py", line 135, in __init__
[...]

PR #1982, opened 2023-06-22
allow dead code to not have an immediate dominator

0xalpharush:
The example in the issue highlights that the expression `i++` will cause issues since it is dead code and thus does not have any dominating nodes 

closes https://github.com/crytic/slither/issues/1973


slither/core/dominators/utils.py
```diff
@@ -95,5 +95,4 @@ def compute_dominance_frontier(nodes: List["Node"]) -> None:
                     runner.dominance_frontier = runner.dominance_frontier.union({node})
                 while runner != node.immediate_dominator:
                     runner.dominance_frontier = runner.dominance_frontier.union({node})
-                    assert runner.immediate_dominator
```
montyly, line 98: For the sake of the type annotation, should we do a break if the immediate dominator is none?

This will also allow us to add a code comment regarding the dead code

PR #1983, opened 2023-06-22
fix typo in Contract.get_state_variable_from_canonical_name()

yisun92:
The function should compare v.canonical_name with the parameter. Currently, it works the same as get_state_variable_from_name()


0xalpharush:
Thanks @yisun92!

PR #2002, opened 2023-06-26
Fix execution in Python 3.11

elopez:
This fixes the following error:

    ValueError: mutable default <class 'slither.utils.loc.LoCInfo'> for field src is not allowed: use default_factory


mimoo:
still getting that error: https://github.com/crytic/slither/issues/2005

Issue #1950, opened 2023-06-07
Yell at people for using ecrecover


bsamuels453:
### Describe the desired feature

Any use of `ecrecover()` outside of generally accepted libraries should be flagged as a finding deserving deeper investigation. 

If we want this to be a more sophisticated detector, figure out a way for it to check statements 310-313 from the ethereum yellow paper: 
![image](https://github.com/crytic/slither/assets/1222451/991fd75e-5d1c-4b81-919a-72a17a4a57c8)

Issue #1544, opened 2023-01-03
improvements to zero address check detector


0xalpharush:
follow on https://github.com/crytic/slither/pull/1043 https://github.com/crytic/slither/pull/1033

I think the usage of the data dependency here: 
https://github.com/crytic/slither/blob/191d6d4e661cd14c715f0f7b86593b69037322c3/slither/detectors/operations/missing_zero_address_validation.py#L81-L82

Will lead to miss TP. For example the following code:

```solidity
contract C{

    address owner;

    modifier onlyOwner(){
        require(msg.sender == owner);
        _;
    }

    function set(address addr) public{
        require(msg.sender == owner);
        owner = addr;
    }
}
```

Would raise a warning before the PR, and would not with these changes. 

I am wondering also if we could not refactor the detector's logic - it becomes complex to follow.

Maybe we can mark all the `Constant(0)` usage as source of dependency, and check for any `==`/`!=` operations where the left variable is a parameter, and the right variable is dependent on one of the `Constant(0)` we detected?

_Originally posted by @montyly in https://github.com/crytic/slither/issues/1033#issuecomment-1034916789_

0xalpharush:
See also https://github.com/crytic/slither/pull/1515

PR #2015, opened 2023-06-30
Detector for ecrecover return value validation and use of nonces

tuturu-tech:
Used https://github.com/crytic/slither/pull/1516 as a starting point, related to issue https://github.com/crytic/slither/issues/1950

Detects the usage of ecrecover and checks if the signer address is validated against the zero address.
If a function uses ecrecover, checks if a nonce is present in any of the hashes.

Currently does so via string comparisons and naively checks any hash for a nonce, so it could have a lot of FPs


slither/detectors/statements/ecrecover.py
```diff
@@ -0,0 +1,144 @@
...
+                    if "ecrecover(bytes32,uint8,bytes32,bytes32)" in str(var.expression):
+                        address_list.append(var)
+                        var_nodes[var].append(node)
+            if SolidityFunction("keccak256(bytes)") in node.solidity_calls:
+                for ir in node.irs:
+                    if isinstance(ir, SolidityCall) and ir.function == SolidityFunction(
+                        "keccak256(bytes)"
+                    ):
+                        if "nonce" not in str(ir.expression):
+                            nonce_results.append(node)
```
0xalpharush, line 138: Some APIs will require that a hash is passed in and I think this would cause FPs. You may have to explore all paths starting from `functions_entry_points` and find the union of values not hashed but passed to ecrecover

slither/detectors/statements/ecrecover.py
```diff
@@ -0,0 +1,144 @@
...
+    for node in function.nodes:
+        if node.contains_if() or node.contains_require_or_assert():
+            for ir in node.irs:
+                expression = str(ir.expression)
+                if isinstance(ir, Binary) and str(var) in expression:
+                    if (
+                        ir.type == BinaryType.NOT_EQUAL or ir.type == BinaryType.EQUAL
+                    ) and "address(0)" in expression:
+                        return True
+    return False
```
0xalpharush, line 110: Not sure if it's helpful but there were some improvements desired in the missing-zero-check outlined in https://github.com/crytic/slither/issues/1544 that may be relevant here

slither/detectors/statements/ecrecover.py
0xalpharush: Like this https://github.com/crytic/slither/blob/3d4f934d3228f072b7df2c5e7252c64df4601bc8/slither/detectors/statements/msg_value_in_loop.py#L13-L31

PR #2027, opened 2023-07-05
UnaryOperation: -variable and +variable doesn't make variable an lvalue

SheldonHolmgren:
Small bugfix for UnaryOperation. Until now in code like this:
```solidity
function some_view() view returns (int) {
    return -someParameter;
}
```
Slither believes that `someParameter` is an lvalue. One of the results is that `slither.visitors.expression.write_var` believes that `someParameter` is written to.

I'm happy to add some test if needed.


0xalpharush:
Thanks @SheldonHolmgren!

Issue #2025, opened 2023-07-03
[Bug]: Ternary expression lifting breaks tuple assignments


kevinclancy:
### Describe the issue:

The following source code:
```
contract Test {

    function getVal(uint z) internal returns(uint, uint) {
        return (0, 1);
    }

    function test(bool flag) external {
        uint a;
        (a, ) = getVal(flag ? 3 : 1);
    }
}
```

generates the following IR:
```
INFO:Printers:Contract Test
        Function Test.getVal(uint256) (*)
                Expression: (0,1)
                IRs:
                        RETURN 0,1
        Function Test.test(bool) (*)
                Expression: flag
                IRs:
                        CONDITION flag
                Expression: (a) = getVal(3)
                IRs:
                        TUPLE_0(uint256,uint256) = INTERNAL_CALL, Test.getVal(uint256)(3)
                        a(uint256) := TUPLE_0([<slither.core.solidity_types.elementary_type.ElementaryType object at 0x1023476d0>, <slither.core.solidity_types.elementary_type.ElementaryType object at 0x102347910>])
                Expression: (a) = getVal(1)
                IRs:
                        TUPLE_1(uint256,uint256) = INTERNAL_CALL, Test.getVal(uint256)(1)
                        a(uint256) := TUPLE_1([<slither.core.solidity_types.elementary_type.ElementaryType object at 0x1023476d0>, <slither.core.solidity_types.elementary_type.ElementaryType object at 0x102347910>])
```

The IR assigns the variable `a` of type `uint256` to a tuple variable. Instead, `a` should store the projection (unpacking) of tuple element 0.

### Code example to reproduce the issue:

contract Test {

    function getVal(uint z) internal returns(uint, uint) {
        return (0, 1);
    }

    function test(bool flag) external {
        uint a;
        (a, ) = getVal(flag ? 3 : 1);
    }
}

### Version:

0.9.5

### Relevant log output:

```shell
INFO:Printers:Contract Test
        Function Test.getVal(uint256) (*)
                Expression: (0,1)
                IRs:
                        RETURN 0,1
        Function Test.test(bool) (*)
                Expression: flag
                IRs:
                        CONDITION flag
                Expression: (a) = getVal(3)
                IRs:
                        TUPLE_0(uint256,uint256) = INTERNAL_CALL, Test.getVal(uint256)(3)
[...]

PR #2033, opened 2023-07-06
fix: preserve None tuple components in the left and right liftings of ternary expressions

kevinclancy:
This PR fixes [Issue 2025](https://github.com/crytic/slither/issues/2025). 

The problem was that the code for lifting ternary expressions would avoid processing `None` as a subexpression. Instead, we need to include the subexpression `None` in both the true and false liftings.

### DIFF
--- a/slither/utils/expression_manipulations.py
+++ b/slither/utils/expression_manipulations.py
@@ -147,7 +147,7 @@ class SplitTernaryExpression:
         for next_expr in expression.expressions:
             # TODO: can we get rid of `NoneType` expressions in `TupleExpression`?
             # montyly: this might happen with unnamed tuple (ex: (,,,) = f()), but it needs to be checked
-            if next_expr:
+            if next_expr is not None:
 
                 if self.conditional_not_ahead(
                     next_expr, true_expression, false_expression, f_expressions
@@ -158,6 +158,9 @@ class SplitTernaryExpression:
                         true_expression.expressions[-1],
                         false_expression.expressions[-1],
                     )
+            else:
+                true_expression.expressions.append(None)
+                false_expression.expressions.append(None)
 
     def convert_index_access(
         self, next_expr: IndexAccess, true_expression: Expression, false_expression: Expression
--- a/tests/unit/slithir/test_data/ternary_expressions.sol
+++ b/tests/unit/slithir/test_data/ternary_expressions.sol
@@ -61,6 +61,7 @@ contract D {
     }
 
     function a(uint n) external {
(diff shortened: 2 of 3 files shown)


kevinclancy:
I'll add a test for this.

kevinclancy:
I added a test, but the new lines of source code are reported as not having coverage. I'm not sure why.

0xalpharush:
After this PR:
```
INFO:Printers:Contract D
        Function D.values(uint256) (*)
                Expression: (0,1)
                IRs:
                        RETURN 0,1
        Function D.a(uint256) (*)
                Expression: n > 0
                IRs:
                        TMP_0(bool) = n > 0
                        CONDITION TMP_0
                Expression: (a,None) = values(1)
                IRs:
                        TUPLE_0(uint256,uint256) = INTERNAL_CALL, D.values(uint256)(1)
                        a(uint256)= UNPACK TUPLE_0 index: 0 
                Expression: (a,None) = values(0)
                IRs:
                        TUPLE_1(uint256,uint256) = INTERNAL_CALL, D.values(uint256)(0)
                        a(uint256)= UNPACK TUPLE_1 index: 0 
```
I would expect there to be one tuple and the indices to be zero and one
```
INFO:Printers:Contract D
        Function D.values(uint256) (*)
                Expression: (0,1)
                IRs:
                        RETURN 0,1
        Function D.a(uint256) (*)
                Expression: n > 0
                IRs:
                        TMP_0(bool) = n > 0
                        CONDITION TMP_0
                Expression: (a,None) = values(1)
                IRs:
                        TUPLE_0(uint256,uint256) = INTERNAL_CALL, D.values(uint256)(1)
                        a(uint256)= UNPACK TUPLE_0 index: 1
                Expression: (a,None) = values(0)
                IRs:
                        TUPLE_0(uint256,uint256) = INTERNAL_CALL, D.values(uint256)(0)
                        a(uint256)= UNPACK TUPLE_0 index: 0 
```
Do you agree? I think this would require inserting the internal call before the expression and then creating the unpack operations in the true/ false sons. I'm not sure how best to do that yet

kevinclancy:
Because the lvalue has the form `(a,)`, we should project the component at index `0` into the variable `a`. I don't think we would want to project the component at index `1`.

The tuple variable is used only for the subsequent unpack instruction, so I don't think it matters whether we have one or two tuple variables.

kevinclancy:
No. The first `UNPACK` operation, like the second, should unpack from index 0 instead of 1.

0xalpharush:
Thanks @kevinclancy! merged in https://github.com/crytic/slither/pull/2120

Issue #1913, opened 2023-05-15
[Bug-Candidate]: When variable is reused in two multi-return assignments, original tuple index is reused 


kevinclancy:
### Describe the issue:

When a variable is used to store a result of a multi-return function call and then later used to store the result of a different multi-return function call, the index of the first UNPACK operation is used for the second UNPACK. Instead, the index of the second UNPACK should be based on where the variable was listed on the lhs of the assignment.

### Code example to reproduce the issue:

contract Test {
    function threeRet(int z) internal returns(int, int, int) {
        return (1,2,3);
    }
    function twoRet(int z) internal returns(int, int) {
        return (3,4);
    }

    function test() external returns(int) {
        (int a, int b, int c) = threeRet(3);
        (a, c) = twoRet(b);
        return b;
    }
}

### Version:

0.9.3

### Relevant log output:

```shell
INFO:Printers:Contract Test
        Function Test.threeRet(int256) (*)
                Expression: (1,2,3)
                IRs:
                        RETURN 1,2,3
        Function Test.twoRet(int256) (*)
                Expression: (3,4)
                IRs:
                        RETURN 3,4
        Function Test.test() (*)
                Expression: (a,b,c) = threeRet(3)
                IRs:
                        TUPLE_0(int256,int256,int256) = INTERNAL_CALL, Test.threeRet(int256)(3)
                        a(int256)= UNPACK TUPLE_0 index: 0 
                        b(int256)= UNPACK TUPLE_0 index: 1 
                        c(int256)= UNPACK TUPLE_0 index: 2 
                Expression: (a,c) = twoRet(b)
                IRs:
                        TUPLE_1(int256,int256) = INTERNAL_CALL, Test.twoRet(int256)(b)
                        a(int256)= UNPACK TUPLE_1 index: 0 
                        c(int256)= UNPACK TUPLE_1 index: 2 
                Expression: b
                IRs:
                        RETURN b

Note that we unpack index 2 into `c` for the second call, but the result of `twoRet` doesn't even have a component with index 2.
```

kevinclancy:
I imagine this is caused by this code in `expression_to_slithir.py`:
```
                        # The following test is probably always true?
                        if (
                            isinstance(left[idx], LocalVariableInitFromTuple)
                            and left[idx].tuple_index is not None
                        ):
                            index = left[idx].tuple_index
```
I wonder why this code exists. Can we use `idx` as `index`?

duckki:
[...]

PR #2034, opened 2023-07-07
fix: preserve empty tuple components during declaration-to-assignment conversion

kevinclancy:
This fixes [issue 1913](https://github.com/crytic/slither/issues/1913).

Previously, when performing an assignment whose variables were originally declared in a tuple-based declaration, the projection components were reused:
```
    function test() external returns(int) {
        (int a, int b, int c) = threeRet(3);
        (a, c) = twoRet(b);
        return b;
    }
...
  Expression: (a,c) = twoRet(b)
  IRs:
    TUPLE_1(int256,int256) = INTERNAL_CALL, Test.twoRet(int256)(b)
    a(int256)= UNPACK TUPLE_1 index: 0 
    c(int256)= UNPACK TUPLE_1 index: 2 
```

This happened because of some code in the `_post_assignment_operation` function in `expression_to_slithir.py`:
```
                        # The following test is probably always true?
                        if (
                            isinstance(left[idx], LocalVariableInitFromTuple)
                            and left[idx].tuple_index is not None
                        ):
                            index = left[idx].tuple_index
```

We shouldn't be reusing the variable's original projection component, but there is a reason that code was written; namely, [this code](https://github.com/crytic/slither/blob/master/slither/solc_parsing/declarations/function.py#L755) translates a multi-variable declaration into a multi-variable assignment. 

It also removes padding `None` entries and records each component index in its corresponding variable instead. As we've seen, this approach doesn't work, because not all tuple assignments are generated from variable declarations. To solve this, this PR leaves the `None` components inside of the tuple in order to select the correct projection index for each variable.

### DIFF
--- a/slither/solc_parsing/declarations/function.py
+++ b/slither/solc_parsing/declarations/function.py
@@ -815,19 +815,26 @@ class FunctionSolc(CallerContextExpression):
                             new_node = self._parse_variable_definition_init_tuple(
                                 new_statement, i, new_node, scope
                             )
+                        else:
+                            variables.append(None)
                         i = i + 1
 
                     var_identifiers = []
                     # craft of the expression doing the assignement
                     for v in variables:
-                        identifier = {
-                            "nodeType": "Identifier",
-                            "src": v["src"],
-                            "name": v["name"],
-                            "referencedDeclaration": v["id"],
-                            "typeDescriptions": {"typeString": v["typeDescriptions"]["typeString"]},
-                        }
-                        var_identifiers.append(identifier)
+                        if v is not None:
+                            identifier = {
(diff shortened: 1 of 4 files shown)


0xalpharush:
Can you move the tests to `tests/unit/slithir` and run `pytest tests/e2e/detectors/test_detectors.py --insta update` please?

kevinclancy:
@0xalpharush When I run with `--insta update-new`, I get an error:
```
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --insta update-new
  inifile: None
  rootdir: /Users/kevin.clancy/my-slither
```

PR #2065, opened 2023-07-24
Perf improvements

montyly:


0xalpharush:
I think I'd prefer to avoid a c dependency in case we want to use PyPy. Also I'm not sure if it would interfere with distributing a binary

montyly:
Yes it's a good point. @smonicas : do you remember what were the bench improvements?

0xalpharush:
I think we are going to deprecate this detector altogether instead of bringing in a dependency solely to speed it up

PR #2068, opened 2023-07-27
Better struct handling in code generation util

webthethird:
Generates better code, which will compile without manual editing, when dealing with structs. 
- Avoids using a struct's canonical name, which will not match the generated interface once an `I` is prepended to the contract name
- Handles arrays of structs in function parameters and returns
- Handles nested mappings inside and outside of structs
- Handles structs defined in other contracts/libraries
- Converts the type of enums used in structs, in case `include_enums` is false

Resolves [this issue](https://github.com/crytic/diffusc/issues/17) in crytic/diffusc, where the generated interface would not compile due to a reference in a function signature to a struct defined in another contract.

Replaces #1933


0xalpharush:
Can you add the failing example as a regression test and run `make reformat` please?

webthethird:
> Can you add the failing example as a regression test and run `make reformat` please?

Done!

PR #2076, opened 2023-08-02
detectors: cache_array_length: include source mapping in finding

elopez:
The detector was manually string-interpolating the code and line information, which left the finding without source mapping metadata and broke filtering. Include the node directly on the finding info so that the metadata is preserved.


elopez:
The printing becomes a bit more unfriendly with this change (e.g. `i_scope_23` instead of `i`) because `Output` uses `.expression` instead of `.content` to display `Node` source mappings.

0xalpharush:
@elopez I didn't realize the ramifications on the filtering when I made the changes in https://github.com/crytic/slither/pull/2019. I'm not sure if we could switch to `.content` generally, but improving the code context and displaying detector findings more clearly is something we'd welcome. For now, we can just revert back to how it was previously and merge this.

The tests just need to be updated with `pytest tests/e2e/detectors/test_detectors.py --insta update`

PR #2084, opened 2023-08-11
Improve mapping-deletion detector for nested mappings

smonicas:


0xalpharush:
Can we add a test the triggers the detector?

Issue #1034, opened 2022-02-04
Allow underscore in variable naming convention


0xalpharush:
It's pretty common to include underscores to prefix internal variables or to distinguish between arguments and state variables. We should not include these in the camel case naming convention detector

stephenctw:
They claim to support the underscore [rule](https://github.com/crytic/slither/wiki/Detector-Documentation#rule-exceptions) but I also hit this issue.

smallbatch-apps:
This is affecting us, too. The use of `_name` or `name_` to prevent collision is very common, and necessary. Using `_name` should not trigger an error, as it is a standard convention.

0xalpharush:
This lines need to be update to include internal state variables:
https://github.com/crytic/slither/blob/0ec487460690482c72cacdea6705e2c51bb3981e/slither/detectors/naming_convention/naming_convention.py#L165

devtooligan:
**bump**

PR #2110, opened 2023-09-04
Allow underscore in variable naming convention for internal state variables

dokzai:
Hello!

This is my attempt at addressing #1034 based on the recommendation provided in the thread. 

Thanks!


tests/e2e/detectors/test_data/naming-convention/0.4.25/naming_convention.sol
```diff
@@ -53,6 +53,7 @@ contract Test {
 
 contract T {
     uint private _myPrivateVar;
+    uint internal _myInternalVar;
     uint _myPublicVar;
```
0xalpharush, line 57:
```suggestion
    uint  public _myPublicVar;
```

0xalpharush: Would you make all of the `_myPublicVar`'s public and run `python tests/e2e/detectors/test_detectors.py --compile` to update the compiled artifacts please? You'll also want to run `pytest tests/e2e/detectors/test_detectors.py --insta update` to update the snapshots as the public variable should be included in the snapshot after updating the solidity code.

0xalpharush (changes requested):

tests/e2e/detectors/test_data/naming-convention/0.4.25/naming_convention.sol
dokzai: Thanks! 

I updated the variable to `public` and updated the snapshots accordingly.

devtooligan:
@dokzai 🫡

Issue #2058, opened 2023-07-19
[Bug]: State variables shadowed by a local variable are incorrectly set to the local variable 


Tiko7454:
### Describe the issue:

A state variable is used before it was shadowed. The dependency goes to the local variable instead of state variable

### Code example to reproduce the issue:

```solidity
// SPDX-License-Identifier: UNLICENSED

pragma solidity 0.8.7;

contract B {
    uint public x = 21;
    function a()  public {
        uint u = 2 * x;  // BUG
        uint x;
    }
}
```

```python
> from slither import Slither
>
>
> a = Slither('b.sol').contracts[0].functions[0]
> buggy_node = a.nodes[1]
> print(buggy_node.variables_read[0].canonical_name)
>> B.a().x
```

### Version:

0.9.3
(tested the freshest dev branch 3e390264a4903c522718b33478e5acb735f3e0e6)

smonicas:
The issue is in early expression parsing because `x` is found by this code as the local variable. Removing this code makes the example works however it will fail if the legacy ast is used and i'm not sure if there are other edge cases that require it.
https://github.com/crytic/slither/blob/e5f2a86f0906fd62c6c4eccb9dbfa5ab30671a78/slither/solc_parsing/expressions/find_variable.py#L83-L86

0xalpharush:
Closed by https://github.com/crytic/slither/pull/2121

PR #2060, opened 2023-07-20
fixed State variables shadowed by a local variable are incorrectly set to the local variable

Tiko7454:
if variables was not seen before, search for it in the state variables

PR #2121, opened 2023-09-08
prioritize reference id in name resolution to avoid shadowing issues

0xalpharush:
replaces https://github.com/crytic/slither/pull/2060
fixes https://github.com/crytic/slither/issues/2058

Prior to 0.5.0 (xref https://github.com/ethereum/solidity/issues/1679), the scoping rules allow references to undeclared variables so shadowing a state variable in a function body would still prioritize the yet-declared local variable. Following 0.5.0, it is no longer possible to have forward references, and the reference will be to the state variable in the case that a shadowing local declaration is declared in predecessors of the same lexical scope.


montyly:
another example of why it would make sense to rewrite a parser for solidity 0.8 only :)

Issue #2083, opened 2023-08-10
[False-Positive]: Consider adding exceptions for i_ and s_ in the naming-conventions detector


devtooligan:
### Describe the false alarm that Slither raise and how you know it's inaccurate:

Slither's `naming-convention` detector strictly picks up any names which are not camelCase or PascalCase.  There is a convention that is becoming increasingly more popular wherby the prefix i_ is used for immutables and s_ is used for state variables.  

Slither should not report those cases for this detector.

### Frequency

Very Frequently

### Code example to reproduce the issue:

```solidity
contract A {
  uint immutable i_myImutableVar = 1;
  uint s_myStateVar;
}
```

### Version:

latest

### Relevant log output:

```shell
`naming-convention (129 results) (Informational)` -- all 129 are FP for this reason
```

PR #2125, opened 2023-09-09
Add exceptions for i_ and s_ in the naming-conventions detector

dokzai:
Hello!

This is my attempt at addressing #2083 with updated test data.

Thanks!


slither/detectors/naming_convention/naming_convention.py
```diff
@@ -167,10 +175,22 @@ def _detect(self) -> List[Output]:
...
 
                 else:
-                    if var.visibility in ["private", "internal"]:
+                    # first priority: check for immutable variable naming conventions
+                    if var.is_immutable:
+                        correct_naming = self.is_immutable_naming(var.name)
+
+                    # second priority: check for state variable naming conventions
+                    elif self.is_state_naming(var.name):
+                        correct_naming = True
```
0xalpharush, line 184: I think these should also be guarded by a check that ensures they're private/immutable as public variables should still be camel case. If this convention is being used, I think public getters with camel cases are created for internal/private variables with the i_/s_ prefix

0xalpharush (changes requested):

devtooligan (approved):
lgtm

montyly (approved):

PR #2153, opened 2023-10-02
Fix inheritance printer rebase

0xPhaze:
This PR nows : 

- Adds a new `include-interfaces` command line switch if the user wants to include interface in the inheritance graph (default is `False`).
- Checks for interfaces before displaying them
- Removes virtual functions
- Adds a test to ensure this works as appropriately

## Summary by CodeRabbit

- **New Features**
  - Added `--include-interfaces` argument to include interfaces in the inheritance-graph printer.

- **Bug Fixes**
  - Excluded contracts with "mock" in their name from the inheritance graph.
  - Removed inherited interfaces when `--include-interfaces` is set to `False`.
  - Excluded virtual functions from the output.

- **Tests**
  - Added tests for include/exclude interface behavior in printer output.


DarkaMaul:
This PR nows : 

- Adds a new `include-interfaces` command line switch if the user wants to include interface in the inheritance graph (default is `False`).
- Checks for interfaces before displaying them
- Removes virtual functions
- Adds a test to ensure this works as appropriately

I think it is ready for review 🚀

PR #2155, opened 2023-10-04
Fix a typo in the help text

SEJeff:
Does what it says on the tin.


montyly:
Good catch. Thanks @SEJeff

Issue #2127, opened 2023-09-12
[Bug]: AssertionError when dealing with unnamed parameters


glarregay-tob:
### Describe the issue:

If you want to print a parameter name, and the parameter is unnamed, you get:
```
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/opt/homebrew/lib/python3.11/site-packages/slither/core/variables/variable.py", line 182, in __str__
    assert self._name
AssertionError
```

### Code example to reproduce the issue:

Contract:
```solidity
//SPDX-License-Identifier: UNLICENSED

pragma solidity 0.8.10;

contract TestSlither {
    function test_function(uint256 param1, uint256, address param3) public {
        
    }
}
```

Script:
```python
from slither import Slither

s = Slither("poc.sol")
c = s.contracts[0]

for f in c.functions:
    for p in f.parameters:
        print(p)
```

### Version:

0.9.6

0xalpharush:
We can change this to be optional and remove the assertion

PR #2158, opened 2023-10-06
Allow function parameter name to be empty when casting to string

dokzai:
Hello!

This is my attempt at addressing #2127  with 1 test added.

Thanks!

### DIFF
--- a/slither/core/variables/variable.py
+++ b/slither/core/variables/variable.py
@@ -179,5 +179,4 @@ class Variable(SourceMapping):
         return f'{name}({",".join(parameters)})'
 
     def __str__(self) -> str:
-        assert self._name
         return self._name
--- a/tests/e2e/compilation/test_resolution.py
+++ b/tests/e2e/compilation/test_resolution.py
@@ -43,3 +43,18 @@ def test_cycle(solc_binary_path) -> None:
     solc_path = solc_binary_path("0.8.0")
     slither = Slither(Path(TEST_DATA_DIR, "test_cyclic_import", "a.sol").as_posix(), solc=solc_path)
     _run_all_detectors(slither)
+
+
+def test_contract_function_parameter(solc_binary_path) -> None:
+    solc_path = solc_binary_path("0.8.0")
+    standard_json = SolcStandardJson()
+    standard_json.add_source_file(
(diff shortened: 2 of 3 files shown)


slither/core/variables/variable.py
```diff
@@ -179,5 +179,4 @@ def solidity_signature(self) -> str:
         return f'{name}({",".join(parameters)})'
 
     def __str__(self) -> str:
```
0xalpharush, line 181:
```suggestion
    def __str__(self) -> Optional[str]:
```

tests/e2e/compilation/test_resolution.py
```diff
@@ -43,3 +43,18 @@ def test_cycle(solc_binary_path) -> None:
...
+    standard_json.add_source_file(
+        Path(TEST_DATA_DIR, "test_contract_data", "test_contract_data.sol").as_posix()
+    )
+    compilation = CryticCompile(standard_json, solc=solc_path)
+    slither = Slither(compilation)
+    contract = slither.contracts[0]
+
+    for function in contract.functions:
+        for parameter in function.parameters:
+            str(parameter)
```
0xalpharush, line 60:
```suggestion
        assert(function.parameters[1].name is None)
```

slither/core/variables/variable.py
0xalpharush: Nvm, the Optional type should apply to the name property

slither/core/variables/variable.py
```diff
@@ -179,5 +179,4 @@ def solidity_signature(self) -> str:
...
-        assert self._name
         return self._name
```
0xalpharush, line 182: Maybe we should return `""` when `self._name is None`

tests/e2e/compilation/test_resolution.py
dokzai: For the test example, the parameter name is set to empty string: https://github.com/crytic/slither/blob/master/slither/solc_parsing/variables/variable_declaration.py#L143

0xalpharush: Ah, so it's not `None` it's an empty string. I guess the assertion failing was only because "" is falsy value. You can disregard my comments

montyly:
This great, thanks @dokzai

Issue #2107, opened 2023-09-01
Slither does not handle multiple contracts with the same name across compilation units


hacker-DOM:
Consider the codebase https://github.com/hacker-DOM/reference-resolving/tree/ecd8bf0e91a3636104b5ffc805537eaac2eeb8d7

These are the results for inheritance-graph and call-graph, both of which are wrong

![image](https://github.com/crytic/slither/assets/18601956/83ad0f67-0172-410f-84c8-b8c14a67e234)

![image](https://github.com/crytic/slither/assets/18601956/fd47b0b8-ea37-420d-a241-ac039e1d8f94)

montyly:
Indeed the printers merge contracts with the same name, which they should not. We will update the printer.

0xalpharush:
We need to make these id's unique across compilation unit
https://github.com/crytic/slither/blob/e5f2a86f0906fd62c6c4eccb9dbfa5ab30671a78/slither/printers/call/call_graph.py#L20-L25
https://github.com/crytic/slither/blob/e5f2a86f0906fd62c6c4eccb9dbfa5ab30671a78/slither/printers/inheritance/inheritance_graph.py#L103-L106

PR #2159, opened 2023-10-07
Update inheritance graph printer to handle multiple contracts with same names 

dokzai:
Hello!

This is my attempt at addressing #2107 with 1 test added. 

The updated inheritance graph printer generates the following graph for the example provided in the issue:

contract_inheritance

The call graph printer already uses contract id to handle multiple contracts with same names: 

callgraph

Thanks!

### DIFF
--- a/tests/e2e/printers/test_printers.py
+++ b/tests/e2e/printers/test_printers.py
@@ -15,24 +15,16 @@ TEST_DATA_DIR = Path(__file__).resolve().parent / "test_data"
 def test_inheritance_printer(solc_binary_path) -> None:
     solc_path = solc_binary_path("0.8.0")
     standard_json = SolcStandardJson()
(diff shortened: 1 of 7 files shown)


tests/e2e/printers/test_printers.py
```diff
@@ -0,0 +1,45 @@
...
+    standard_json.add_source_file(
+        Path(TEST_DATA_DIR, "test_contract_names", "C.sol").as_posix()
+    )
+    compilation = CryticCompile(standard_json, solc=solc_path)
+    slither = Slither(compilation)
+    printer = PrinterInheritanceGraph(slither=slither, logger=None)
+
+    content = ""
+    for contract in slither.contracts:
+        content += printer._summary(contract)
```
dokzai, line 36: Should I convert the test to use printer.output instead?

Issue #1622, opened 2023-01-23
[Bug]: Library call function is not an instance of `Function`


0xalpharush:
### Describe the issue:

Library call function is not an instance of `Function`
Related with this library call here https://github.com/code-423n4/2023-01-timeswap/blob/ef4c84fb8535aad8abd6b67cc45d994337ec4514/packages/v2-option/src/TimeswapV2Option.sol#L112

### Code example to reproduce the issue:

```
git clone https://github.com/code-423n4/2023-01-timeswap
yarn
cd v2-token
slither .
```

### Version:

0.9.2

### Relevant log output:

```shell
File "/Users/alpharush/tob/slither/slither/core/cfg/node.py", line 920, in _find_read_write_call
    assert isinstance(ir.function, Function)
AssertionError
```

0xGusMcCrae:
I tried to figure this out after noticing it in the c4 audit for timeswap.

As far as I can tell, the issue is that v2-option, v2-pool, and v2-token all have ParamLibrary contracts with `check` functions in their respective structs/Param.sol files. When they are used in the same directory, as they are in packages/v2-pool and packages/v2-token, slither mixes up the ParamLibraries.

The issue appears to be with ParamLibrary.check(param), which is only present in v2-pool and v2-token's ParamLibrary. v2-option's ParamLibrary.check functions all have a second parameter, `blockTimestamp`. The assertion error occurs while slither is processing ParamLibrary.check in both TimeswapV2Token.sol and TimeswapV2Pool.sol.

I added some print statements throughout slither to try to debug. 

in slither/slithir/convert.py (around line 100):

    def convert_expression(expression, node):
    # handle standlone expression
    # such as return true;
    from slither.core.cfg.node import NodeType

    print("beginning of convert_expression")
    print(expression)
    try:
        print(expression.called.expression.value.file_scope)
        print([f.name for f in expression.called.expression.value.functions])
    except:
        None

Running slither . in packages/v2-token:

    beginning of convert_expression
    ParamLibrary.check(param)
    ../v2-option/src/structs/Param.sol
    ['check', 'check', 'check', 'check']
    Traceback (most recent call last):
    (error is thrown here)

This is occurring in TimeswapV2Token.mint(TimeswapV2TokenMintParam) (as you've stated above). It's using check(TimeswapV2TokenMintParam memory param) from v2-token's ParamLibrary.

Running slither . in packages/v2-pool gets the same result:

    beginning of convert_expression
[...]

PR #2167, opened 2023-10-12
Scope imports only correct items

montyly:
Fixes https://github.com/crytic/slither/issues/1622

Depends on https://github.com/crytic/slither/pull/2219 (will require updating with TopLevelEvent)


montyly:
We need a bit of further work and testing for this one I think

montyly:
This PR reduces the scoping import to only the ones needed (ex: if you import A from file.sol, dont import everything from file.sol)

PR #2185, opened 2023-10-18
revert https://github.com/crytic/slither/pull/2105

0xalpharush:
Revert https://github.com/crytic/slither/pull/2105 until format is supported by echidna.


ggrieco-tob:
This is already supported by echidna, can we have this PR un-reverted? :smile:

PR #1984, opened 2023-06-22
Dominator deadcode problem fix

Tiko7454:
Fixes #1973 the way I mentioned in the bug report

PR #2201, opened 2023-10-24
added update reachability to modifiers

Tiko7454:
The modifiers also needed to update their node reachability
(issue #1984)


0xalpharush:
Included in https://github.com/crytic/slither/pull/2321

Issue #2190, opened 2023-10-18
[Bug]: Vyper Example Contract Fails - send function


ritzdorf:
### Describe the issue:

I tried the new vyper feature with the vyper example BlindAuction contract (https://docs.vyperlang.org/en/v0.3.7/vyper-by-example.html). I get the error:

```
slither.solc_parsing.exceptions.VariableNotFound: Variable not found: send (context reveal)
```

https://docs.vyperlang.org/en/v0.3.7/built-in-functions.html#send

### Code example to reproduce the issue:

```
# Blind Auction. Adapted to Vyper from [Solidity by Example](https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst#blind-auction-1)

struct Bid:
  blindedBid: bytes32
  deposit: uint256

# Note: because Vyper does not allow for dynamic arrays, we have limited the
# number of bids that can be placed by one address to 128 in this example
MAX_BIDS: constant(int128) = 128

# Event for logging that auction has ended
event AuctionEnded:
    highestBidder: address
    highestBid: uint256

# Auction parameters
beneficiary: public(address)
biddingEnd: public(uint256)
revealEnd: public(uint256)

# Set to true at the end of auction, disallowing any new bids
ended: public(bool)

# Final auction state
highestBid: public(uint256)
highestBidder: public(address)

# State of the bids
bids: HashMap[address, Bid[128]]
bidCounts: HashMap[address, int128]

# Allowed withdrawals of previous bids
pendingReturns: HashMap[address, uint256]


# Create a blinded auction with `_biddingTime` seconds bidding time and
# `_revealTime` seconds reveal time on behalf of the beneficiary address
# `_beneficiary`.
@external
def __init__(_beneficiary: address, _biddingTime: uint256, _revealTime: uint256):
    self.beneficiary = _beneficiary
    self.biddingEnd = block.timestamp + _biddingTime
    self.revealEnd = self.biddingEnd + _revealTime


# Place a blinded bid with:
#
# _blindedBid = keccak256(concat(
#       convert(value, bytes32),
#       convert(fake, bytes32),
#       secret)
# )
#
# The sent ether is only refunded if the bid is correctly revealed in the
# revealing phase. The bid is valid if the ether sent together with the bid is
# at least "value" and "fake" is not true. Setting "fake" to true and sending
# not the exact amount are ways to hide the real bid but still make the
# required deposit. The same address can place multiple bids.
@external
@payable
[...]

PR #2212, opened 2023-11-01
add support for send builtin

0xalpharush:
closes https://github.com/crytic/slither/issues/2190

### DIFF
--- a/slither/core/declarations/solidity_variables.py
+++ b/slither/core/declarations/solidity_variables.py
@@ -114,6 +114,7 @@ SOLIDITY_FUNCTIONS: Dict[str, List[str]] = {
     "_abi_encode()": [],
     "slice()": [],
     "uint2str()": ["string"],
+    "send()": [],
 }
 
 
--- a/slither/vyper_parsing/expressions/expression_parsing.py
+++ b/slither/vyper_parsing/expressions/expression_parsing.py
@@ -153,10 +153,10 @@ def parse_expression(
                 parsed_expr.set_offset(expression.src, caller_context.compilation_unit)
                 return parsed_expr
 
+            # `raw_call` and `send` are treated specially in order to force `extract_tmp_call` to treat this as a `HighLevelCall` which will be converted
+            # to a `LowLevelCall` by `convert_to_low_level`. This is an artifact of the late conversion of Solidity...
             if called.value.name == "raw_call()":
                 args = [parse_expression(a, caller_context) for a in expression.args]
-                # This is treated specially in order to force `extract_tmp_call` to treat this as a `HighLevelCall` which will be converted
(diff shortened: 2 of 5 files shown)


montyly:
I think we should convert `send` to our IR (LowLevel call?) instead of having a specific function

0xalpharush:
This is converting to the same [operation](https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/slithir/operations/send.py#L15) that solidity is:
```
contract C {
    function f() public {
        payable(msg.sender).send(address(this).balance);
    }
}
```
```
        Function C.f() (*)
                Expression: address(msg.sender).send(address(this).balance)
                IRs:
                        TMP_0 = CONVERT msg.sender to address
                        TMP_1 = CONVERT this to address
                        TMP_2(uint256) = SOLIDITY_CALL balance(address)(TMP_1)
                        TMP_3 = SEND dest:TMP_0 value:TMP_2
```

slither/core/declarations/solidity_variables.py
```diff
@@ -114,6 +114,7 @@
     "_abi_encode()": [],
     "slice()": [],
     "uint2str()": ["string"],
+    "send()": [],
```
montyly, line 117: This part is telling me that we are doing something off: 
- On the solidity side, we parse the send as a high level call (https://github.com/crytic/slither/blob/a2d88b80cddb28ab6cae987a8856374973a716ef/slither/slithir/convert.py#L1224)
- but on vyper we use a solidity variable as a placeholder to parse them

It sounds to me that we are creating a disparity in the way we handle these type of operations between solidity and vyper, which will likely create issues in the long term

0xalpharush: https://github.com/crytic/slither/pull/2212/files#diff-b294f0d085108aec57e21453268799188e4dec7295f054523bb2fabafbcdf1d4R156-R157

0xalpharush: I explained that in the comment. There may be another way, but raw_call and send are use this same codepath since we create a MemberAccess which is then converted to a HighLevelCall in `extract_tmp_call` https://github.com/crytic/slither/blob/a2d88b80cddb28ab6cae987a8856374973a716ef/slither/slithir/convert.py#L1113

0xalpharush: In vyper the builtin are `Call` unliked Solidity where they are `FunctionCall(MemberAccess...)` The use of `SolidityFunction` allows us to know we are dealing with builtins to special case here. I guess we could add a LowLevelCallExpression that is only used by Vyper and add new code to support it in the IR conversion step
https://github.com/crytic/slither/blob/a2d88b80cddb28ab6cae987a8856374973a716ef/slither/vyper_parsing/expressions/expression_parsing.py#L109

Issue #2229, opened 2023-11-11
[Bug-Candidate]: --disable-color ignored, printer produces colored outputs


ongrid:
### Describe the issue:

Flag `--disable-color` seem to be ignored and printer produces colored output with ASCII escape characters not suitable to capture into plaintext files

```
slither --help   
usage: slither target [flag]
Additional options:
...
  --disable-color       Disable output colorization
```

Workaround: pass the output through the following sed script:
```
slither . --print function-summary 2>&1 | sed 's/\x1b\[[0-9;]*m//g'
```

### Code example to reproduce the issue:

<img width="1192" alt="image" src="https://github.com/crytic/slither/assets/7992612/850e41d6-e60e-4383-bdb4-c6d6a385c320">

### Version:

slither --version
0.10.0

From docker image `ghcr.io/trailofbits/eth-security-toolbox:nightly`

elopez:
Thanks for the report @ongrid ! It looks like this printer uses [`MyPrettyTable`](https://github.com/crytic/slither/blob/master/slither/utils/myprettytable.py) to print a table, and that class always uses a colored table. Maybe the `MyPrettyTable` underlying implementation should choose to use the non-colored `PrettyTable` class when colors are disabled: https://github.com/jazzband/prettytable#changing-the-appearance-of-your-table---with-colors

theexoticman:
Thanks, @elopez for the pointer. I took the liberty to propose a fix. Let me know if using an environment variable follows the dev best practices guidelines. I'll add a unit test after confirmation.

PR #2230, opened 2023-11-11
Fixing myprettytable ignoring the --disable-color command line argumen…

theexoticman:
…t. Proposing a fix for  #2229.

Using environment variables to set the disable color behavior for myprettytables, instead of passing the color as parameters to every instantiation.


0xalpharush:
Replaced by https://github.com/crytic/slither/pull/2310

Issue #2233, opened 2023-11-16
Add support for filtering printer output


lucas-manuel:
### Describe the desired feature

It would be great to have more control over outputs from printers, specifically table formatted ones such as `function-summary` and `vars-and-auth`.

Using `function-summary` as an example, it would be great to be able to filter the following (regex would be sufficient for all probably):
1. Function names
2. Visibility
3. Modifiers
4. Read
5. Write
6. Internal Calls
7. External Calls

Using regex a user could have full flexibility to filter out based on specifics or based on whether something is empty

"Return all functions without `get*` at the start that are external, have external calls, don't have any modifiers, and modify state."

shortdoom:
regex is not needed, all of the listed information is available to slither as formatted py object (string) in either Function or Contract class. 

i think the Issue could be generalized to better control over **any** output as slither relies extensively on prettyTable for formatting cli display and on the other hand complex json data type for output with --json flag. both of those have drawbacks when integrating with other external to slither scripts. amazingly, as much as Slither allows for easy integration directly with Class objects, it's output pipeline is pretty "hardcoded" from what I've seen

some middleware allowing for specifying custom ouput easily is a good idea (say, simplified-json). either in form of a standalone method accessible from the core (by user or other slither classes) or as a separate util.

devtooligan:
hi @lucas-manuel 👋🏽 

Great idea.

We have talked about making `slither-printer` it's own thing which could take some clargs for filtering as well as output options as @shortdoom  suggests.  Would also be nice to  have a toggle for --fully-derived-contracts only, for example, for certain printers. 

Making `slither-printer` a separate command is not a requirement to this but it would be cleaner than trying to tack on more  clargs to the base `slither` command.

I can work on this when I have some time (soon ™️)  or if someone else wants to take a shot at implementing this we can provide guidance.

shortdoom:
@lucas-manuel @devtooligan @0xalpharush

if you are fine with having this as a separate "tool" and not as a "printer", I can make a PR _today_, just green-light following logic and reasoning behind it:

- printers are built around Contract class and do not accept any args. it's a problem because some of the data needed is residing in the Function class. basically, how printers are structured disallows, without some changes of AbstractPrinter or creating new type of abstract interface, to implement "filtering printer" effectively. hence why the "tool" and not printer. tooling interface allows for multiple arguments and it's more relaxed.

**example use case:**

"Return all functions without get* at the start that are external, have external calls, don't have any modifiers, and modify state."

[...]

PR #2238, opened 2023-11-18
slither-function-filter tool

shortdoom:
reference: https://github.com/crytic/slither/issues/2233

`slither-function-filter` tool. for the time being it supports basic flag-based filtering (no way to specify details of search, only true/false)

two example inputs:

`slither-function-filter contract.sol --visibility external --modifiers --ext-calls` or

`slither-function-filter contract.sol --read-only` etc.

so, it's possible to:

_"Return all functions that are {visibility}, have/don't have internal OR external calls , have/don't have modifiers, are/aren't mutable (involve a potential state change)"_

all allowed flags:

```
--contract-name (str)
--visibility (str)
--modifiers (bool)
--ext-calls (bool)
--int-calls (bool)
--state-change (bool)
--read-only (bool)
```

every flag is optional, if none is specified, all contract functions will be returned

implementation is focused around output of function.get_summary() but filter_functions() is easy to extend.

output is formatted in logger and with coloring.

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ setup(
             "slither-doctor = slither.tools.doctor.__main__:main",
             "slither-documentation = slither.tools.documentation.__main__:main",
             "slither-interface = slither.tools.interface.__main__:main",
+            "slither-function-filter = slither.tools.function_filter.__main__:main",
         ]
     },
 )
--- a/slither/tools/function_filter/__main__.py
+++ b/slither/tools/function_filter/__main__.py
@@ -1,0 +1,109 @@
+import sys
+import logging
+from argparse import ArgumentParser, Namespace
+
+from crytic_compile import cryticparser
+from slither import Slither
+from slither.core.declarations import Function
+from slither.utils.colors import green
+
+logging.basicConfig()
+logging.getLogger("Slither").setLevel(logging.INFO)
+
+
+def parse_args() -> Namespace:
+    """
+    Parse the underlying arguments for the program.
+    :return: Returns the arguments for the program.
+    """
+    parser: ArgumentParser = ArgumentParser(
+        description="FunctionSummarySelection",
+        usage="function_summary_selection.py filename [options]",
+    )
+
+    parser.add_argument(
+        "filename", help="The filename of the contract or truffle directory to analyze."
+    )
+
+    parser.add_argument(
+        "--contract-name", type=str, help="If set, filter functions declared only in that contract."
+    )
+
+    parser.add_argument("--visibility", type=str, help="Visibility of the functions.")
+
+    parser.add_argument(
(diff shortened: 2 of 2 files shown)


0xalpharush:
I had wanted something similar and made a note to myself:

>We should expose a tool for semantic grepping with the ability to target contract/ function names, filter by visibility/purity,
>Proposed CLI:
>
```
slither grep contracts/ [--contract|--function] [--pure,--nonpure] [--external-calls] [--unchecked] [--assembly]
```

I think that this is a decent start but I would imagine showing the [source](https://crytic.github.io/slither/slither/core/source_mapping/source_mapping.html#Source.content) and [line info](https://crytic.github.io/slither/slither/core/source_mapping/source_mapping.html#Source.to_detailed_str) so that it can be quickly examined or jumped to in the editor would be better. Wdyt?

shortdoom:
Some time passed from this PR but I remember that choosing exact `Contract.sol` as target to query is preferable over iterating the whole directory - avoids information overload in the cli output. 

So you won't need this part of the command:

`...contracts/ [--contract|--function]...` just `Contract.sol`

Current version was designed as _query contract to find function_. It's an interesting direction to also allow some grepping at the level of Function itself (has_require, has_loop, etc.).

I noticed you also included  `--unchecked` and `--assembly`, I would also add `--low_lvl_calls`. 

I also agree on adding one more string input flag to allow to _find a specific function_ by `full_name`. 

Showing source code of a Function is trickier UX-wise. Some functions will be long and cli output will get cluttered. Hence why I resorted to displaying structured info (it tends to be more or less the same length). Displaying a line reference is a great idea, most IDEs will just allow to jump automatically and we could avoid displaying source code in terminal.

To have a full _grep like tool_ we could allow to actually grep the function's source code without cluttering the cli though! 

`slither grep Contract.sol --match 'try:'` - that would find all functions using try/catch block ("try:" string) in the function source code.

I can ofc bring this PR home.

shortdoom:
Also, wouldn't it be cool to have it as PythonAPI? This is how I actually implemented it for my own use, as an external class taking a list of Function to filter, returning Function(s) matching. I needed to resort to writing an external query class because there's no API (and this PR also doesn't propose it in the current state)

shortdoom:
Added `lines` reference to the output (allows to Jump-to-line)

Added more filters:

```
--contains-asm (bool)
--low-lvl-calls (bool)
--full-name (str)
--in-source (str)
```

`full-name` allows user to search inside the function.full_name string (not an exact match)

`in-source` will search inside of the functions source code (not an exact match)

[slither-function-filter.webm](https://github.com/crytic/slither/assets/56428630/87cbe581-983c-4e69-9b91-65168d930e0a)


I think that covers most immediate needs we've seen mentioned by developers?

PR #2241, opened 2023-11-22
Consider local vars with storage location as storage vars

smonicas:
When computing the storage/local variables read/written we should consider the local variables with storage location as storage variables read/written instead of local variables.


montyly:
This makes sense to me, but I think we need to add a unit test for this

PR #2250, opened 2023-11-29
Experimental alias import support

montyly:
to note merge, still experimental


0xalpharush:
I think this is redundant with https://github.com/crytic/slither/pull/2167

PR #2251, opened 2023-11-29
Track function natspec comments

smonicas:


montyly:
I think this needs more sync with the API provided by crytic-compile for the natspec.

Ex:
```solidity
% cat test.sol
contract A{

        /// @dev testdev
        /// @notice testnotice
        /// @param a the parameter
	function f(uint a) public {

	}

}
```


```python
slither = Slither("test.sol")

cc = slither.compilation_units[0]
a = cc.contracts[0]

filename = a.source_mapping.filename
natspec_of_a = cc.crytic_compile_compilation_unit.source_unit(filename).natspec["A"]

print(natspec_of_a) # it's a NatSpec object https://github.com/crytic/crytic-compile/blob/master/crytic_compile/utils/natspec.py#L222

print(natspec_of_a.userdoc.methods["f(uint256)"].export())

print(natspec_of_a.devdoc.methods["f(uint256)"].export())
```

We probably need something unified here, that can return the Natspec object (structured), and potential comments that dont follow natspec

PR #2281, opened 2024-01-26
Add support Solidity 0.8.24

smonicas:
Add the support of Solidity 0.8.24 and BASEFEE for Yul


0xalpharush:
Can we add parsing tests please

PR #1781, opened 2023-03-20
Test/slithir read write

0xalpharush:
-[ ] Add helper function to add `_vars_read` which checks `_is_non_slithir_var`

PR #2288, opened 2024-02-04
Fixed data dependencies for `ReferenceVariable`

YanhuiJessica:
In an `Index` operation, `ir.lvalue` depends on `ir.variable_left` and `ir.variable_right`.

### DIFF
--- a/examples/scripts/data_dependency.py
+++ b/examples/scripts/data_dependency.py
@@ -126,3 +126,19 @@ print(f"{var_tainted} is tainted: {is_tainted(var_tainted, contract)}")
 assert is_tainted(var_tainted, contract)
 print(f"{var_not_tainted} is tainted: {is_tainted(var_not_tainted, contract)}")
 assert not is_tainted(var_not_tainted, contract)
+
+print("Index contract")
+contracts = slither.get_contract_from_name("Index")
+assert len(contracts) == 1
+contract = contracts[0]
+ref = contract.get_state_variable_from_name("ref")
+assert ref
+mapping_var = contract.get_state_variable_from_name("mapping_var")
+assert mapping_var
+
+print(f"{ref} is dependent of {mapping_var}: {is_dependent(ref, mapping_var, contract)}")
+assert is_dependent(ref, mapping_var, contract)
+print(f"{ref} is dependent of {msgsender}: {is_dependent(ref, msgsender, contract)}")
+assert is_dependent(ref, msgsender, contract)
+print(f"{mapping_var} is dependent of {msgsender}: {is_dependent(mapping_var, msgsender, contract)}")
+assert not is_dependent(mapping_var, msgsender, contract)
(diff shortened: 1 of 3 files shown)


0xalpharush:
I think it makes sense to include both the lefthand and righthand side, but I'd prefer to to do it by modifying what is considered "read" (in the `read` attribute of the IR operation) as was done in an unmerged fix for this issue (https://github.com/crytic/slither/pull/1781/commits/5083c82e0736dc264173e2c2cdbe83b16af94c4b). Then, we can just taint `ir.read` as is done for the operations and not special case it in the data dependency.

YanhuiJessica:
```py
if isinstance(ir, OperationWithLValue) and ir.lvalue:
    if isinstance(ir.lvalue, LocalIRVariable) and ir.lvalue.is_storage:
        continue
    if isinstance(ir.lvalue, ReferenceVariable):
        lvalue = ir.lvalue.points_to
        if lvalue:
            add_dependency(lvalue, function, ir, is_protected)
    add_dependency(ir.lvalue, function, ir, is_protected)
```
For an `Index` operation, `add_dependency()` will be called twice, one for `ir.lvalue.points_to` and one for `ir.lvalue`. They share `ir.read`. I still think it's a special case in the data dependency :)

0xalpharush:
Sorry, I miss understood and this is a separate issue. `Index` does consider both `ir.variable_left` and `ir.variable_right` as `read` [here](https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/slithir/operations/index.py#L24-L29), so I think we can accomplish the same thing by changing this [code](https://github.com/crytic/slither/blob/d354952e837e481f7e180ba63dd0cf525b56616b/slither/analyses/data_dependency/data_dependency.py#L421-L428) in `add_dependency` to the following:     
```
    if isinstance(ir, InternalCall) and ir.function:
        read = ir.function.return_values_ssa
    else:
        read = ir.read
```

YanhuiJessica:
I have simplified the [condition](https://github.com/crytic/slither/pull/2288/commits/fa1aff524b3abc48a843c6e0c1924a8435be16eb).
We can not just use `ir.read`. Consider the following `Index` operation:
```
REF_1(uint256) -> mapping_var_1[msg.sender]
```
`add_dependency(lvalue,function,ir)` will be called with `(ir.lvalue.points_to,function,ir)` and `(ir.lvalue,function,ir)`, i.e. `(mapping_var_1,function,ir)` and `(REF_1,function,ir)`.
`ir.read` is `mapping_var_1` and `msg.sender`, but `mapping_var_1` should not depend on `msg.sender`.

0xalpharush:
Thanks for the example as it helped me understand the changes. As I understand, previously the data dependency would not consider `ref` tainted by `msg.sender` but now it will with these changes for the following:
```
ref = mapping[msg.sender]
```
I think this makes sense, but I want to double check with @montyly considering this may be a purposeful under-approximation to prevent additional false positives in detectors that use the data dependency analysis.

PR #2302, opened 2024-02-12
slither-mutate: fit and finish

bohendo:
This PR takes care of a handful of random quality-of-life improvements that will make slither-mutate as useful as possible out-of-the box during code reviews. So far, improvements include:

- count valid/total revert, comment, and tweak mutants separately
- exclude any mutants that fail to compile from the cumulative counts
- only replace a line with revert if it's not already a revert
- exclude interfaces from mutation campaign
- add very-verbose logs with useful debugging information
- tidy up output logs for better at-a-glance comprehension

One maybe-controversial change is a rebrand from reporting "in/valid" mutants to reporting "un/caught" mutants. The word "valid" sounds like a good thing to most people, but more valid mutants means worse test coverage, which is bad (& often confuses me). In contrast, "uncaught" (previously "valid") sounds to the untrained ear like something was missed, which it was. While "caught" (previously "invalid") sounds like tests did their job, which they did. This also frees up the term "invalid mutant" to refer to mutants which do not compile, which makes more sense to me. Note that this new terminology does not follow the standards established in academic literature. But, it should make it easier for newcomers to wrap their head around what's going on. A worthwhile tradeoff IMO but lmk if you disagree.

Some additional features I'd like to include in this PR before we merge:
- [x] exit immediately when we get a ctrl-c interrupt, don't just stop the current test run & move on to the next mutant
- [x] when we exit after a ctrl-c interrupt, remove mutated files and restore the originals
- [x] run the tests once before commencing the mutation testing campaign, abort if tests don't pass
- [x] During the above, time how long the tests take normally. If no timeout specified, default to 2x as long as the tests usually take.
- [x] make `--quick` the default (skip tweak mutants if revert/comment mutants in the same expression are uncaught), replace it with a `--comprehensive` flag that will test all mutants (current default)


bohendo:
Pylint failure is due to the [mutate function](https://github.com/crytic/slither/blob/mutator/fit-and-finish/slither/tools/mutator/mutators/abstract_mutator.py#L77) which has branches for adding up different mutant types. Could use a switch, but that hardly cleans things up. Should I ignore this warning? Or do you all have any recommendations for refactoring it?

0xalpharush:
Probably fine to ignore at the function level

bohendo:
This PR is done & ready for a final review

Issue #1644, opened 2023-02-02
create a variable API that filters out constants and immutables


0xalpharush:
There's discussion about changing code data location keywords and it would be convenient/ more robust to have one API rather than filtering state_variables.

For example, this code could use the new API https://github.com/crytic/slither/pull/1451

Ref https://github.com/ethereum/solidity/issues/13723

dokzai:
I would like to help out with this issue. Can you share more information?

0xalpharush:
There's several places where code like `contract.state_variables_ordered if not (variable.is_constant or variable.is_immutable)` exists. We could an API that only returns variables that have a storage location instead of filtering everywhere and having to update if a new non-storage keyword is add e.g. `codedata` in place of `constant` and `immutable` https://crytic.github.io/slither/slither/core/declarations/contract.html#Contract.state_variables_ordered

0xalpharush:
For isntance 
https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/detectors/variables/unchanged_state_variables.py#L95-L100
https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/printers/summary/variable_order.py#L32
https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/tools/upgradeability/checks/variable_initialization.py#L46-L47
https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/tools/read_storage/read_storage.py#L401
https://github.com/crytic/slither/blob/e3dcf1ecd3e9de60da046de471c5663ab637993a/slither/core/compilation_unit.py#L299-L300

dokzai:
Hi @0xalpharush 

https://github.com/dokzai/slither/blob/4f4acae1af3e4c68555972ea09afe686427d20db/slither/core/declarations/contract.py#L440-L464

Please let me know if the following changes are what you had in mind for the new API.

Thanks!

0xalpharush:
Looks good! Go ahead and open a PR please

PR #2323, opened 2024-02-20
Create a variable API that filters out constants and immutables

dokzai:
The following changes were added to address #1644: 
- `Contract.stored_state_variables`: Filters and returns state variables with storage, excluding constant or immutable ones. 
- `Contract.stored_state_variables_ordered`: Returns stored state variables in their declaration order, filtering out constants or immutables.
- `Variable.is_stored`: Determines if a variable is stored, based on it not being constant or immutable, ready for future updates.
- Updated code to use the new methods

### DIFF
--- a/slither/core/compilation_unit.py
+++ b/slither/core/compilation_unit.py
@@ -302,10 +302,7 @@ class SlitherCompilationUnit(Context):
 
             slot = 0
             offset = 0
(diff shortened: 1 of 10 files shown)


0xalpharush:
Thanks @dokzai! This is a lot cleaner

PR #2333, opened 2024-02-21
Add virtual and override attribute in Function

### LINKED ISSUE #2327: expose virtual and override attributes from AST in solidity `Function`

This is missing and would be useful e.g. we can remove virtual functions that aren't overridden from the inheritance graph

smonicas:
Closes #2327


0xalpharush:
Can we add tests?

PR #2349, opened 2024-03-02
add known solc bugs, remove allowed versions, improve formatting

0xalpharush:
closes https://github.com/crytic/slither/issues/1593

Also, we will no longer explicitly endorse an allowed version beyond using solc 0.8.0 for checked arithmetic and other old bug fixes. Instead just warn if a bug is disclosed.


slither/detectors/attributes/incorrect_solc.py
```diff
@@ -46,64 +47,36 @@ class IncorrectSolc(AbstractDetector):
...
     COMPLEX_PRAGMA_TXT = "is too complex"
-    OLD_VERSION_TXT = "allows old versions"
+    OLD_VERSION_TXT = (
+        "is an outdated solc version. Use a more recent version (at least 0.8.0), if possible."
+    )
     LESS_THAN_TXT = "uses lesser than"
 
     BUGGY_VERSION_TXT = (
-        "is known to contain severe issues (https://solidity.readthedocs.io/en/latest/bugs.html)"
+        "contain known severe issues (https://solidity.readthedocs.io/en/latest/bugs.html)"
```
0xalpharush, line 63:
```suggestion
        "contains known severe issues (https://solidity.readthedocs.io/en/latest/bugs.html)"
```

PR #2357, opened 2024-03-08
slither-mutate: bugfix when two files have the same name

DarkaMaul:
This PR provides several bugfixes on top of #2302:

- If to files were backuped / mutated with the same name (e.g. `path/Token.sol` and `path2/Token.sol`).
- If the mutation directory was not set to  `mutation_campaign`, the mutants were not written in the appropriate directory
- If the `contracts-name` parameter was used, contracts were not properly filtered
- Contracts could be mutated more than once

On the quality side, it also try to replace most of the `os` module usage with `Pathlib` for additional clarity.

### DIFF
--- a/slither/tools/mutator/__main__.py
+++ b/slither/tools/mutator/__main__.py
@@ -148,7 +148,7 @@ class ListMutators(argparse.Action):  # pylint: disable=too-few-public-methods
 ###################################################################################
 
 
-def main() -> (None):  # pylint: disable=too-many-statements,too-many-branches,too-many-locals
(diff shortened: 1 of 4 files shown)


0xalpharush:
@DarkaMau Thanks for your interesting in contributing. l Would you sign the CLI please?

bohendo:
This is great, this fixes an issue I hadn't encountered organically yet but is important. Plus, the other bits of code clean up are things I'd been thinking to do anyway.
Love it. CLA signature aside, this is ready to merge this into the fit-and-finish branch IMO

PR #2367, opened 2024-03-19
Fix IR for top level functions with using-for

smonicas:
Previously when an item in a top level function with using-for the IR was an `HIGH_LEVEL_CALL` instead of `INTERNAL_CALL`/`LIBRARY_CALL`. This was happening because in the conversion we only looked at the using-for for `FunctionContract`.
```solidity
pragma solidity 0.8.24;

library Lib {
 function a(uint q) public {}
}
function c(uint z)  {}

using {Lib.a} for uint;
using {c} for uint;
function b(uint y)  {
 Lib.a(4);
 y.c();
 y.a();
}
```
Before
```
Top level functions
	Function c(uint256)
	Function b(uint256)
		Expression: Lib.a(4)
		IRs:
			LIBRARY_CALL, dest:Lib, function:Lib.a(uint256), arguments:['4'] 
		Expression: y.c()
		IRs:
			TMP_1(None) = HIGH_LEVEL_CALL, dest:y(uint256), function:c, arguments:[]  
		Expression: y.a()
		IRs:
			TMP_2(None) = HIGH_LEVEL_CALL, dest:y(uint256), function:a, arguments:[]  
```
After
```
Top level functions
	Function c(uint256)
	Function b(uint256)
		Expression: Lib.a(4)
		IRs:
			LIBRARY_CALL, dest:Lib, function:Lib.a(uint256), arguments:['4'] 
		Expression: y.c()
		IRs:
			INTERNAL_CALL, c(uint256)(y)
		Expression: y.a()
		IRs:
			LIBRARY_CALL, dest:Lib, function:Lib.a(uint256), arguments:['y'] 
```


slither/core/declarations/function_top_level.py
```diff
@@ -1,26 +1,53 @@
...
 
+USING_FOR_KEY = Union[str, Type]
+USING_FOR_ITEM = List[Union[Type, Function]]
+
 
 class FunctionTopLevel(Function, TopLevel):
     def __init__(self, compilation_unit: "SlitherCompilationUnit", scope: "FileScope") -> None:
         super().__init__(compilation_unit)
         self._scope: "FileScope" = scope
+        self._using_for_complete: Optional[Dict[USING_FOR_KEY, USING_FOR_ITEM]] = None
```
0xalpharush, line 23: Can we define a type for `Dict[USING_FOR_KEY, USING_FOR_ITEM]` and replace the current uses

slither/core/declarations/function_top_level.py
```diff
@@ -1,26 +1,53 @@
...
 
     @property
     def file_scope(self) -> "FileScope":
         return self._scope
 
+    @property
+    def using_for_complete(self) -> Dict[USING_FOR_KEY, USING_FOR_ITEM]:
+        """
+        Dict[Union[str, Type], List[Type]]: Dict of top level directive
```
0xalpharush, line 32: This type doesn't match in the comment and the return signature is different than the definition. If we return an empty dict and not `None`, it isn't `Optional`

slither/slithir/convert.py
```diff
@@ -1209,7 +1211,7 @@ def extract_tmp_call(ins: TmpCall, contract: Optional[Contract]) -> Union[Call,
         internalcall.set_expression(ins.expression)
         return internalcall
 
-    raise Exception(f"Not extracted {type(ins.called)} {ins}")
+    raise Exception(f"Not extracted {type(ins.called)} {ins}") # pylint: disable=bad-option-value
```
0xalpharush, line 1214: This is fixed in `dev` if you'll merge it into your branch

0xalpharush:
Can we add the PR example as a regression test please?

Issue #2309, opened 2024-02-15
[Bug]: Slither cannot process file aliases


Tiko7454:
### Describe the issue:

When I aliased a file, I cannot access its contracts using `alias.contract` notation

### Code example to reproduce the issue:

```solidity
# a.sol
pragma solidity 0.7.6;
import "b.sol" as B;
contract A is B.C{}
```
```solidity
# b.sol
pragma solidity 0.7.6;
contract C{}
```

### Version:

0.10.0

### Relevant log output:

```shell
'solc --version' running
'solc a.sol --combined-json abi,ast,bin,bin-runtime,srcmap,srcmap-runtime,userdoc,devdoc,hashes,compact-format --allow-paths .,/home/tigran/temp/reduced' running
Compilation warnings/errors on a.sol:
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> a.sol

Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> b.sol

Warning: Source file does not specify required compiler version! Consider adding "pragma solidity ^0.7.6;"
--> b.sol

Warning: Source file does not specify required compiler version! Consider adding "pragma solidity ^0.7.6;"
--> a.sol


Traceback (most recent call last):
  File "/home/tigran/Projects/integrator/query-engine/venv/bin/slither", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/tigran/Projects/integrator/query-engine/venv/lib/python3.11/site-packages/slither/__main__.py", line 727, in main
    main_impl(all_detector_classes=detectors, all_printer_classes=printers)
  File "/home/tigran/Projects/integrator/query-engine/venv/lib/python3.11/site-packages/slither/__main__.py", line 833, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #2274, opened 2024-01-15
[Bug]: `import as` unsupported


matmilbury:
### Describe the issue:

Slither can’t create an IR when a smart contract with renamed import (`import { X as Y } from 'xyz.sol';`).

Importing without the rename 'fixes' the issue.

### Code example to reproduce the issue:

```sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.23;

contract Bar {
    constructor() { }
}
```

```sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.23;

import { Bar as Bar2 } from "./Bar.sol";

contract Foo {
    constructor() {
        new Bar2();
    }
}
```

### Version:

0.10.0

0xalpharush:
Thanks for your interest in Slither. This is a known issue that will take a big lift to properly support https://github.com/crytic/slither/issues/1452

Issue #1594, opened 2023-01-11
AssertionError when setting type for `UserDefinedType`


FredCoen:
### Describe the issue:

After installing the latest slither version 0.9.2 I try to run `slither .` against my foundry project. unfrotunately I get the following trace below. Unfortunately the repo is currently closed source. Can I provide any more information?

Code example to reproduce the issue:
its a pretty standard foundry repo that cant be shared currently

Version:
0.9.2

Relevant log output:
```
'forge clean' running (wd: /Users/friedrichcoen/priv-contracts-v2)
'forge build --build-info --force' running
Compiling 196 files with 0.8.13
Solc 0.8.13 finished in 64.61s
Compiler run successful
Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.9/site-packages/slither/__main__.py", line 834, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/opt/homebrew/lib/python3.9/site-packages/slither/__main__.py", line 98, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/opt/homebrew/lib/python3.9/site-packages/slither/__main__.py", line 76, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/opt/homebrew/lib/python3.9/site-packages/slither/slither.py", line 133, in init
    self._init_parsing_and_analyses(kwargs.get("skip_analyze", False))
  File "/opt/homebrew/lib/python3.9/site-packages/slither/slither.py", line 153, in _init_parsing_and_analyses
    raise e
  File "/opt/homebrew/lib/python3.9/site-packages/slither/slither.py", line 149, in _init_parsing_and_analyses
    parser.analyze_contracts()
  File "/opt/homebrew/lib/python3.9/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 529, in analyze_contracts
    self._convert_to_slithir()
  File "/opt/homebrew/lib/python3.9/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 740, in _convert_to_slithir
    func.generate_slithir_and_analyze()
  File "/opt/homebrew/lib/python3.9/site-packages/slither/core/declarations/function.py", line 1750, in generate_slithir_and_analyze
    node.slithir_generation()
  File "/opt/homebrew/lib/python3.9/site-packages/slither/core/cfg/node.py", line 719, in slithir_generation
    self._irs = convert_expression(expression, self)
[...]

Issue #1452, opened 2022-11-04
Tracking: support for imports with alias (either symbol alias or unit alias)


SheldonHolmgren:
### Describe the issue:

See the linked repo with a minimal reproducible example. Running `slither . --compile-force-framework foundry` leads to the crash with a backtrace pasted as the log output below. This is because Slither cannot find a contract named `c.Counter` (as referenced from `src/Importer.sol`).

If I edit Importer.sol in the following way:
```diff
diff --git a/src/Importer.sol b/src/Importer.sol
index 316a7e3..5fbc2fe 100644
--- a/src/Importer.sol
+++ b/src/Importer.sol
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.13;
 
-import "src/Counter.sol" as c;
+import "src/Counter.sol";
 
 contract Importer {
     constructor() {
-        new c.Counter();
+        new Counter();
     }
 }
```
slither runs without problem.

### Code example to reproduce the issue:

https://github.com/SheldonHolmgren/slither_bug_example

### Version:

0.9.0

### Relevant log output:

```shell
'forge build --extra-output abi --extra-output userdoc --extra-output devdoc --extra-output evm.methodIdentifiers --force' running
Compiling 19 files with 0.8.16
Solc 0.8.16 finished in 1.52s
Compiler run successful

Traceback (most recent call last):
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/__main__.py", line 826, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/__main__.py", line 97, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/__main__.py", line 75, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/slither.py", line 130, in __init__
    parser.analyze_contracts()
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 503, in analyze_contracts
    self._convert_to_slithir()
  File "/home/holmgren/.local/lib/python3.10/site-packages/slither/solc_parsing/slither_compilation_unit_solc.py", line 688, in _convert_to_slithir
[...]

PR #2370, opened 2024-03-20
fix: support aliases for NewContract operation

0xalpharush:
- closes https://github.com/crytic/slither/issues/2309 
- closes https://github.com/crytic/slither/issues/2274
- closes https://github.com/crytic/slither/issues/1594

This also works toward https://github.com/crytic/slither/issues/1452 but does not complete it


0xalpharush:
the lint issue is fixed in https://github.com/crytic/slither/pull/2371

PR #2384, opened 2024-03-29
Update PR#2034

0xalpharush:
replaces https://github.com/crytic/slither/pull/2034


0xalpharush:
The lint is fixed in https://github.com/crytic/slither/pull/2383

Issue #2351, opened 2024-03-02
remove deprecated flags for next breaking release


0xalpharush:
https://github.com/crytic/slither/blob/f7238c4fdf969d7ca402ae0d9019293fd2663270/slither/utils/command_line.py#L78

0xalpharush:
https://github.com/crytic/slither/blob/f7238c4fdf969d7ca402ae0d9019293fd2663270/slither/utils/command_line.py#L101-L106

PR #2410, opened 2024-04-08
Remove deprecated flags and their migration.

DarkaMaul:
This PR solves #2351 

It removes the deprecated flags, the migration handler and their usage in some test file.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the configuration file reading process by removing deprecated flags and related migration logic.
- **Tests**
	- Adjusted end-to-end tests to reflect changes in failure conditions and path filtering criteria in configuration settings.

### DIFF
--- a/tests/e2e/config/test_path_filtering/slither.config.json
+++ b/tests/e2e/config/test_path_filtering/slither.config.json
@@ -2,6 +2,6 @@
     "detectors_to_run": "all",
     "exclude_informational": true,
     "exclude_low": true,
-    "fail_pedantic": false,
+    "fail_on": "none",
     "filter_paths": "libs|src/ReentrancyMock.sol"
 }
--- a/slither/utils/command_line.py
+++ b/slither/utils/command_line.py
@@ -75,13 +75,6 @@ defaults_flag_in_config = {
     **DEFAULTS_FLAG_IN_CONFIG_CRYTIC_COMPILE,
 }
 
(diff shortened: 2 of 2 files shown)


tests/e2e/config/test_path_filtering/slither.config.json
```diff
@@ -2,6 +2,6 @@
     "detectors_to_run": "all",
     "exclude_informational": true,
     "exclude_low": true,
-    "fail_pedantic": false,
+    "fail_on": "none",
```
0xalpharush, line 5: I think this is different than the current default and may mess up CI usage 
cc @elopez

elopez: this is config for a test though, right? it's not a default

DarkaMaul:
```python

    if key.startswith("fail_") and getattr(args, "fail_on") == defaults_flag_in_config["fail_on"]:
        if key == "fail_pedantic":
            pedantic_setting = elem
            fail_on = FailOnLevel.PEDANTIC if pedantic_setting else FailOnLevel.NONE
            setattr(args, "fail_on", fail_on)
            logger.info(f"Migrating fail_pedantic: {pedantic_setting} as fail_on: {fail_on.value}")
```

The original migration was like this so I think `fail_pedantic: False` was migrated to `fail_on: None`

PR #2411, opened 2024-04-08
Add vyper to dev dependencies to run tests

DarkaMaul:
When running tests on local computer using `make test`, some fails with the following error:  

```
...
ERROR tests/e2e/vyper_parsing/test_ast_parsing.py - FileNotFoundError: [Errno 2] No such file or directory: 'vyper'
...
```

This PR adds `vyper` to the dev dependencies and remove the GitHub Workflow to install it.

## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflow to enhance test execution efficiency.
- **New Features**
	- Integrated Vyper into the project dependencies for improved functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
             "coverage[toml]",
             "filelock",
             "pytest-insta",
+            "vyper",
         ],
         "doc": [
             "pdoc",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,23 +57,6 @@ jobs:
             npm install hardhat
             popd || exit
           fi
-      - name: Install Vyper
-        run: |
-          INSTALLDIR="$RUNNER_TEMP/vyper-install"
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -36,6 +36,7 @@
             "coverage[toml]",
             "filelock",
             "pytest-insta",
+            "vyper",
```
0xalpharush, line 39: We should pin this to 0.3.7

DarkaMaul: In this case, we loose compatibility with Python 3.12+ : 

https://github.com/vyperlang/vyper/blob/6020b8bbf66b062d299d87bc7e4eddc4c9d1c157/setup.py#L92

0xalpharush: This is partly why we were using a binary so we aren't constrained by their packages/version

DarkaMaul: I see! Lets close this PR and just update the wiki to precise that running the tests requires to install `vyper` (and `ganache`).

PR #2416, opened 2024-04-09
Use snapshot based tests for printers

DarkaMaul:
Right now, the printers are only run in the CI but their results is not checked.

This PR moves the system to use snapshot based results, like it is done for detectors.

# Remaining TODO
- [ ] Review the known failures
- [ ] Generate and commit snapshot files
- [ ] Review the generated snapshot files to assert they do not contain errors.


DarkaMaul:
@0xalpharush any idea on how to handle the EVM CFG printer ?

Should we keep a test in the CI or add the missing dependency to the new test case ?

0xalpharush:
Let's just keep the EVM printer in the existing bash script

Issue #2398, opened 2024-04-04
API change: rename sons/father to next/predecessor in nodes


montyly:
I think next/predecessor will be more intuitive than sons/father for the CFG (https://github.com/crytic/slither/blob/master/slither/core/cfg/node.py)

This will be a breaking change, but most likely not too many external code relies on this. 

I would be to keep in parallel keep `sons` and `fathers` for a few iteration but making it clear that they are deprecated

PR #2423, opened 2024-04-12
Rename father/son to predecessor/successor

DarkaMaul:
This is a breaking change in the API but for the moment, we keep the compatibility layer in the cfg/node.py.

Fixes https://github.com/crytic/slither/issues/2398

The main changes are in `node.py` where we use a `__getattr__` to proxy calls to `son` or `father` methods (with an explicit list), print a deprecation warning and forward to the new implementation.  

I expect the linter to break, but I tried to keep changes minimal.
Of note, this PR is quite large because it tries to replace usages of `son` or `father` variable in common places in the code.

### DIFF
--- a/slither/core/cfg/node.py
+++ b/slither/core/cfg/node.py
@@ -2,6 +2,7 @@
     Node module
 """
 from enum import Enum
+import logging
 from typing import Optional, List, Set, Dict, Tuple, Union, TYPE_CHECKING
 
 from slither.all_exceptions import SlitherException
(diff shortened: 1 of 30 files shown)


slither/core/cfg/node.py
```diff
@@ -1046,6 +1049,43 @@ def __str__(self) -> str:
         txt = str(self._node_type.value) + additional_info
         return txt
 
+    def __getattr__(self, item: str):
```
montyly, line 1052: We should avoid this, and do a breaking change

Doing the __getattr__ trick will lead to more issues in the long term I think

montyly (changes requested):

slither/core/cfg/node.py
DarkaMaul: I've attempted to keep the compatibility layer minimal here to avoid future issues - however, I believe that implementing a complete breaking change will significantly simplify the code.

PR #2428, opened 2024-04-17
Filtering

DarkaMaul:
Adds three new options to the Slither CLI to filter results.

This PR adds granularity in the filtering system provided by slither on which files, contracts or function to run.

# Help

```shell
➜ slither --help

...

Filtering:
  The following options allow to control which files will be analyzed by Slither. While the initial steps (parsing, generating the IR) are done on the full project, the
  target for the detectors and/or printers can be restricted using these options. The filtering allow to to select directories, files, contract and functions. Each part
  is compiled as a regex (so A*\.sol) will match every file that starts with A and ends with .sol. 
Examples :
 - `sub1/A.sol:A.constructor` will analyze only the
  function named constructor in the contract A found in file A.sol a directory sub1. 
- `sub1/.*` will analyze all files found in the directory sub1/ 
- `.*:A` will analyze all the contract named A

  --include INCLUDE     Include directory/files/contract/functions and only run the analysis on the specified elements.
  --remove REMOVE       Exclude directory/files/contract/functions and only run the analysis on the specified elements.
  --filter FILTER       Include/Exclude directory/files/contract/functions and only run the analysis on the specified elements. Prefix by +_(or nothing) to include, and by - to exclude.

```
 
# Implementation details

1. Migrate `filter-paths`, `include-paths` option to the new system and add a deprecation warning

The previous options are kept for backward compatibility but they are removed from the help output
They are migrated to the new filtering system.

2. Filtering

The filtering is done at the `CompilationUnit` level, the `contracts`, and `functions` method now only return the restricted results instead of the complete results. 
This way the impact on current detectors and/or printers is kept minimal.

3. Filtering syntax

We use a comma separted list of values with this format : 
```shell
<+-><path/subdir/file>:<contract>.<function>
```
The leading +- can be used for the `filter` option to list positive and negative rules.

Each element is compiled as a regex so the following examples work:

- `.*:A.*`: Matches every contract that contains an A
- `+sub1/,-sub1/sub12` : Matches everything in sub1 that is not in sub2.

Of note, the exact regex is: 
```python
    regex = re.compile(
        r"^(?P<type>[+-])?(?P<path>[^:]+?)(?::(?P<contract>[^.])(?:\.(?P<function>.+)?)?)?$"
    )
```

And the tests were made using [here](https://regex101.com/r/NPJlgh/1)

### DIFF
--- a/slither/__main__.py
+++ b/slither/__main__.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import pstats
+import re
 import sys
 import traceback
 from importlib import metadata
@@ -19,6 +20,7 @@ from crytic_compile.platform.standard import generate_standard_export
 from crytic_compile.platform.etherscan import SUPPORTED_NETWORK
 from crytic_compile import compile_all, is_supported
 
+from slither.core.filtering import FilteringRule, FilteringAction
 from slither.detectors import all_detectors
 from slither.detectors.abstract_detector import AbstractDetector, DetectorClassification
 from slither.printers import all_printers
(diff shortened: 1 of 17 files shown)


DarkaMaul:
Some note : 

The filtering system improve the overall performances because the printers / detectors are not run on code before being filtered out. However, some properties could be cached to further improve performance. This will be tacked in another PR with a better focus on performance.

0xalpharush:
I think for directory/file we may be able to also avoid lowering the AST to SlithIR by filtering out them here:
https://github.com/crytic/slither/blob/fdf54f624d8c468afb3d8197eb9e98e377dac440/slither/slither.py#L151-L170

Need to keep in mind the references to other code may necessitate its analysis even if it's filtered

PR #2438, opened 2024-04-22
Features/perf improvment

DarkaMaul:
Slither feels kind of slow sometimes, so this PR tries to tackle some bottlenecks observed during profiling.

Of note, I refrain from `concurrent.futures` and parallel processing for the moment, I think they are some nice quick wins to gain first.

# Comparison
## Without the PR
```
INFO:Slither:/Users/dm/Projects/openzeppelin-contracts analyzed (275 contracts)
slither --print  --detect all ~/Projects/openzeppelin-contracts  109,40s user 5,53s system 78% cpu 2:26,17 total
```
## With the PR
```
INFO:Slither:/Users/dm/Projects/openzeppelin-contracts analyzed (275 contracts)
slither --print  --detect all ~/Projects/openzeppelin-contracts  74,72s user 5,38s system 76% cpu 1:44,56 total
```

Overall, **a 30% ** performance improvment gain.

# Changes

- Using `isinstance` for ABC metaclass is extremely slow, so we reduce their usage
- As `hash` functions are computed heavily, simply their computation.


slither/core/declarations/function.py
```diff
@@ -94,14 +94,11 @@ class FunctionType(Enum):
...
     ret = []
+
     for expression in expressions:
-        if isinstance(expression, Identifier):
-            ret.append(expression)
-        if isinstance(expression, UnaryOperation):
-            ret.append(expression.expression)
-        if isinstance(expression, MemberAccess):
+        if isinstance(expression, (Identifier, UnaryOperation, MemberAccess)):
             ret.append(expression.expression)
```
0xalpharush, line 100: For `Identifier`, before it was:
```
        if isinstance(expression, Identifier):
            ret.append(expression)
```
Now it is:
```
        if isinstance(expression, (Identifier, UnaryOperation, MemberAccess)):
            ret.append(expression.expression)
```
Idk if it matters but was thing purposeful?

slither/core/expressions/identifier.py
```diff
@@ -78,3 +78,6 @@ def value(
 
     def __str__(self) -> str:
         return str(self._value)
+
+    def expression(self):
+        return self
```
elopez, line 83: Should this be decorated as `@property`?
```suggestion

    @property
    def expression(self):
        return self
```

DarkaMaul: Good catch, I missed that!

slither/core/declarations/function.py
DarkaMaul: The idea was to reduce the number of `isinstance` calls from 3 to 1, and add the property on `Identifier`.

PR #2442, opened 2024-04-23
Fix regex patterns

DarkaMaul:
To properly consider the identifiers, we need to use `()` instead of `[]` in regex definition.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the regular expression pattern in the parsing function to ensure accurate keyword matching.

- **Improvements**
  - Updated the regular expression in the parsing function to include a non-capturing group for better type identification.

### DIFF
--- a/slither/solc_parsing/expressions/expression_parsing.py
+++ b/slither/solc_parsing/expressions/expression_parsing.py
@@ -501,7 +501,7 @@ def parse_expression(expression: Dict, caller_context: CallerContextExpression) 
                 referenced_declaration = expression["attributes"]["referencedDeclaration"]
 
         if t:
-            found = re.findall(r"[struct|enum|function|modifier] \(([\[\] ()a-zA-Z0-9\.,_]*)\)", t)
+            found = re.findall(r"struct|enum|function|modifier \(([\[\] ()a-zA-Z0-9\.,_]*)\)", t)
             assert len(found) <= 1
             if found:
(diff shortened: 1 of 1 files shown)


0xalpharush:
Do you have an example of what this was failing on?

DarkaMaul:
The regex matched a single character in this list : 
`strucenmfiod|`

So any of the following patterns would have matched : 

- `s (bytes)`
- `i (bytes)`

The new regex forces the identifier to be complete : 
- `struct (bytes)`

0xalpharush:
@DarkaMaul That makes sense, but the existing tests are failing. Do you have a solidity test case that fails and succeeds with this fix?

DarkaMaul:
I fixed the regex, I'm a bit unsure why this non-capturing group makes sense here.

0xalpharush:
>Do you have a solidity test case that fails and succeeds with this fix?

Still not sure what necessitated the change, so I'm going to leave this as a draft

DarkaMaul:
I don't think it is easy to write an example where the actual code fails in real world scenarios as the `typeString` is generated by the compiler.

The modification only align the user intent : matching expressions like `function (uint256)` or `enum ()` instead of allowing the match of `n ()`.


I've set up a print to check if for any of the current tests in the repo, it would have different results, and so far none have been found.

0xalpharush:
Can you run the formatter?

PR #2449, opened 2024-04-29
ecrecover usage detector for hedera

arianejasuwienas:


0xalpharush:
@arianejasuwienas Did you intend to contribute this upstream? It currently highlights all uses of `ecrecover` without knowing whether a contract is intended to be deployed to another blockchain. Perhaps this is a better fit for a plugin https://github.com/crytic/slither/tree/master/plugin_example

PR #2452, opened 2024-05-03
WIP: Migrate from argparse to Typer

DarkaMaul:
# Description
_This PR replaces https://github.com/crytic/slither/pull/2220_ 

## 1. Replace `argparse` with `typer`

Modify Slither CLI from `argparse` to [Typer](https://typer.tiangolo.com/).

With this migration we gain : 

- Argument validation using Python type hints
- A nicely formatted help message
- Auto completion support ( type `slither c<TAB>`)

I kept the `crytic-compile` module as if, and created a backward compatibility layer with it.

Notable points : 
- Arguments are either associated to the main command (`slither`) or with one of its subcommand (`detect`). For instance, the formatting or the output file.
- A backward compatibility layer to keep the `slither .` has been added, converting it to the new format : `slither detect .`.
- A new `SlitherState` object is accessible. It contains every parsed argument on the upper command to be accessed by the sub commands if needed. For instance, this is where are stored the `crytic-compile` arguments.

## 2. Update tools to use sub-commands

Each `slither` tool now has its own subcommand.
Use like `slither mutate <>`

# Screenshots
Capture d’écran 2024-05-03 à 17 14 46

# Remaining TODO

- [ ]  make sure flags are under the right subcommand
- [ ]  determine if we want to gate dependencies e.g. pip install slither[read-storage] to use slither read-storage
- [ ] **EXTENSIVE TESTING**
      While I tried to test the changes I made, it is touching almost every part of the code. Moreover, it contains breaking changes. So we should try to test a bit more (and on more than a single computer) before merging.
- [ ] review where each flags is used and if not, simplify the CLI


0xalpharush:
- [ ] add something like `--long-help` from https://github.com/crytic/slither/pull/2220 to show the crytic-compile flags that can be used 
- [ ] `--no-fail` only applies to the echidna printer and does not need to be apart of the top level command line options
- [ ] `--wiki-detectors` should remain suppressed as it is not user facing
- [ ] It looks like slither mutate is missing from the list? We can probably not show `demo` and delete `check-kpsec`

0xalpharush:
It'd be great to put together better testing infra for our CLI so we can feel confident about this and also changes like https://github.com/crytic/slither/pull/2428 (and quickly iterate on them in the future)

dguido:
## Code Review for PR #2452 (Typer Migration)

### Bugs Found

1. **Interface detection regression** - `slither/tools/mutator/__main__.py:286`
   ```python
   if target_contract.name.startswith("I"):
   ```
   CLAUDE.md explicitly states: "Use `contract.is_interface` not `contract.name.startswith("I")`". This heuristic incorrectly identifies contracts like `ImmutableToken` as interfaces.

2. **kspec_coverage passes wrong type** - `slither/tools/kspec_coverage/__main__.py:40-41`
   ```python
   kspec_coverage(
       target,  # Target object, not string
   ```
   The `kspec_coverage()` function expects a string path, but receives a `Target` object. Should be `target.target`.

3. **choose_detectors mutates global DETECTORS list** - `slither/__main__.py:246,251`
   ```python
   detectors_to_run = all_detector_classes  # Reference to DETECTORS
   ...
   detectors_to_run.remove(local_detectors[detector])  # Mutates global!
   ```
   When `--exclude` is used, this removes detectors from the global `DETECTORS` list, breaking subsequent runs in the same process. Should use `list(all_detector_classes)` to create a copy.

4. **fastText import uses wrong case** - `slither/tools/similarity/model.py:3-4`
   ```python
   from fastText import load_model
   from fastText import train_unsupervised
   ```
   The package name is `fasttext` (lowercase), not `fastText`. This will cause `ModuleNotFoundError` on case-sensitive systems.

5. **similarity tool input key mismatch** - `slither/tools/similarity/__main__.py:67` vs `plot.py:38`, `test.py:19`, `train.py:17`
   ```python
   # __main__.py stores as:
   "input_": input_,
   
   # But consumers expect:
   kwargs.get("input")  # Without underscore
   ```
   The key stored is `input_` but all consumers look for `input`. The similarity tool will always fail with "requires input parameter" errors.

6. **CI scripts use old command format** - `scripts/ci_test_interface.sh`, `scripts/ci_test_upgradability.sh:186,199`
   Scripts still use standalone commands like `slither-interface` and `slither-check-upgradeability` instead of the new subcommand format (`slither interface`, `slither check-upgradeability`). Tests 14 and 15 in upgradability appear to still use the old format.

### CLAUDE.md Violations

7. **Function length violation** - `slither/__main__.py:413-784`
   The `detect()` function is ~371 lines. CLAUDE.md requires "≤80 lines/function". Consider extracting the parameter definitions, detector execution logic, and output handling into separate functions.

8. **Commented-out code** - `slither/tools/mutator/__main__.py:174`
   ```python
   # test_command: str = args.test_cmd
   ```
   CLAUDE.md states: "No commented-out code (delete it)".

---

Issue #2477, opened 2024-06-05
optimize and make sure unused-imports is accurate


0xalpharush:
It was added in https://github.com/crytic/slither/pull/2392 but it is really slow and has some edge cases e.g. 
this is wrong
```
The following unused import(s) in src/universal/CrossDomainMessenger.sol should be removed:
        -import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol"; (src/universal/CrossDomainMessenger.sol#4)
```
`slither contracts.zip --detect unused-import `
[contracts.zip](https://github.com/user-attachments/files/15597731/contracts.zip)

0xalpharush:
@DarkaMaul Maybe you will have a chance to give this a go

smartcontracts:
Yeah this would be great, we're looking to check for unused imports to keep things clean https://github.com/ethereum-optimism/optimism/issues/11806

PR #2479, opened 2024-06-10
Unused bug fix

DarkaMaul:
# The bug

It is possible to import twice the same symbol in Solidity. This was uncovered with this pattern: 

With the following layout
```
original/X.sol
upgradable/X.sol
Counter.sol
Counter2.sol
```

`original/X,sol` and `upgradable/X.sol`:
```contract Initializable { }```

`Counter.sol`:
```
import { Initializable } from "./original/Initializable.sol";
contract Counter is Initializable {}
```

`Counter2.sol`
```
import {Counter} from "./Counter.sol";
import {Initializable} from "./upgradable/Initializable.sol";
contract SuperCounter is Counter, Initializable {}
```

In the `Counter2.sol` scope, we have both `Initializable` from `original/X` and `upgradable/X`.

# The Fix

I went for a simple fix, that's maybe not perfect : 
We are keeping the order in which symbols are seen within the `exported_symbols` list in the `FileScope` class. Because Python list keep the insertion order in place, we know which symbols came first.

When resolving contract names, we are doing this in the same order and only adding it to the list of seen contracts if no other exist with the same name : 

```python
for refId in scope.exported_symbols:
    if refId in sol_parser.contracts_by_id:
        # Only add elements if they are not present. Since we kept the exported symbols in
        # we resolve from the most local imports first.
        if contract.name not in scope.contracts:
            scope.contracts[contract.name] = contract
```


dguido:
## PR Review: #2479 - Unused-imports bug fix

### Summary

This PR addresses issue #2477, fixing a false positive in the `unused-import` detector when two contracts share the same name but exist at different paths (e.g., `original/Initializable.sol` vs `upgradeable/Initializable.sol`).

### Key Changes

1. **Symbol order preservation**: Changed `exported_symbols` from `Set[int]` to `List[int]` in `FileScope` class
2. **First-wins resolution**: Only adds contracts to scope if not already present, ensuring local imports take precedence
3. **Minor cleanup**: Detector now iterates `.values()` instead of `.items()`
4. **Test cases**: Added comprehensive test coverage with CrossDomainMessenger pattern

### Critical Issue (Blocking CI)

The PR adds 39 test cases referencing `all_detectors.UnusedImport`, but the detector is **commented out** in `slither/detectors/all_detectors.py`:

```python
# from .statements.unused_import import UnusedImport
```

This causes CI to fail with:
```
AttributeError: module 'slither.detectors.all_detectors' has no attribute 'UnusedImport'
```

**Fix required**: Uncomment the import in `all_detectors.py` or remove the test cases if the detector is intentionally disabled.

### Performance Consideration

The change from `Set` to `List` for `exported_symbols` introduces O(n) membership tests in several places:

- `scope.py:84`: `if not set(new_scope.exported_symbols).issubset(self.exported_symbols)`
- `scope.py:91`: `if symbol not in self.exported_symbols`
- `slither_compilation_unit_solc.py:260`: `if symbol not in scope.exported_symbols`

For projects with many exported symbols, this could impact performance. Consider maintaining both a list (for order) and a set (for O(1) lookups):

```python
self.exported_symbols: List[int] = []
self._exported_symbols_set: Set[int] = set()
```

Additionally, line 84 creates a new set on each iteration, which is inefficient. A generator expression would be more efficient:
```python
if any(symbol not in self.exported_symbols for symbol in new_scope.exported_symbols):
```

### Test Coverage

The snapshot shows the detector correctly identifies the unused `console.sol` import while preserving the correct `Initializable` import:
```
The following unused import(s) in CrossDomainMessenger.sol should be removed:
    -import "./utils/console.sol";
```

### Verdict

The fix correctly addresses the symbol shadowing issue by preserving insertion order. However, **CI cannot pass until the `UnusedImport` detector is enabled** in `all_detectors.py`.

Issue #175, opened 2019-02-19
Data dependency: add "must depend on"


montyly:
Right now, the data dependency is an over-approximation that merges all the dependencies coming from all the node's fathers.

As a result, in a code like:

```solidity
function f(address destination) internal{
     // use of destination
}
function g() internal{
     f(msg.sender):
     f(owner); 
}
```
`destination` in `f` is both dependent of `msg.sender` and `owner`. We should have a *must depend on* mode, to detect if a variable is **always** dependent of another one.

We can implement it by adding a new mode in [analyses/data_dependency/data_dependency.py](https://github.com/trailofbits/slither/blob/master/slither/analyses/data_dependency/data_dependency.py), where only the dependencies present in all the node's fathers are kept.

montyly:
TODO @montyly: write examples/testcases

priyankabose:
@montyly I am planning to start implementing this in Slither. Just need a few clarifications. When you say that the "must depend on" mode should should only consider the dependencies present in all the node's fathers, do you mean by immediate fathers?

0xalpharush:
Could this just be a constant value/ constant propagation analysis. If the variable was always msg.sender, it is a constant, and we can check if the value that must depended on is equivalent to the constant value. Otherwise, it is not a constant/ undefined, and we can return false.

priyankabose:
I have a clear idea now where to start after talking to @0xalpharush. I will start with the constants first.

PR #2502, opened 2024-07-03
Must depend on Analysis

priyankabose:
This implements #175 . 

Linter and reformatter ran locally.

Brief analysis description:

1.  Given a target variable, infer its dependencies within the function
2. Then for each such dependencies, take the intersection of the original variables that resulted in those dependencies
3. Stop when we find public/external function parameters, constants, and solidity variable as origin variables

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality to determine and compute must dependencies for variables in smart contracts.

- **Tests**
  - Added tests to validate the new must dependency features, ensuring correct behavior and reliability.

### DIFF
--- a/slither/analyses/data_dependency/data_dependency.py
+++ b/slither/analyses/data_dependency/data_dependency.py
@@ -332,7 +332,7 @@ def compute_must_dependencies(v: SUPPORTED_TYPES) -> Set[Variable]:
         lvalue = lvalue_details[0]
         ir = lvalue_details[2]
 
-        if not lvalue in function_dependencies["context"]:
+        if lvalue not in function_dependencies["context"]:
             function_dependencies["context"][lvalue] = set()
         read: Union[List[Union[LVALUE, SolidityVariableComposed]], List[SlithIRVariable]]
 
@@ -348,11 +348,7 @@ def compute_must_dependencies(v: SUPPORTED_TYPES) -> Set[Variable]:
     function_dependencies["context"] = convert_to_non_ssa(function_dependencies["context"])
 
     must_dependencies = set()
-    data_dependencies = (
-        list(function_dependencies["context"][v])
-        if function_dependencies["context"] is not None
-        else []
-    )
+    data_dependencies = list(function_dependencies["context"].get(v, set()))
     for i, data_dependency in enumerate(data_dependencies):
         result = compute_must_dependencies(data_dependency)
         if i > 0:
--- a/tests/unit/core/test_data/must_depend_on.sol
+++ b/tests/unit/core/test_data/must_depend_on.sol
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.19;
 
 interface IERC20 {
@@ -21,7 +22,7 @@ contract Unsafe {
     // This is not detected
     function bad2(address from, address to, uint256 am) public {
         address from_msgsender = msg.sender;
-        int_transferFrom(from_msgsender, to, amount); // from is not a constant
+        int_transferFrom(from_msgsender, to, am); // from is not a constant
     }
 
     function int_transferFrom(address from, address to, uint256 amount) internal {
--- a/tests/unit/core/test_must_depend_on.py
+++ b/tests/unit/core/test_must_depend_on.py
@@ -1,12 +1,11 @@
 from pathlib import Path
+from typing import Union
+
 from slither import Slither
 from slither.analyses.data_dependency.data_dependency import get_must_depends_on
-from slither.core.variables.variable import Variable
 from slither.core.declarations import SolidityVariable, SolidityVariableComposed
-from typing import Union
-from slither.slithir.variables import (
-    Constant,
-)
+from slither.core.variables.variable import Variable
+from slither.slithir.variables import Constant
 
 TEST_DATA_DIR = Path(__file__).resolve().parent / "test_data"
 SUPPORTED_TYPES = Union[Variable, SolidityVariable, Constant]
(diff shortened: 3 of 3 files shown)


slither/analyses/data_dependency/data_dependency.py
```diff
@@ -293,6 +293,72 @@ def get_all_dependencies_ssa(
...
+
+    must_dependencies = set()
+    data_dependencies = (
+        list(function_dependencies["context"][v])
+        if function_dependencies["context"] is not None
+        else []
+    )
+    for i, data_dependency in enumerate(data_dependencies):
+        result = compute_must_dependencies(data_dependency)
+        if i > 0:
```
0xalpharush, line 356: I don't understand this condition as it seems it will always take the intersection. Could you clarify why this is done?

slither/analyses/data_dependency/data_dependency.py
```diff
@@ -293,6 +293,72 @@ def get_all_dependencies_ssa(
...
+            # if not isinstance(variable, Constant):
+            function_dependencies["context"][lvalue].add(variable)
+    function_dependencies["context"] = convert_to_non_ssa(function_dependencies["context"])
+
+    must_dependencies = set()
+    data_dependencies = (
+        list(function_dependencies["context"][v])
+        if function_dependencies["context"] is not None
+        else []
+    )
```
0xalpharush, line 353:
```suggestion
    data_dependencies = list(function_dependencies["context"].get(v, set()))
```
This could probably be a defaultdict with an empty set as the default to avoid this

slither/analyses/data_dependency/data_dependency.py
```diff
@@ -293,6 +293,72 @@ def get_all_dependencies_ssa(
...
 
+def get_must_depends_on(variable: SUPPORTED_TYPES) -> SUPPORTED_TYPES | None:
+    """
+    Return must dependency of a variable if exist otherwise return None.
+
+    :param variable: target variable whose must dependency needs to be computed
+    :return: Variable | None
+    """
+    must_dependencies = compute_must_dependencies(variable)
+    if len(must_dependencies) > 1 or len(must_dependencies) == 0:
```
0xalpharush, line 304:
```suggestion
    if len(must_dependencies) != 1:
```

slither/analyses/data_dependency/data_dependency.py
```diff
@@ -293,6 +293,72 @@ def get_all_dependencies_ssa(
     return context.context[KEY_SSA]
 
 
+def get_must_depends_on(variable: SUPPORTED_TYPES) -> SUPPORTED_TYPES | None:
```
0xalpharush, line 296:
```suggestion
def get_must_depends_on(variable: SUPPORTED_TYPES) -> SUPPORTED_TYPES | None:
```
The return type doesn't agree with what is actually returned (a list). Since we always return 1 item, maybe it should be `Union[SUPPORTED_TYPES, None]` (using `Union` since `|` limits which python versions can be used

slither/analyses/data_dependency/data_dependency.py
priyankabose: `i` here is the index for the element `data_dependency`, for the first `data_dependency` element, my idea was to put its must dependency into the result , next onwards I try to see whether must dependency is the same or not. that is why for `i>0` i do intersection, but for `i=0` i do union

dguido:
## Review of PR #2502: Must Depend On Analysis

Thanks for implementing this feature from issue #175. The concept of "must depend on" analysis (intersection of dependencies across all paths) is valuable for detecting patterns like whether a variable is *always* derived from `msg.sender`.

### Critical Issues

**1. Test Solidity file has a compilation error (causing CI failure)**

In `tests/unit/core/test_data/must_depend_on.sol:24`:
```solidity
int_transferFrom(from_msgsender, to, amount); // from is not a constant
```
`amount` is undefined - the parameter is named `am`. This should be:
```solidity
int_transferFrom(from_msgsender, to, am); // from is not a constant
```

**2. Test logic bug - string comparison with objects**

In `tests/unit/core/test_must_depend_on.py:22`:
```python
if contract == "Unsafe" and function == "int_transferFrom":
```
This compares `Contract` and `Function` objects to strings, which will never be true. Should be:
```python
if contract.name == "Unsafe" and function.name == "int_transferFrom":
```

### Code Quality Issues

**3. Return type annotation mismatch**

The function signature and docstring are inconsistent:
```python
def get_must_depends_on(variable: SUPPORTED_TYPES) -> List:
    """
    Return must dependency of a variable if exist otherwise return None.
    ...
    :return: Variable | None
    """
```

Per @0xalpharush's earlier suggestion, consider:
```python
def get_must_depends_on(variable: SUPPORTED_TYPES) -> Optional[SUPPORTED_TYPES]:
```
and return the single item directly rather than a list.

**4. Membership test style (ruff E713)**

Line 335:
```python
if not lvalue in function_dependencies["context"]:
```
Should be:
```python
if lvalue not in function_dependencies["context"]:
```

**5. Potential KeyError**

Lines 351-355:
```python
data_dependencies = (
    list(function_dependencies["context"][v])
    if function_dependencies["context"] is not None
    else []
)
```
This checks if the dict is not None, but should check if `v` is in the dict. Use `.get()`:
```python
data_dependencies = list(function_dependencies["context"].get(v, set()))
```

**6. Import order in test file**

Standard library imports (`from typing import Union`) should come before third-party imports.

**7. Missing SPDX license and final newline**

The Solidity test file is missing an SPDX license identifier and a final newline.

### Minor Suggestions

- The `function_dependencies` dict with a single `"context"` key seems like unnecessary nesting - could just be `dependencies = {}` directly
- Consider using `defaultdict(set)` instead of manual dict initialization
- The algorithm's docstring could benefit from a brief explanation of the intersection logic

### Summary

The core algorithm looks sound for implementing the "must depend on" semantics described in #175. However, the PR has blocking issues:
1. CI is failing due to the typo in the Solidity test file
2. The test logic has a bug that would prevent it from properly testing the functionality

Once these are fixed, the remaining issues are code style improvements that could be addressed incrementally.

PR #2513, opened 2024-07-25
fix(command line): markdown_root url

sambacha:
- Fixes the path to correctly have `tree` in front of the `{branch}` for generating a correct URL.

- Fixes replacing `tree/` with `blob/HEAD` so that the generated URL does not have to define the default branch name in the case that `master` is not the default branch name.

### DIFF
--- a/slither/utils/command_line.py
+++ b/slither/utils/command_line.py
@@ -396,14 +396,15 @@ def check_and_sanitize_markdown_root(markdown_root: str) -> str:
 
         if not match.group(4):
             logger.warning(
-                "Appending 'master/tree/' in markdown_root url for better code referencing"
+                "Appending 'tree/master/' in markdown_root url for better code referencing"
             )
-            markdown_root = markdown_root + "master/tree/"
+            markdown_root = markdown_root + "tree/master/"
+        # Use blob/HEAD so that we get the default branch regardless of its name
         elif match.group(4) == "tree":
             logger.warning(
-                "Replacing 'tree' with 'blob' in markdown_root url for better code referencing"
+                "Replacing 'tree' with 'blob/HEAD' in markdown_root url for better code referencing"
             )
             positions = match.span(4)
-            markdown_root = f"{markdown_root[:positions[0]]}blob{markdown_root[positions[1]:]}"
+            markdown_root = f"{markdown_root[:positions[0]]}blob/HEAD{markdown_root[positions[1]:]}"
 
     return markdown_root


0xalpharush:
Can you give an example for when this isn't working? It works like `slither . --checklist --markdown-root https://github.com/ORG/REPO/blob/COMMIT/ ` normally

sambacha:
> Can you give an example for when this isn't working? It works like `slither . --checklist --markdown-root https://github.com/ORG/REPO/blob/COMMIT/ ` normally

When just using the remote origin URL, i.e. `https://github.com/ORG/REPO/` 

Screenshot 2024-07-25 at 11 18 27 AM

I understand for GitHub Workflow CI this works:
```yaml
        slither-args: --checklist --markdown-root ${{ github.server_url }}/${{ github.repository }}/blob/${{ github.sha }}/
```

dguido:
## PR Review

Thanks for working on this fix! The current `master/tree/` path order is definitely broken.

### What This PR Does Well
1. **Fixes the path order** - `tree/master/` is valid (current `master/tree/` produces broken URLs)
2. **Uses `blob/HEAD`** for dynamic default branch support - this is a good idea for repos using `main` instead of `master`

### Suggestions for Improvement

**1. Inconsistent branch handling**

Change 1 hardcodes `master`:
```python
markdown_root = markdown_root + "tree/master/"
```

But Change 2 uses dynamic `HEAD`:
```python
markdown_root = f"{markdown_root[: positions[0]]}blob/HEAD{markdown_root[positions[1] :]}"
```

Consider using `blob/HEAD/` in both cases for consistency:
```python
markdown_root = markdown_root + "blob/HEAD/"
```

**2. `tree` vs `blob`**

Since this function generates links to source files (with line numbers), `blob/` is more appropriate than `tree/`:
- `tree/` shows directory views
- `blob/` shows file views with syntax highlighting and line anchors

GitHub redirects `tree/` to `blob/` for files, so it works, but `blob/` is semantically correct.

**3. Test coverage**

There are no unit tests for `check_and_sanitize_markdown_root()`. Adding tests would help prevent regressions:
```python
def test_sanitize_markdown_root_bare_url():
    result = check_and_sanitize_markdown_root("https://github.com/ORG/REPO/")
    assert result == "https://github.com/ORG/REPO/blob/HEAD/"

def test_sanitize_markdown_root_tree_url():
    result = check_and_sanitize_markdown_root("https://github.com/ORG/REPO/tree/branch")
    assert result == "https://github.com/ORG/REPO/blob/HEAD/branch"
```

### Summary

The PR improves on the current broken behavior. With the suggestions above (consistent `blob/HEAD/` usage), it would fully address the use case of supporting repos with any default branch name.

PR #2518, opened 2024-07-28
chore: fix some comments

withbest:


montyly:
Good catch, thanks @withbest

PR #2534, opened 2024-08-15
tool: support reinitializer in slither-check-upgradeability

QiuhaoLi:
Currently, the `slither-check-upgradeability` tool doesn't support the [reinitializer](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/Initializable.sol#L152) in OZ's Initializable contract. For example, the `CounterV2` below is a new implementation of `Counter` which adds a new storage variable and changed the `changeX` function:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.24;
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
contract Counter is Initializable {
    uint256 public x;

    function initialize(uint256 _x) public initializer {
        x = _x;
    }

    function changeX() public {
        x++;
    }
}

contract CounterV2 is Initializable {
    uint256 public x;
    uint256 public y;

    function initialize(uint256 _y) public reinitializer(2) {
        y = _y;
    }

    function changeX() public {
        x = x + y;
    }
}
```

Since `slither-check-upgradeability` can't recognize the `reinitializer` modifier, it will report a high-seveirity finding that the CounterV2.initialize(uint256) does not call the initializer modifier:

```bash
qiuhao@laptop:~/tmp/foundry_test/upgrade$ slither-check-upgradeability src/Counter.sol Counter --new-contract-name Counte
rV2
...
CounterV2.initialize(uint256) (src/Counter.sol#20-22) does not call the initializer modifier.
Reference: https://github.com/crytic/slither/wiki/Upgradeability-Checks#initializer-is-not-called
```

This PR supports the `reinitializer` modifier and fixes some tests for `tests/tool/check_upgradeability`.

### DIFF
--- a/scripts/ci_test_upgradability.sh
+++ b/scripts/ci_test_upgradability.sh
@@ -2,7 +2,8 @@
 
 ### Test slither-check-upgradeability
 
-DIR_TESTS="tests/check-upgradeability"
+DIR_TESTS="tests/tools/check_upgradeability"
+solc-select install "0.5.0"
 solc-select use "0.5.0"
 
 slither-check-upgradeability  "$DIR_TESTS/contractV1.sol" ContractV1 --proxy-filename "$DIR_TESTS/proxy.sol" --proxy-name Proxy  > test_1.txt 2>&1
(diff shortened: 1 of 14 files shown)


0xalpharush:
This was merged as part of https://github.com/crytic/slither/pull/2536

PR #2547, opened 2024-09-03
Add node to CallType

smonicas:


0xalpharush:
Should we make this a new API or consider it a breaking change?

smonicas:
I think it's fine to consider it a breaking change. This API is not used in a lot of places which i should have already changed them in this PR, i don't expect third party scripts to use this API and in any case the change to be made is really simple. For simplicity i kept the same *CallType also in the Node which with the addition of the node it's expected to be a redundant information. I also did not check for places where this new API could be used to optimize the process instead of iterating all the nodes as it's probably done right now, it can be for another PR if we decide to merge this one.

PR #2560, opened 2024-09-18
Add assert information for echidna

smonicas:
Echidna now supports this format. If #2555 will be merged i think it's possible to use `all_solidity_calls` again for improved performace because the node is present `ir.node`.


montyly:
Todo
- Fix conflict
- Test with echidna

PR #2565, opened 2024-09-25
feat: add detector for unused custom errors

gvladika:
New detector looks for Solidity custom errors which are declared, but never used. This can be indication of an error, ie. error is declared but check which would invoke the error was never put in place. 

It can be executed like this:
```
slither .  --detect unused-custom-errors
```

Output will list identified unused errors and the respective file names. Ie.
```
The following unused error(s) should be removed:
        -UnusedParentErr() (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ContractToTestForCustomErrors.sol)

        -UnusedParentErrWithArg(uint256) (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ContractToTestForCustomErrors.sol)

        -UnusedParentErr() (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ContractToTestForCustomErrors.sol)

        -UnusedParentErrWithArg(uint256) (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ContractToTestForCustomErrors.sol)

        -UnusedLibError() (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/CustomErrorsLib.sol)

        -UnusedErrorLibConsumerErr() (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ErrorLibConsumer.sol)

        -UnusedTopLevelErr() (tests/e2e/detectors/test_data/unused-custom-errors/0.8.15/ContractToTestForCustomErrors.sol)
```


slither/detectors/statements/unused_custom_errors.py
```diff
@@ -0,0 +1,94 @@
...
+        for contract in self.compilation_unit.contracts:
+            for function in contract.functions_and_modifiers:
+                for internal_call in function.internal_calls:
+                    if isinstance(internal_call.function, SolidityCustomRevert):
+                        custom_reverts.add(internal_call.function)
+
+        # Find unused custom errors
+        for declared_error in declared_custom_errors:
+            if not any(
+                declared_error.name in custom_revert.name for custom_revert in custom_reverts
```
smonicas, line 74: Checking the name could have false negative. For example try adding this in the contract test. It won't get detected.
```solidity
contract C {
    error UsedParentErr(address x);
}
```
A better approach could be to have `custom_errors` with the actual CustomError instance instead of the name and then check here with `declared_error == custom_revert`. When you are adding the custom error you can get the instance with `internal_call.function.custom_error`

slither/detectors/statements/unused_custom_errors.py
```diff
@@ -0,0 +1,94 @@
...
+        results = []
+        if len(unused_custom_errors) > 0:
+            info: DETECTOR_INFO = ["The following unused error(s) should be removed:"]
+            for custom_error in unused_custom_errors:
+                file_scope = (
+                    custom_error.file_scope
+                    if isinstance(custom_error, CustomErrorTopLevel)
+                    else custom_error.contract.file_scope
+                )
+                info += ["\n\t-", custom_error.full_name, " (", file_scope.filename.short, ")\n"]
```
smonicas, line 91: You can just use `custom_error` and in the output you will have automatically its filename. Also it would be better to have individual findings for each custom error instead of one with everything.

smonicas (changes requested):

dguido:
## Code review

Found 3 issues (1 already identified by @smonicas, 2 additional):

1. **Substring matching causes both false negatives and false positives** (already noted by @smonicas). The check `declared_error.name in custom_revert.name` has two problems:
   - False negative: If `error Foo()` and `error Foo(uint256)` exist, using only `Foo(uint256)` incorrectly marks `Foo()` as used
   - False positive: If `error Error()` exists and `MyError()` is used, `Error()` is incorrectly marked as used since `"Error" in "revert MyError(...)"` is True

   https://github.com/crytic/slither/blob/e5475d06c4f0b8b5b01ea978e0063230d5afe6e2/slither/detectors/statements/unused_custom_errors.py#L72-L76

   Fix: Use `declared_error == custom_revert.custom_error` for object comparison

2. **Missing top-level functions check**. The detector only checks `contract.functions_and_modifiers` but not `compilation_unit.functions_top_level`. Custom errors used only in top-level functions would be incorrectly reported as unused.

   https://github.com/crytic/slither/blob/e5475d06c4f0b8b5b01ea978e0063230d5afe6e2/slither/detectors/statements/unused_custom_errors.py#L65-L69

3. **Missing `require(bool, error)` support** (Solidity 0.8.26+). The detector only checks for `SolidityCustomRevert` in revert statements, but Solidity 0.8.26+ allows `require(condition, MyError())`. Errors used this way would be incorrectly reported as unused.

Issue #2566, opened 2024-09-25
[Bug-Candidate]: Fails to generate IR for solidity file that reads an Event Selector


hamdiallam:
### Describe the issue:

slither checks fails due to an error in this PR: ethereum-optimism/optimism#11592

With some debugging on when it fails, the faulty line is the `SentMessage` member here
```solidity
if (selector != SentMessage.selector) revert EventPayloadNotSentMessage();
```

Looking at when the assertion fails, it seems like adding `Event` as a a valid instance of `variable_left` for Member fixes this bug

### Code example to reproduce the issue:

if (selector != SentMessage.selector) revert EventPayloadNotSentMessage();

### Version:

0.10.4

### Relevant log output:

```shell
Failed to generate IR for L2ToL2CrossDomainMessenger.relayMessage. Please open an issue https://github.com/crytic/slither/issues.
L2ToL2CrossDomainMessenger.relayMessage (src/L2/L2ToL2CrossDomainMessenger.sol#132-182):
 	_id.origin != Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER
	revert IdentifierOriginNotL2ToL2CrossDomainMessenger()()
	CrossL2Inbox(Predeploys.CROSS_L2_INBOX).validateMessage(_id,keccak256(bytes)(_sentMessage))
	(selector,_destination,_target,_nonce) = abi.decode(_sentMessage,(bytes32,uint256,address,uint256))
	selector != SentMessage.selector
	revert EventPayloadNotSentMessage()()
	_destination != block.chainid
	revert MessageDestinationNotRelayChain()()
	_target == Predeploys.CROSS_L2_INBOX
	revert MessageTargetCrossL2Inbox()()
	_target == Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER
	revert MessageTargetL2ToL2CrossDomainMessenger()()
	(_sender,_message) = abi.decode(_sentMessage,(address,bytes))
	_source = _id.chainId
	messageHash = keccak256(bytes)(abi.encode(_destination,_source,_nonce,_sender,_target,_message))
	successfulMessages[messageHash]
	revert MessageAlreadyRelayed()()
	_storeMessageMetadata(_source,_sender)
	success = SafeCall.call(_target,msg.value,_message)
	success
	successfulMessages[messageHash] = true
	RelayedMessage(_source,_nonce,messageHash)
	FailedRelayedMessage(_source,_nonce,messageHash)
[...]

PR #2567, opened 2024-09-25
bugfix: IR generation when parsing Event as left variable

hamdiallam:
Fix for #2566 which adds Event as a valid left variable


montyly:
thanks for the fix @hamdiallam

PR #2568, opened 2024-09-26
fix: mapping to type value lookup with top-level constant

0xalpharush:


montyly:
@0xalpharush do you have a test case for this one by chance?

0xalpharush:
There is a codebase causing slither to crash that @priyankabose shared on slack which I tested this against, but I do not have a minimal reproduction

PR #2571, opened 2024-10-02
CI Improvement

montyly:
Overwrite https://github.com/crytic/slither/pull/2555 with CI improvements tentative


mandervrik (commented):
solidity

PR #2572, opened 2024-10-02
Reduce the number of CI tests on PR

montyly:
Everything is still run on a daily basis with cron, but this PR remove unecessary/problematic CI checks on every PR

### DIFF
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -5,14 +5,12 @@ defaults:
   run:
     shell: bash
 
-# We run all the CI tests only on cron job
-# This is to reduce the CI time, and prevent issues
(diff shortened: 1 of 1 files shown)


elopez:
Can this be done without duplicating the workflow? See e.g. how we do to differenciate the python versions between PRs and other runs

https://github.com/crytic/slither/blob/dfab5ae696a26c2efce9de6cc569d49663af8001/.github/workflows/ci.yml#L28

montyly:
@elopez : maybe? 

I do think that in the long term, we will likely have two different sets of tests for the ones we run on every PR and the ones we run through the cron. 

Right now, there is an intersection, but most likely, we don't need that once we have a better testing structure.

PR #2585, opened 2024-10-10
Add StateVariable location

smonicas:
Add the location to a `StateVariable` so we know if it's in persistent or transient storage.


montyly:
Maybe we should use a boolean or an enum? Having the location as a string is likely going to create misuse of the API

smonicas:
I used a string to keep the same behavior as for `LocalVariable`

PR #2589, opened 2024-10-11
Fix IR conversion when an Event selector is accessed

smonicas:


montyly:
I ended up adding the support for `this.func.selector + added test cases for different `selector` usage + fixed some python types

PR #2648, opened 2025-01-27
Mutation repairs

bohendo:
- try/catch mutant generation, so one bugged mutator won't crash the entire campaign (and it'll print more helpful logs now)
- more tweaks to logging (eg uncaught mutants use white text so the above errors stand out better)
- bug fixes to a few mutators that eg crashed on assembly instructions


bohendo:
closed in favor of PR#2662

PR #2653, opened 2025-01-31
slither-mutate: fix AOR mutator

smonicas:
Before the AOR mutator could crash when an array was accessed through multi levels such as as a value of a mapping because the `ir_expression.called.expression` would be an Index operator. Removing the specific checks on the `Identifier` being an `ArrayType` make it works for all the possible cases with no practical downsides, maybe it could be a problem if there is a library with custom functions called `pop` and `push` used with the using-for directive
```
contract Counter {
    mapping(uint => uint[]) p;

    function s() public {
       p[1].push(2);
   }
}
```

### DIFF
--- a/slither/tools/mutator/README.md
+++ b/slither/tools/mutator/README.md
@@ -33,8 +33,7 @@ options:
   --timeout TIMEOUT     Set timeout for test command (by default 30 seconds)
   --output-dir OUTPUT_DIR
                         Name of output directory (by default 'mutation_campaign')
-  -v, --verbose         log mutants that are caught as well as those that are uncaught
-  -vv, --very-verbose   log mutants that are caught, uncaught, and fail to compile. And more!
+  -v, --verbose         log mutants that are caught, uncaught, and fail to compile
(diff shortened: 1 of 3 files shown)


slither/tools/mutator/README.md
```diff
@@ -33,8 +33,7 @@ options:
   --timeout TIMEOUT     Set timeout for test command (by default 30 seconds)
   --output-dir OUTPUT_DIR
                         Name of output directory (by default 'mutation_campaign')
-  -v, --verbose         log mutants that are caught as well as those that are uncaught
-  -vv, --very-verbose   log mutants that are caught, uncaught, and fail to compile. And more!
```
montyly, line 37: is this removal expected?

smonicas: Yes. It was done [here](https://github.com/crytic/slither/pull/2625/commits/9ea73293ad7de1a5a20a9526f55e7e258f25d47c) but the the CLI arg and README were not updated

PR #2661, opened 2025-02-05
remap offsets from bytes to chars during source map conversion

bohendo:
Remaps byte offsets to char offsets in the `_convert_source_mapping` method of the `Source` class.

This method now:
- uses utf-8 encoding to convert the source code string to a bytes array
- applies the byte offsets
- decodes back to a normal string
- uses the length of the resulting string to determine new char-denominated offsets

Assumptions:
- this method is only ever used to ingest output from the solc compiler, which [infamously](https://github.com/ethereum/solidity/issues/14733) outputs byte-denominated src map offsets.
- other platforms (eg foundry, hardhat) already provide char-denominated src map offsets

Problem: `if f not in sourceUnits` then we don't have reliable access to the source filename or code so we can't use the aforementioned method to convert to byte offsets. It's not clear to me when or why this logic branch would be taken, however.

### DIFF
--- a/slither/core/source_mapping/source_mapping.py
+++ b/slither/core/source_mapping/source_mapping.py
@@ -165,16 +165,19 @@ def _convert_source_mapping(
     if len(position) != 1:
         return Source(compilation_unit)
 
-    s, l, f = position[0]
-    s = int(s)
-    l = int(l)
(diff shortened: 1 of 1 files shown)


bohendo:
This PR demonstrates a deep misunderstanding of the underlying problem

PR #2667, opened 2025-02-17
Add storage_variables printer

nisedo:
Add a printer for storage variables, which serves the same purpose as the entry-points printer: `slither . --print storage-variables`


montyly:
For visibility, we have `VariableOrder` which is a bit similar. Should we keep both, or merge them?

nisedo:
**Current screenshot of the `storage-variables` printer:**

Screenshot 2025-02-26 at 10 49 33

**Current screenshot of the `variable-order` printer:**

Screenshot 2025-02-26 at 10 51 12

smonicas:
I'm not sure about merging this one as it would be almost an exact duplicate of the `variable-order` printer. I would rather improve `variable-order`.

PR #2668, opened 2025-02-17
Dev update entry points printer

nisedo:
Update entry-points printer by improving filtering and sorting functions by inheritance


nisedo:
**Current screenshot of the `entry-points` printer:**

Screenshot 2025-02-26 at 17 09 49

PR #2748, opened 2025-07-09
Fix #2747: Deduplicate functions in get_msg_sender_checks to avoid repeated msg.sender condition nodes

### LINKED ISSUE #2747: [Bug-Candidate]: Printer vars-and-auth contains duplicated message
### Describe the issue:

Printer vars-and-auth put duplicated message under the `Conditions on msg.sender ` section.

Like this:
```
+----------+-------------------------+----------------------------------------------------------------------------------------+
| Function | State variables written | Conditions on msg.sender                                                               |
+----------+-------------------------+----------------------------------------------------------------------------------------+
| func     | []                      | ['require(bool)(msg.sender != address(0))', 'require(bool)(msg.sender != address(0))'] |
+----------+-------------------------+----------------------------------------------------------------------------------------+
```
it contains `require(bool)(msg.sender != address(0))` twice.

I think the problem is from this part, `isinstance(ir.function, Function)` will return True for both `Function` type and `Modifier` type

https://github.com/crytic/slither/blob/dac531400a2574b8acfcf11967375727a08f49ad/slither/printers/functions/authorization.py#L20-L29

I guess remove line 28 will solve the problem : )

This is the solidity code I used for test
```solidity
contract Test {
    modifier mod() {
        require(msg.sender != address(0));
        _;
    }

    function func() mod public {
    }
}
```

command
```
slither test/testM.sol --print vars-and-auth
```

This code shows that `isinstance(m, Function)` returns True for the a modifier `mod`
```python
from slither.core.declarations.function import Function

slither = Slither('test.sol')
contract = slither.contracts[0]

for m in contract.modifiers:
    print(m.name)
    print(isinstance(m, Function))   # True
```

### Version:

0.11.3

### Relevant log output:

```shell
INFO:Printers:
Contract Test
+----------+-------------------------+----------------------------------------------------------------------------------------+
| Function | State variables written | Conditions on msg.sender                                                               |
+----------+-------------------------+----------------------------------------------------------------------------------------+
| func     | []                      | ['require(bool)(msg.sender != address(0))', 'require(bool)(msg.sender != address(0))'] |
+----------+-------------------------+----------------------------------------------------------------------------------------+

INFO:Slither:test.sol analyzed (1 contracts)
```

sidarth16:
Fix: Deduplicate functions in get_msg_sender_checks to avoid repeated msg.sender condition nodes (#2747)

✅ Applied Fix (safe & immediate):
- Wraps the combined function list in a `set()` to deduplicate `Function` instances.
- Ensures complete analysis while avoiding repeated conditions in output.

⚠️ Alternative (riskier but potentially optimal):
- modifier functions are being included both via internal calls and explicitly in function.modifiers.
- Removing explicit `function.modifiers` in `all_functions` could also prevent duplicates,
  but risks missing edge cases where a modifier is attached but not present in internal calls.

The current fix ensures correctness, safety, and consistent analysis output.

### DIFF
--- a/slither/printers/functions/authorization.py
+++ b/slither/printers/functions/authorization.py
@@ -18,7 +18,7 @@ class PrinterWrittenVariablesAndAuthorization(AbstractPrinter):
 
     @staticmethod
     def get_msg_sender_checks(function: Function) -> List[str]:
-        all_functions = (
+        all_functions = set(
             [
                 ir.function
                 for ir in function.all_internal_calls()


smonicas:
Closing this in favor of #2865 as it's more correct

Issue #2209, opened 2023-11-01
[Bug-Candidate]: Misinterpreting function arguments when the called function is overloaded


Tiko7454:
### Describe the issue:

In the code below there is a call from function g to function f of the contract A. Slither gets the first argument of the call and misinterprets it. Slither expects List[Variable] but another thing, it throws an exception and refuses to convert to slithir. I've printed the argument list (`arguments`) from _find_function_from_parameter of slither/slithir/convert.py. I got `[[0, <slither.core.variables.local_variable.LocalVariable object at 0x7fe4b3e0f990>]]`. (in the log the line numbers can differ because of the prints added by me)

### Code example to reproduce the issue:

contract A {
    function f(uint256[2] calldata arr) external {}
    function f(uint256[4] calldata arr) external {}
}

contract C {
    function g(A a, uint256 num) public {
        a.f([0, num]);
    }
}

### Version:

0.10.0

### Relevant log output:

```shell
'solc --version' running
'solc a.sol --combined-json abi,ast,bin,bin-runtime,srcmap,srcmap-runtime,userdoc,devdoc,hashes,compact-format --allow-paths .,/home/tigran/Projects/slither/slither' running
Compilation warnings/errors on a.sol:
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> a.sol

Warning: Source file does not specify required compiler version! Consider adding "pragma solidity ^0.8.7;"
--> a.sol

ERROR:ContractSolcParsing:Impossible to generate IR for C.g (a.sol#7-9):
 'list' object has no attribute 'type'
Traceback (most recent call last):
  File "/home/tigran/Projects/slither/slither/.slither/bin/slither", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/tigran/Projects/slither/slither/.slither/lib/python3.11/site-packages/slither/__main__.py", line 727, in main
    main_impl(all_detector_classes=detectors, all_printer_classes=printers)
  File "/home/tigran/Projects/slither/slither/.slither/lib/python3.11/site-packages/slither/__main__.py", line 833, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
[...]

PR #2806, opened 2025-10-18
Bug Fix: Overloaded functions broke when array literals were called. #2209

quirps:
Fixes a bug in SlithIR generation where array literals containing complex elements (primitives or struct initializations like [B(x), B(y)]) caused analysis to crash when used as function arguments.

Two new unit tests (test_primitive_array_instantiation_fix and test_struct_array_instantiation_fix) have been added to prevent regression.

### DIFF
--- a/.github/linters/.markdownlint.json
+++ b/.github/linters/.markdownlint.json
@@ -1,0 +1,11 @@
+{
+  "MD013": false,
+  "MD024": false,
+  "MD025": false,
+  "MD033": false,
+  "MD040": false,
+  "MD041": false,
+  "MD045": false,
+  "MD046": false,
+  "MD056": false
+}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,5 @@
 ---
-name: Lint with Ruff
+name: Lint
 
 defaults:
   run:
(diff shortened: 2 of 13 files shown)


dguido:
## Code Review

Good fix for issue #2209. The implementation correctly addresses the crash when calling overloaded functions with array literal arguments.

### Issue Analysis

**Root cause**: When `_find_function_from_parameter` received an array literal (represented as a Python list), the code:
1. Correctly resolved the type at the start: `type_args = [f"{get_type(arg[0].type)}[{len(arg)}]"]`
2. But then unconditionally accessed `arg.type` further down, causing `AttributeError: 'list' object has no attribute 'type'`

**Fix**: Wrap the implicit type conversion logic (for uint/int/bytes) in a guard: `if not isinstance(arg, (list,)):`

### What the PR does well

1. **Minimal and focused fix** - Only adds the necessary guard without restructuring unrelated code
2. **Good comments** - Clear explanation of why the guard is needed ("Array Instantiation Guard")
3. **Readability improvement** - Changing `not_found` to `found` and inverting the logic is cleaner
4. **Comprehensive tests** - Covers both primitive arrays (`[0, num]`) and struct arrays (`[B(num), B(num + 1)]`)

### Minor observation

The code accesses `arg[0].type` without checking if the list is empty. This is fine because:
- Solidity fixed-size arrays must have length >= 1
- Empty array literals can't match any valid function signature

### No blocking issues found

The fix is correct and well-tested. LGTM for merge.

PR #2828, opened 2025-12-03
feat(mutator): add --target-functions flag to filter mutations by selector

stevennevins:
Add CLI flag to slither-mutate that filters mutations to specific functions by their selector. Supports both hex selectors (0xa9059cbb) and full signatures (transfer(address,uint256)).

Features:
- Parse comma-separated selectors with parenthesis-aware splitting
- Include modifiers used by target functions
- Abort with error if no functions match selector
- Combine with --contract-names for precise targeting


stevennevins:
@bsamuels453 for viz

bohendo (approved):
Love it, good work. Implementation is clean and I appreciate you providing an associated test suite.

PR #2830, opened 2025-12-05
fix: use block comments in CR mutator to prevent compilation errors

stevennevins:
### Summary

CR mutator now uses block comments (`/* */`) instead of line comments (`//`).

### Problem

Line comment `//` comments out everything after it on the line. This breaks compilation when code continues:
```solidity
// Original
function test() public onlyOwner {

// With // - BROKEN
function test() public // onlyOwner {

// With /* */ - WORKS  
function test() public /* onlyOwner */ {
```

### Solution
```python
new_str = "/* " + old_str + " */"
```

### Scenarios Improved

**Function modifier same line as brace:**
```solidity
// Original
function foo() onlyOwner {

// Before (broken)
function foo() // onlyOwner {

// After (works)
function foo() /* onlyOwner */ {
```

**Single-line functions:**
```solidity
// Original
function getX() view returns (uint) { return x; }

// Before (broken) - commenting return statement
function getX() view returns (uint) { // return x; }

// After (works)
function getX() view returns (uint) { /* return x; */ }
```

### DIFF
--- a/.github/linters/.markdownlint.json
+++ b/.github/linters/.markdownlint.json
@@ -1,0 +1,11 @@
+{
+  "MD013": false,
+  "MD024": false,
+  "MD025": false,
+  "MD033": false,
+  "MD040": false,
+  "MD041": false,
+  "MD045": false,
+  "MD046": false,
+  "MD056": false
+}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,5 @@
 ---
-name: Lint with Ruff
+name: Lint
 
 defaults:
   run:
@@ -11,6 +11,7 @@ on:
   pull_request:
     branches: [master]
     paths:
+      - "docs/**/*.md"
       - "**/*.py"
       - "**/*.yml"
       - "**/*.yaml"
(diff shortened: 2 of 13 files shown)


dguido:
## Code review

Found 1 issue:

1. **Byte offset used as character index** - The new semicolon-checking code uses `stop` (a byte offset from Solidity's source mapping) as a character index into `source_code` (a Python string). This is incorrect for files containing multi-byte UTF-8 characters (e.g., unicode in comments or string literals). The codebase explicitly documents this limitation in `source_mapping.py:content`: "Use this property instead of eg source_code[start:end] Above will return incorrect content if source_code contains any unicode because self.start and self.end are byte offsets, not char offsets"

https://github.com/crytic/slither/blob/3fa85c33df0515a31655a7c6b95b54da7477ab0b/slither/tools/mutator/mutators/CR.py#L31-L36

The fix should encode the source code to UTF-8 bytes before indexing, similar to how `abstract_mutator.py` does it:
```python
source_bytes = self.slither.source_code[self.in_file].encode("utf8")
if stop < len(source_bytes) and source_bytes[stop:stop+1] == b";":
```

Reference: https://github.com/crytic/slither/blob/fd71cabd0770f16784d5046262c214743afa9e62/slither/core/source_mapping/source_mapping.py#L71-L89

PR #2832, opened 2025-12-08
fix: handle Unicode characters correctly in mutator byte offset calculation

stevennevins:
Fix source code corruption when `slither-mutate` processes Solidity files containing Unicode characters (e.g., arrow symbols, non-ASCII comments).

### Example

```solidity
// → Unicode arrow
contract Counter {
    function increment() public {
        number++;
    }
}
```

The `→` arrow is 3 bytes but 1 character. This 2-byte difference causes mutations to be applied at wrong positions.

**Without fix** - garbage output:
```
nurevert()
fufunction setNumber
nu++number
```

**With fix** - correct mutations:
```
revert()
function setNumber
++number
```

### DIFF
--- a/slither/tools/mutator/utils/testing_generated_mutant.py
+++ b/slither/tools/mutator/utils/testing_generated_mutant.py
@@ -92,12 +92,15 @@ def test_patch(
     function to verify whether each patch is caught by tests
     returns: 0 (uncaught), 1 (caught), or 2 (compilation failure)
     """
-    with open(file, "r", encoding="utf8") as filepath:
+    with open(file, "rb") as filepath:
         content = filepath.read()
     # Perform the replacement based on the index values
-    replaced_content = content[: patch["start"]] + patch["new_string"] + content[patch["end"] :]
+    # Note: patch offsets are byte offsets from solc, so we must work with bytes
+    replaced_content = (
+        content[: patch["start"]] + patch["new_string"].encode("utf8") + content[patch["end"] :]
+    )
     # Write the modified content back to the file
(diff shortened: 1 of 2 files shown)


slither/tools/mutator/utils/testing_generated_mutant.py
```diff
@@ -96,12 +96,15 @@ def test_patch(  # pylint: disable=too-many-arguments
...
         content = filepath.read()
     # Perform the replacement based on the index values
-    replaced_content = content[: patch["start"]] + patch["new_string"] + content[patch["end"] :]
+    # Note: patch offsets are byte offsets from solc, so we must work with bytes
+    replaced_content = (
+        content[: patch["start"]] + patch["new_string"].encode("utf8") + content[patch["end"] :]
+    )
     # Write the modified content back to the file
-    with open(file, "w", encoding="utf8") as filepath:
+    with open(file, "wb") as filepath:
```
bohendo: Man. I fixed a related issue in a [prev PR](https://github.com/crytic/slither/pull/2662) and I thought that had us good to go wrt unicode support. But looking at this now, idk how the prev fix actually fixed anything. This is the same byte-wise indexing-against-strings bug that I was wrestling w earlier, just in a different but still critical spot. Good job finding & fixing this.

bohendo (approved):
Very nice, lgtm

smonicas:
Hi, thanks for the PR. If you can merge dev and fix the conflict i'll merge it. Don't worry about the current pip-audit failure it will go away.

stevennevins:
all set @smonicas, thanks for the reviews

PR #2840, opened 2025-12-20
Fix function-summary printer incorrectly reporting internal calls as external

### LINKED ISSUE #2073: [Bug-Candidate]: all statements with `.` are treated as external calls
### Describe the issue:

I ran `slither . --print function-summary` to view external calls for functions. The resulting "External Calls" column treats any statement with a `.` as an external call. This includes solidity built-ins like `abi.encode`, library accesses like `Address.sendValue`, and struct field accesses.

```bash
git clone https://github.com/scopelift/pooltogether-governor-upgrade/
cd pooltogether-governor-upgrade
forge install
slither . --print function-summary
```

### Version:

0.9.6

### Relevant log output:

_No response_

edlsh:
Fixes #2073

## Problem
The function-summary printer incorrectly classified all member access expressions (`x.foo()`) as external calls, including:
- Solidity built-ins: `abi.encode()`, `abi.decode()`
- Library calls: `SafeMath.add()` (direct and via `using X for Y`)

## Solution
- Parse-time filtering in `_classify_calls()` for built-ins and direct library calls
- IR-based `high_level_calls - library_calls` in `get_summary()` for `using X for Y` accuracy

Includes regression test covering all cases.

### DIFF
--- a/tests/e2e/printers/test_function_summary.py
+++ b/tests/e2e/printers/test_function_summary.py
@@ -80,9 +80,9 @@ def test_function_summary_external_calls(solc_binary_path) -> None:
 
     call_multiple = test_external.get_function_from_signature("callMultipleExternal(uint256)")
     ext_calls = _get_external_calls_from_summary(call_multiple)
-    assert (
-        len(ext_calls) == 2
-    ), "TestExternalCalls.callMultipleExternal should have 2 external calls"
+    assert len(ext_calls) == 2, (
+        "TestExternalCalls.callMultipleExternal should have 2 external calls"
+    )
 
     # Test TestMixed contract - mixedOperations should have exactly 1 external call
     test_mixed = slither.get_contract_from_name("TestMixed")[0]


dguido:
## Code Review

Good fix for issue #2073. The implementation correctly addresses the problem of misclassifying internal calls as external calls.

### What the PR does well

1. **Two-layer approach** - Correctly uses both parse-time classification (`_classify_calls`) and IR-based filtering (`high_level_calls - library_calls`) in `get_summary()`. This is necessary because:
   - Parse-time can detect `abi.encode()` and direct library calls like `SafeMath.add()`
   - IR-based is needed for `using X for Y` syntax (e.g., `a.add(b)` where `a` is `uint256`)

2. **Correct handling of `this.foo()`** - Properly treats `this.foo()` as external (uses CALL opcode) while `abi.encode()` is internal.

3. **Comprehensive tests** - Good coverage of all edge cases.

4. **Code quality** - Well-commented, follows project patterns, passes linting.

### One limitation to note

The same buggy pattern exists in Vyper and Yul parsing (as mentioned in the original issue comment by @0xalpharush):
- `slither/vyper_parsing/cfg/node.py:60`
- `slither/solc_parsing/yul/parse_yul.py:94`

This PR only fixes Solidity parsing, which is reasonable scope. These could be addressed in a follow-up PR.

### No blocking issues found

The fix is correct and well-tested. LGTM for merge.

edlsh:
Follow-up PR #2938 addresses the Vyper/Yul call-classification gap you noted by reusing the shared classifier for those parsers and adding targeted unit tests (raw_call/send + call/callcode/delegatecall/staticcall).

PR #2867, opened 2026-01-14
Add detector for `msg.value` usage unreachable from payable entry points

### LINKED ISSUE #2781: Enhancement: Detect msg.value Usage in Non-Payable Functions
## Summary
Add a detector that identifies uses of `msg.value` in functions that can never receive ETH, including non-payable functions and their entire call chains, helping developers identify logic errors and dead code.

## Motivation

Using `msg.value` in non-payable functions is a common mistake that indicates:
- **Logic errors** - The developer expected the function to receive ETH
- **Dead code** - The msg.value check will always be zero
- **Incorrect assumptions** - May lead to bypassed security checks
- **Gas waste** - Unnecessary checks that always fail

This is more sophisticated than simple visibility checking - it requires analyzing the entire call graph to ensure `msg.value` is only used in functions that can be reached from payable functions.

## Problem Examples

```solidity
contract VulnerableContract {
    // ISSUE 1: Direct msg.value in non-payable
    function processPayment() external {
        require(msg.value > 0, "Payment required"); // Always fails!
        // Process...
    }
    
    // ISSUE 2: Indirect msg.value usage
    function internalCheck() internal view returns (bool) {
        return msg.value >= 1 ether;
    }
    
    function publicEntry() external view returns (bool) {
        return internalCheck(); // msg.value always 0 here
    }
    
    // ISSUE 3: msg.value in modifier used by non-payable
    modifier requiresPayment() {
        require(msg.value > 0, "ETH required");
        _;
    }
    
    function doSomething() external requiresPayment {
        // This can never execute successfully
    }
    
    // OK: Proper payable function
    function properPayable() external payable {
        require(msg.value > 0, "Payment required"); // Valid use
    }
}
```

## Detection Algorithm

The detector should:
1. Build a complete function call graph
2. Find all msg.value accesses
3. For each access, check if it can be reached from any payable function
4. Track modifier usage and inheritance patterns
5. Report unreachable msg.value uses with call chain information

## Expected Output

```
msg.value used in non-payable context
  Location: Contract.sol:45
  Function: processPayment() [non-payable]
  Issue: msg.value will always be 0
  
  Call chain analysis:
    Entry points that can reach this function:
    - publicEntry() [external, non-payable] -> processPayment()
    - adminFunction() [external, non-payable] -> internalHelper() -> processPayment()
    
  No payable functions can reach this code.
  
  Suggestion: Either:
    1. Make the function payable if it should receive ETH
    2. Remove the msg.value check as it will always be 0
    3. Ensure this function is only called from payable contexts
```

## Test Cases

```solidity
contract TestCases {
    modifier needsPayment() {
        require(msg.value > 0);
        _;
    }
    
    // DETECT: Direct msg.value in non-payable
    function test1() external {
        if (msg.value > 0) { } // Issue
    }
    
    // DETECT: msg.value in view function
    function test2() external view returns (uint) {
        return msg.value; // Always 0
    }
    
    // DETECT: Modifier with msg.value on non-payable
    function test3() external needsPayment {
        // Can never execute
    }
    
    // OK: Payable function
    function test4() external payable {
        require(msg.value > 0); // Valid
    }
    
    // OK: Internal called from payable
    function test5Internal() internal {
        require(msg.value > 0); // OK if called from payable
    }
    
    function test5() external payable {
        test5Internal(); // Valid call chain
    }
    
    // DETECT: Complex call chain, no payable entry
    function helper() internal view returns (bool) {
        return msg.value > 0;
    }
    
    function middle() internal view returns (bool) {
        return helper();
    }
    
    function entry() external view returns (bool) {
        return middle(); // Detects msg.value in helper
    }
}
```

## Benefits

1. **Catch logic errors early** - Identify functions that can't work as intended
2. **Remove dead code** - Eliminate checks that always fail
3. **Improve gas efficiency** - Remove unnecessary msg.value checks
4. **Prevent security issues** - Avoid bypassed payment checks
5. **Better code understanding** - Clarify payment flow in contracts

## Priority

**High** - This is a common error pattern that indicates fundamental misunderstanding of Solidity's payment model. It can lead to functions that can never execute successfully, wasted gas, and potentially security issues if payment checks are bypassed. The detector has high precision and provides immediate value to developers.

sidarth16:
## Summary

This PR resolves #2781 and  introduces a new Slither detector, `msg-value-in-nonpayable`, which identifies uses of `msg.value` in functions that can never be reached from any payable public or external entry point.

In such cases, `msg.value` is provably meaningless:

- it is always zero, or
- execution always reverts .

This detector helps developers catch logic errors, dead code, and incorrect assumptions about ETH flow early.

## Detection Logic
The detector:
- Identifies functions that directly read `msg.value` (including via modifiers)
- Builds reachability information using Slither’s call graph
- Checks whether the function can be reached from any payable public or external entry point
- Reports the function only if no payable entry point can reach it
- Provides call-chain context for non-payable entry points that can reach the function
Only the function that uses `msg.value` is reported; callers are included for explanatory context.

## Example
```soldiity
contract Subscription {
    mapping(address => bool) public subscribed;

    function subscribe() external {
        require(msg.value >= 1 ether, "Subscription fee required");
        subscribed[msg.sender] = true;
    }
}

```
Here, `subscribe()` is not payable and cannot receive ETH.
As a result, `msg.value` is always zero and the `require` statement always fails.

## Notes
The detector follows the semantics discussed in #2781 and reports only `msg.value` usages that are provably unreachable from any payable execution path.

### DIFF
--- a/tests/e2e/detectors/snapshots/detectors__detector_MsgValueInNonPayable_0_4_25_msg_value_in_nonpayable_sol__0.txt
+++ b/tests/e2e/detectors/snapshots/detectors__detector_MsgValueInNonPayable_0_4_25_msg_value_in_nonpayable_sol__0.txt
@@ -1,0 +1,9 @@
+C.helper() (tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.4.25/msg_value_in_nonpayable.sol#40-42) uses msg.value but is not reachable from any payable function
+	Non-payable entry points that can reach this function:
+		C.entry()
+
+C.test1() (tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.4.25/msg_value_in_nonpayable.sol#11-13) uses msg.value but is not reachable from any payable function
+
+C.test2() (tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.4.25/msg_value_in_nonpayable.sol#16-18) uses msg.value but is not reachable from any payable function
+
+C.test3() (tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.4.25/msg_value_in_nonpayable.sol#21-23) uses msg.value but is not reachable from any payable function
--- a/tests/e2e/detectors/snapshots/detectors__detector_MsgValueInNonPayable_0_8_0_msg_value_in_nonpayable_sol__0.txt
+++ b/tests/e2e/detectors/snapshots/detectors__detector_MsgValueInNonPayable_0_8_0_msg_value_in_nonpayable_sol__0.txt
@@ -1,0 +1,3 @@
+C.helper() (tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.8.0/msg_value_in_nonpayable.sol#20-22) uses msg.value but is not reachable from any payable function
+	Non-payable entry points that can reach this function:
+		C.entry()


sidarth16:
Detector output when executed on the `TestCases` contract provided in issue #2781 :

`slither testcases.sol --detect msg-value-in-nonpayable`


image

Flagged functions (as expected from #2781):

- `test1()`
- `test2()`
- `test3()`
- `helper()`

sidarth16:
Hi @ep0chzer0 ,
thanks for the review.

I’ve now addressed the points you raised:

- Output format has been updated to use `DETECTOR_INFO` with Slither objects (e.g. `Function`) instead of raw strings.
- Added the `is_implemented` check to skip interface / abstract functions.
- Updated `_get_entry_point_callers` to include the function itself when it is public or external, to preserve correct entry-point semantics.
  - To avoid noisy output, the trivial self-edge is filtered only at presentation time, so it does not produce empty or redundant call-chain entries.

Here is the updated detector output now when running on the `TestCases` contract from #2781:

```
INFO:Detectors:
Detector: msg-value-in-nonpayable
TestCases.test1() (testcases.sol#11-13) uses msg.value but is not reachable from any payable function
TestCases.test2() (testcases.sol#16-18) uses msg.value but is not reachable from any payable function
TestCases.test3() (testcases.sol#21-23) uses msg.value but is not reachable from any payable function
TestCases.helper() (testcases.sol#40-42) uses msg.value but is not reachable from any payable function
        Non-payable entry points that can reach this function:
        - TestCases.entry() (testcases.sol#48-50)

Reference: https://github.com/crytic/slither/issues/2781
INFO:Slither:testcases.sol analyzed (1 contracts with 1 detectors), 4 result(s) found

```

I will also be adding the `TestCases.sol` contract referenced in #2781 to the test suite.
This contract is useful as it covers both:

- Solidity 0.4.x semantics (where `msg.value` in non-payable functions is allowed but semantically incorrect), and
- Solidity ≥0.8.x-compatible scenarios via internal functions reachable only from non-payable entry points.

sidarth16:
Hi @smonicas and @ep0chzer0 
I’ve now added E2E test cases for this detector under:
```
tests/e2e/detectors/test_data/msg-value-in-nonpayable/
├── 0.4.25/
│   └── msg_value_in_nonpayable.sol
└── 0.8.0/
    └── msg_value_in_nonpayable.sol
```

Could you please guide me on the expected next steps to integrate these tests correctly?
In particular, should the detector tests include pre-generated .zip artifacts per Solidity version, or is there a standard workflow/script to generate them?

Thanks!

ep0chzer0:
Hi @sidarth16, thanks for addressing the feedback - the changes look good!

Regarding the test setup: yes, you'll need to generate the snapshot files. The process is:

1. Add your test entry to `tests/e2e/detectors/test_detectors.py`
2. Generate the compiled artifacts (`.zip` files) using `crytic-compile`
3. Run the test once to generate the snapshot file, or create it manually with the expected output

For the zip artifacts, you can use Docker if you don't have the right solc version:
```bash
docker run --rm -v $(pwd):/src -w /src --platform linux/amd64 ghcr.io/crytic/crytic-compile:latest crytic-compile tests/e2e/detectors/test_data/msg-value-in-nonpayable/0.8.0/msg_value_in_nonpayable.sol --export-zip msg_value_in_nonpayable.sol-0.8.0.zip
```

---

**Note to maintainers (@elopez, @smonicas):** I also submitted PR #2923 for the same issue (#2781) before noticing sidarth16 had already addressed my review feedback here. Happy to close #2923 if you'd prefer to proceed with this PR - just let me know what works best for the project.

PR #2902, opened 2026-01-16
Add detector for unused custom error definitions

### LINKED ISSUE #2587: Add detection for Unused User-Defined Error types

If there exists a user-defined error type such as "Unauthorized" in the below example:

`// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.4;

contract VendingMachine {
    error Unauthorized();  // WARNING HERE ... DEFINED BUT NOT USED
    address payable owner = payable(msg.sender);

    function withdraw() public {
        owner.transfer(address(this).balance);
    }
    // ...
}`

the expectation is that the user-defined error type should either be leveraged somewhere in the code (see below example), or the user-defined error type should be removed from the code.

`// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.4;

contract VendingMachine {
    error Unauthorized();
    address payable owner = payable(msg.sender);

    function withdraw() public {
        if (msg.sender != owner)
            revert Unauthorized();   // WARNING GOES AWAY AS IT IS NOW USED

        owner.transfer(address(this).balance);
    }
    // ...
}`

The thought process is that if a user-defined error was defined but not used, then it's possible the developer missed insertion of a feature they intended to use.

This feature should work for also work for user-defined errors with parameters (the above example is of a user-defined error without parameters).  An example of a user-defined error type with and without parameters can be found in the solidity documentation here: https://soliditylang.org/blog/2021/04/21/custom-errors/ .

It's possible that a user-defined error type can be used more than once, but slither should show a warning when it is never used.

Also note that solidity allows the user-defined error to be defined inside of , or outside of the contract.  If the user-defined error is in a solidity file and defined outside of the contract, it should be used in that file, or there should be a way to silence the warning for files that are intended to be used as inclusions to other files.

ep0chzer0:
## Summary

Implements a new detector (`unused-error`) that identifies custom error definitions that are declared but never used in `revert` statements.

Fixes #2587

## Features

- Detects unused contract-level custom errors
- Detects unused top-level (file-level) custom errors  
- Handles errors with and without parameters
- Skips interface contracts (signature-only)

## Example

```solidity
contract VendingMachine {
    error Unauthorized();  // WARNING: declared but never used
    address payable owner = payable(msg.sender);

    function withdraw() public {
        // Missing: if (msg.sender != owner) revert Unauthorized();
        owner.transfer(address(this).balance);
    }
}
```

The detector will report: `Unauthorized is declared but never used in VendingMachine`

## Test

Added test case in `tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol` covering:
- Contract-level unused errors (with and without parameters)
- Top-level unused errors
- Used errors (should not be flagged)
- Interface errors (should not be flagged)

## Checklist

- [x] Detector implementation
- [x] Test file
- [x] Registered in `all_detectors.py`
- [x] Added to test suite
- [ ] Wiki documentation (to be added after merge)

### DIFF
--- a/tests/e2e/detectors/snapshots/detectors__detector_UnusedCustomErrors_0_8_4_unused_custom_error_sol__0.txt
+++ b/tests/e2e/detectors/snapshots/detectors__detector_UnusedCustomErrors_0_8_4_unused_custom_error_sol__0.txt
@@ -1,6 +1,6 @@
 TopLevelUnused() (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#5) is declared but never used
 
-UnusedErrorTest.Unauthorized() (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#12) is declared but never used in UnusedErrorTest (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#10-35)
-
 UnusedErrorTest.InvalidAmount(uint256) (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#18) is declared but never used in UnusedErrorTest (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#10-35)
+
+UnusedErrorTest.Unauthorized() (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#12) is declared but never used in UnusedErrorTest (tests/e2e/detectors/test_data/unused-error/0.8.4/unused_custom_error.sol#10-35)
 


ep0chzer0:
Hi maintainers, could someone please approve the CI workflow runs when you get a chance? This PR is ready for review. Thanks!

ep0chzer0:
Note: I noticed there's an existing PR #2565 for similar functionality. However, this PR addresses the review feedback given on #2565:

1. **Instance comparison vs name comparison**: #2565 compares errors by name, which can cause false negatives when different contracts define errors with the same name. This PR compares actual `CustomError` instances using `ir.function.custom_error`, avoiding name collisions.

2. **Individual findings**: #2565 groups all unused errors into a single finding. This PR generates individual findings for each unused error, making it easier to track and fix each one separately.

3. **Cleaner output**: Each finding includes the contract context (e.g., "Unauthorized is declared but never used in VendingMachine").

Happy to close this if maintainers prefer to continue with #2565, or this could serve as an alternative implementation addressing the review concerns.

Issue #2150, opened 2023-09-29
ignore interfaces for `inheritance-graph` printer


0xPhaze:
### Describe the desired feature

Interfaces can clutter the output of the inheritance graph printer and don't add much value. Perhaps these should be removed by default from the output.

0xalpharush:
Potentially this can be updated https://github.com/crytic/slither/tree/fix-inheritance-printer

PR #2915, opened 2026-01-16
fix: only exclude unoverridden virtual functions in inheritance-graph

ep0chzer0:
## Summary
Refines the inheritance-graph printer to only exclude virtual functions that are not overridden, instead of excluding all virtual functions.

## Problem
The current logic excludes ALL virtual functions from the inheritance graph:
```python
and not f.is_virtual
```

This is overly aggressive - virtual functions that ARE overridden by derived contracts represent actual implementations being used and should be shown.

## Solution
Changed the condition to only exclude virtual functions that have no overrides:
```python
and not (f.is_virtual and not f.overridden_by)
```

This means:
- Virtual functions with overrides → **included** (they're being used)
- Virtual functions without overrides → **excluded** (just interface, no implementation)
- Non-virtual functions → **included** (unchanged behavior)

## Example
```solidity
abstract contract Base {
    function foo() public virtual;  // No override → excluded
    function bar() public virtual;  // Has override → included
}

contract Derived is Base {
    function bar() public override { }  // Overrides Base.bar
}
```

Before: Both `foo` and `bar` excluded from Base node
After: Only `foo` excluded, `bar` is shown

Partially addresses #2150

### DIFF
--- a/slither/printers/inheritance/inheritance_graph.py
+++ b/slither/printers/inheritance/inheritance_graph.py
@@ -122,7 +122,7 @@ class PrinterInheritanceGraph(AbstractPrinter):
             for f in contract.functions
             if not f.is_constructor
             and not f.is_constructor_variables
-            and not (f.is_virtual and not f.overridden_by)  # Exclude unoverridden virtual functions
+            and not (f.is_virtual and not f.overridden_by)
             and f.contract_declarer == contract
             and f.visibility in visibilities
         ]


dguido:
## Bulk PR Review Summary

### Research Findings
- **Change**: One-line modification in `slither/printers/inheritance/inheritance_graph.py`
- **Before**: `and not f.is_virtual` (excludes ALL virtual functions)
- **After**: `and not (f.is_virtual and not f.overridden_by)` (excludes only virtual functions without overrides)
- **Addresses**: Issue #2150 comment requesting removal of "virtual functions which aren't overridden"

### Test Status
- Rebased on current master (26e522429)
- Unit tests: 198 passed (same as master)
- E2E printer tests: Skipped due to pre-existing solc-select environment issues (same failures on master)
- Ruff lint: Passes

### Code Review (Pedantic Level)
1. **Logic verification**: Correct. `overridden_by` is initialized as `[]` (line 128 in function.py), so `not f.overridden_by` is `True` when no overrides exist.
2. **Comment quality**: Good inline comment explains the filtering behavior
3. **CLAUDE.md compliance**: Single-line change, no additional requirements triggered
4. **Edge cases**: None identified - the `overridden_by` property is always a list, never None

### Merge Readiness
**Ready to merge.** Clean, minimal change that improves inheritance graph output by showing virtual functions that have implementations while still hiding pure interface methods. No regressions introduced.

dguido:
## Code review

Found 1 issue:

1. Comment repeats what the code does (CLAUDE.md says: "Code should be self-documenting. No commented-out code (delete it). No comments that repeat what code does.")

https://github.com/crytic/slither/blob/dd00ad3e8e99b155874681aa5e813425572c6506/slither/printers/inheritance/inheritance_graph.py#L124-L126

The comment `# Exclude unoverridden virtual functions` directly restates the boolean logic `not (f.is_virtual and not f.overridden_by)`. The property names `is_virtual` and `overridden_by` already make the intent clear.

PR #2742, opened 2025-06-26
Fix arguments passed to `func` callable in `get_all_storage_variables`

BowTiedDevil:
Running `slither-read-storage` with the `--variable-name` argument results in an exception: `TypeError: 'StateVariable' object is not subscriptable`

Here is an example of reading the WETH balance for the Uniswap V4 PoolManager on mainnet:

```
(slither) btd@dev:~/code/slither_read_storage_fix$ slither-read-storage 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 --variable-name balanceOf --key 0x000000000004444c
5dc75cB358380D2e3dE08A90 --value --rpc-url http://localhost:8545
Traceback (most recent call last):
  File "/home/btd/code/slither_read_storage_fix/.venv/bin/slither-read-storage", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/btd/code/slither_read_storage_fix/slither/tools/read_storage/__main__.py", line 156, in main
    srs.get_all_storage_variables(lambda x: bool(x[1].name == args.variable_name))
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/btd/code/slither_read_storage_fix/slither/tools/read_storage/read_storage.py", line 400, in get_all_storage_variables
    if func(var):
       ~~~~^^^^^
  File "/home/btd/code/slither_read_storage_fix/slither/tools/read_storage/__main__.py", line 156, in <lambda>
    srs.get_all_storage_variables(lambda x: bool(x[1].name == args.variable_name))
                                                 ~^^^
TypeError: 'StateVariable' object is not subscriptable
```

The issue occurs because `func` is called within `read_storage.py` with a single argument `var`, but the lambda passed in the entrypoint expects a tuple:

```
if args.variable_name:
    # Use a lambda func to only return variables that have same name as target.
    # x is a tuple (`Contract`, `StateVariable`).
    srs.get_all_storage_variables(lambda x: bool(x[1].name == args.variable_name))
    srs.get_target_variables(**vars(args))
else:
    srs.get_all_storage_variables()
    srs.get_storage_layout()
```

This PR fixes the call by passing `(contract, var)` as the argument to `func`.

After fix:
```
[...]

PR #2920, opened 2026-01-16
feat: add support for constant and immutable variables in slither-read-storage

### LINKED ISSUE #1614: slither-read-storage: add support for `constant|immutable` variables

It would be nice if `slither-read-storage` was able to output values of `constant` and `immutable` variables. This would save a lot of time by not having to look up these variables on scanners. Please see minimal code and relevant and desired log outputs.

### Version

0.9.1

### Minimal code to reproduce the issue

```solidity
pragma solidity ^0.8.0;

contract TestConstant {
    bytes32 public constant BYTES =
        0x1111111111111111111111111111111111111111111111111111111111111111;
    address public immutable someAddress;

    uint256 public someUint256 = 3;

    constructor() {
        someAddress = 0x2222222222222222222222222222222222222222;
    }
}
```
then `slither-read-storage TestConstant.sol --table`

### Relevant log output

```
+-------------+-------------+------+------+--------+-------+
|     name    | type_string | slot | size | offset | value |
+-------------+-------------+------+------+--------+-------+
| someUint256 |   uint256   |  0   | 256  |   0    |  None |
+-------------+-------------+------+------+--------+-------+
```

### Suggestion for desired output

Variables' values are contracted.

```
+-------------+-------------+--------+------+----------+-------+
|     name    | type_string |  slot  | size |  offset  | value |
+-------------+-------------+--------+------+----------+-------+
| BYTES       |   bytes32   |  n/a   | n/a  |   n/a    |  0x1  |
| someAddress |   address   |  n/a   | n/a  |   n/a    |  0x2  |
| someUint256 |   uint256   |  0     | 256  |   0      |  None |
+-------------+-------------+--------+------+----------+-------+
```

ep0chzer0:
## Summary

This PR adds support for displaying constant and immutable variables in `slither-read-storage`, addressing issue #1614.

### Features
- New `--include-immutable` CLI flag to include constant/immutable variables in output
- Constant values extracted from expressions using `ConstantFolding`
- Immutable values retrieved via RPC getter calls for public variables
- Variables displayed with `(constant)` or `(immutable)` type suffix
- Slot shows `-1` for non-storage variables (they don't have storage slots)

### Example Usage

```bash
# Include immutable and constant variables
slither-read-storage contract.sol 0xAddress --include-immutable --table

# With RPC to get values for public immutables
slither-read-storage contract.sol 0xAddress --include-immutable --table --rpc-url https://eth-mainnet.g.alchemy.com/v2/... --value
```

### Example Output

```
+-------------+----------------------+------+------+--------+-------+
|     name    |     type_string      | slot | size | offset | value |
+-------------+----------------------+------+------+--------+-------+
| BYTES       | bytes32 (constant)   |  -1  | 256  |   0    | 0x1...  |
| someAddress | address (immutable)  |  -1  | 160  |   0    | 0x2...  |
| someUint256 | uint256              |  0   | 256  |   0    |  3    |
+-------------+----------------------+------+------+--------+-------+
```

### Implementation Details

1. Added `_immutable_variables` and `_constant_variables` lists to track these variable types
2. Modified `get_all_storage_variables()` to capture immutable/constant when `include_immutable=True`
3. Added `get_immutable_constant_layout()` method to create SlotInfo for these variables
4. Added `_get_immutable_value()` - calls getter via RPC for public immutables
5. Added `_get_constant_value()` - extracts value from expression using ConstantFolding
6. Modified `get_slot_values()` to skip variables with `slot=-1`

### Test Plan

- [x] Added unit tests for immutable/constant detection
- [x] Added test for disabled-by-default behavior
- [x] Added test for storage layout inclusion
- [x] Added test for constant value extraction
- [x] All tests passing

Fixes #1614

### DIFF
--- a/slither/printers/all_printers.py
+++ b/slither/printers/all_printers.py
@@ -3,6 +3,7 @@ from .summary.contract import ContractSummary
 from .summary.loc import LocPrinter
 from .inheritance.inheritance import PrinterInheritance
 from .inheritance.inheritance_graph import PrinterInheritanceGraph
+from .inheritance.c3_linearization import PrinterC3Linearization
 from .call.call_graph import PrinterCallGraph
 from .functions.authorization import PrinterWrittenVariablesAndAuthorization
 from .summary.slithir import PrinterSlithIR
--- a/slither/printers/inheritance/c3_linearization.py
+++ b/slither/printers/inheritance/c3_linearization.py
@@ -1,0 +1,96 @@
+"""
+Module printing the C3 linearization order for contract inheritance.
+
+The C3 linearization shows the method resolution order (MRO) which determines:
+- Function override order
+- Constructor execution order (reverse of linearization)
+- Super call resolution
+"""
+
+from slither.printers.abstract_printer import AbstractPrinter
+from slither.utils.colors import blue, green, yellow
+
+
+class PrinterC3Linearization(AbstractPrinter):
+    ARGUMENT = "c3-linearization"
+    HELP = "Print the C3 linearization order for each contract"
+
+    WIKI = "https://github.com/crytic/slither/wiki/Printer-documentation#c3-linearization"
+
+    def output(self, filename):
+        """
+        Output the C3 linearization order for all contracts.
+
+        Args:
+            filename(string): not used
+        """
+        info = ""
+        result = {"linearizations": {}}
+
+        _identity = lambda x: x
+
+        for contract in self.contracts:
+            if contract.is_interface:
+                continue
+
+            info += blue(f"\nC3 Linearization for {contract.name}\n")
+            info += "=" * (24 + len(contract.name)) + "\n"
+
+            # The contract.inheritance is already the C3 linearization order
+            linearization = [contract] + list(contract.inheritance)
+
+            result["linearizations"][contract.name] = {
+                "order": [],
+                "constructor_order": [],
+            }
+
+            for i, c in enumerate(linearization):
+                # Determine the type label
+                if i == 0:
+                    label = "[SELF]"
+                    color_fn = green
+                elif c in contract.immediate_inheritance:
+                    label = "[BASE]"
+                    color_fn = yellow
+                else:
+                    label = "[INHERITED]"
+                    color_fn = _identity
+
+                # Get source location
+                source_loc = ""
+                if c.source_mapping and c.source_mapping.filename and c.source_mapping.lines:
+                    source_loc = f" ({c.source_mapping.filename.short}:{c.source_mapping.lines[0]})"
+
+                # Check for constructor
+                has_constructor = "✓" if c.constructor else "✗"
+
+                info += f" {i:2}. {color_fn(f'{label:12} {c.name}')}{source_loc}"
+                info += f"  [constructor: {has_constructor}]\n"
+
+                result["linearizations"][contract.name]["order"].append(
+                    {
+                        "index": i,
+                        "contract": c.name,
+                        "type": label.strip("[]"),
+                        "has_constructor": c.constructor is not None,
+                        "source": c.source_mapping.filename.short
+                        if c.source_mapping and c.source_mapping.filename
+                        else None,
+                    }
+                )
+
+            # Show constructor execution order (reverse of linearization)
+            constructors = [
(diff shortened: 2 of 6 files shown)


ep0chzer0:
Hey, just a heads up - while I was going through the code to implement this feature, I noticed a pre-existing bug in the `--variable-name` flag handling.

In `__main__.py`, the filter lambda was using `x[1].name` expecting a tuple, but `get_all_storage_variables()` passes just a `StateVariable` object to the filter. This would cause issues when someone tries to use `--variable-name`.

I went ahead and fixed that too while I was in there, and added some tests to cover it. Figured it made sense to include it in this PR since I was already touching that code path.

elopez:
Care should be taken if #2742 is merged first, as that PR corrects the behavior of get_all_storage_variables instead

ep0chzer0:
Thanks for the heads up @elopez. I checked PR #2742 - it fixes the same `--variable-name` bug by correcting the return type of `get_all_storage_variables()` to return tuples.

If #2742 merges first, I'll rebase this PR and remove my fix for that bug (since it'll be redundant). The core feature (constant/immutable variable support) is independent of that fix.

I'll keep an eye on #2742's status.

dguido (commented):
## Code Review: feat: add support for constant and immutable variables in slither-read-storage

**Reviewing files:**
- `slither/tools/read_storage/__main__.py`
- `slither/tools/read_storage/read_storage.py`
- `tests/unit/tools/read_storage/test_immutable_constant.py`

---

### Critical (90-100)

No critical issues found.

---

### Important (80-89)

**1. Potential exception when `type_` is `None` - Confidence: 85**

`slither/tools/read_storage/read_storage.py`, lines 220-223 and 253-256

In `get_immutable_constant_layout()`, the code accesses `type_.storage_size` but the guard only checks if `type_` is truthy for the `type_string` assignment, not before accessing `storage_size`:

```python
type_string = str(type_) if type_ else "unknown"
byte_size, _ = type_.storage_size if type_ else (32, 0)
```

This is correct, but the ternary could be clearer. Consider using explicit `None` check for consistency:

```python
byte_size, _ = type_.storage_size if type_ is not None else (32, 0)
```

**2. Silent failure in `_get_immutable_value` on missing `checksum_address` - Confidence: 82**

`slither/tools/read_storage/read_storage.py`, lines 275-292

When `self.rpc_info` is truthy but `self.storage_address` is not set, calling `self.checksum_address` will raise a `ValueError`. The try/except only catches `ValueError` and `TypeError`, but this failure path returns `None` silently without logging. Consider adding a debug log when the exception is caught:

```python
except (ValueError, TypeError) as e:
    logger.warning(f"Could not retrieve immutable {var.name} via RPC: {e}")
```

This is already present, so this is fine. No change needed.

---

### Positive Observations

1. **Good fix for existing bug**: The lambda signature change from `lambda x: bool(x[1].name == args.variable_name)` to `lambda x: bool(x.name == args.variable_name)` correctly fixes an existing bug - the `get_all_storage_variables` method passes `var` (a `StateVariable`) directly to the filter function, not a tuple. The old comment was incorrect.

2. **Comprehensive test coverage**: The tests cover multiple scenarios including:
   - Detection of immutable/constant variables
   - Default disabled behavior
   - Storage layout inclusion
   - Constant value extraction
   - Variable name filtering

3. **Clean separation of concerns**: The new `_get_immutable_value` and `_get_constant_value` methods encapsulate retrieval logic cleanly.

4. **Sensible defaults**: Using `slot=-1` for non-storage variables is a reasonable sentinel value that won't conflict with actual storage slots.

5. **Proper skip logic in `get_slot_values`**: The early return when `slot_info.slot == -1` prevents unnecessary RPC calls for immutable/constant variables.

---

### Minor Suggestions (Not blocking)

1. **Type annotation for `max_depth` parameter**: In `SlitherReadStorage.__init__`, `max_depth` is typed as `int` but the default `rpc_info` parameter is typed as `RpcInfo = None` which should be `RpcInfo | None = None`.

2. **Consider using `typing.TYPE_CHECKING` import guard** for the `Expression` type hint in `find_hardcoded_slot_in_exp` to avoid runtime import overhead.

---

Overall, this is a well-implemented feature that addresses issue #1614. The code is clean, tests are comprehensive, and it correctly fixes a pre-existing bug in the lambda filter signature.

PR #2925, opened 2026-01-19
perf: cache derived_contracts lookup in compilation unit

dguido:
## Summary
- Add cached reverse inheritance map for O(1) derived_contracts lookup

## Problem
The `derived_contracts` property iterated over all contracts in the compilation unit to find children, resulting in O(n) complexity every time it was accessed.

## Solution
- Added `_derived_contracts_map` field to `SlitherCompilationUnit` that maps parent → list of direct children
- Added `derived_contracts_map` property with lazy initialization
- Updated `Contract.derived_contracts` to use the cached map

## Impact
- Complexity: O(n) → O(1) per lookup after first access
- Expected improvement: Significant for inheritance-heavy analyses

## Test Plan

### DIFF
--- a/scripts/benchmark_perf.py
+++ b/scripts/benchmark_perf.py
@@ -1,0 +1,399 @@
+#!/usr/bin/env python3
+"""
+Benchmark Slither performance against real-world smart contract projects.
+
+Usage:
+    # First time: clone all projects and install dependencies
+    python scripts/benchmark_perf.py --setup
+
+    # Run benchmark on current checkout (no branch comparison)
+    python scripts/benchmark_perf.py --no-branch-compare
+
+    # Run on specific projects only
+    python scripts/benchmark_perf.py --projects v4-core,comet --no-branch-compare
+
+    # Compare across branches (default behavior)
+    python scripts/benchmark_perf.py
+
+    # Output JSON for CI/automation
+    python scripts/benchmark_perf.py --no-branch-compare --json
+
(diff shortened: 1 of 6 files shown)


dguido:
## Code review

Found 1 issue:

1. **Semantic change breaks `external_function.py` detector** (bug)

The optimization changes the behavior of `derived_contracts` from returning **all descendants** to returning only **direct children**:

- **Old**: `[c for c in candidates if self in c.inheritance]` - uses full inheritance chain
- **New**: builds map using `contract.immediate_inheritance` - direct parents only

For inheritance `A <- B <- C`:
- Old: `A.derived_contracts` returns `[B, C]`
- New: `A.derived_contracts` returns `[B]` only

This breaks the `external_function.py` detector which explicitly requires all descendants to find function overrides:

https://github.com/crytic/slither/blob/4f5f095ee3fac65d4c8165b45998b4917447bf35/slither/detectors/functions/external_function.py#L113-L121

The comment says: "We assume the provided function is the base-most function, so we check **all derived contracts** for a redefinition." With the new behavior, overrides in grandchild contracts will be missed.

**Suggested fix**: Use `contract.inheritance` instead of `contract.immediate_inheritance` when building the map:

```python
for contract in self.contracts:
    for parent in contract.inheritance:  # NOT immediate_inheritance
        if contract not in derived_map[parent]:
            derived_map[parent].append(contract)
```

---

---

PR #2927, opened 2026-01-19
perf: cache functions_and_modifiers property

dguido:
## Summary
- Memoize the `functions_and_modifiers` property to avoid list concatenation on every access

## Problem
The `functions_and_modifiers` property created a new combined list (`self.functions + self.modifiers`) on every access.

## Solution
- Added `_functions_and_modifiers_cached` field
- Added memoization to `functions_and_modifiers` property

## Impact
- Complexity: O(n) → O(1) for repeated access
- Expected improvement: Small but noticeable in tight loops over contract members

## Test Plan


dguido:
Closing based on benchmark results. This PR showed a -11.6% regression on v4-core. The caching overhead appears to outweigh the benefits for typical projects.

PR #2929, opened 2026-01-19
perf: use derived_contracts for faster implementation lookup

dguido:
## Summary
- Optimize `get_all_implementations()` by using `target.derived_contracts` instead of iterating all contracts

## Problem
The function iterated over all contracts to find implementations of abstract contracts/interfaces, resulting in O(n) complexity.

## Solution
- Replace the loop over all contracts with direct access to `target.derived_contracts`
- Added type annotation for `implementations` variable

## Impact
- Complexity: O(n) → O(k) where k is the number of derived contracts
- Expected improvement: Significant for large codebases with many contracts

## Note
This PR works standalone, but benefits most when combined with the `derived_contracts_map` optimization (PR #X).

## Test Plan

### DIFF
--- a/slither/utils/source_mapping.py
+++ b/slither/utils/source_mapping.py
@@ -47,18 +47,21 @@ def get_implementation(target: SourceMapping) -> Source:
 
 def get_all_implementations(target: SourceMapping, contracts: list[Contract]) -> set[Source]:
     """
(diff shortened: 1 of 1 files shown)


dguido:
Closing based on benchmark results. This PR showed a -9.0% regression on v4-core. The caching overhead appears to outweigh the benefits for typical projects.

PR #2930, opened 2026-01-19
perf: cache base list properties in Contract

dguido:
## Summary
- Memoize the base list properties (`structures`, `enums`, `events`, `functions`, `modifiers`) to avoid repeated `list()` calls

## Problem
Properties like `structures`, `enums`, etc. called `list(self._structures.values())` on every access, creating new list objects repeatedly.

## Solution
Added cache fields and memoization for:
- `structures` (was `list(self._structures.values())`)
- `enums` (was `list(self._enums.values())`)
- `events` (was `list(self._events.values())`)
- `functions` (was `list(self._functions.values())`)
- `modifiers` (was `list(self._modifiers.values())`)

## Impact
- Complexity: O(n) → O(1) for repeated access
- Expected improvement: Small but cumulative across all property accesses

## Test Plan

### DIFF
--- a/slither/core/declarations/contract.py
+++ b/slither/core/declarations/contract.py
@@ -120,9 +120,6 @@ class Contract(SourceMapping):
             dict[StateVariable, set[StateVariable | Function]] | None
         ) = None
 
-        self._structures_list: list[StructureContract] | None = None
-        self._enums_list: list[EnumContract] | None = None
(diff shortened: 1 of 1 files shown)


dguido:
Closing based on benchmark results. This PR showed a -9.9% regression on v4-core. The caching overhead appears to outweigh the benefits for typical projects.

PR #2992, opened 2026-03-31
fix: resolve shadowed modifiers via IdentifierPath

### LINKED ISSUE #2838: [Bug-Candidate]: Slither is unable to locate the Modifier referenced in the IdentifierPath format.
### Describe the issue:

Slither is unable to locate the Modifier referenced in the IdentifierPath format.

```
contract A {
    modifier m() virtual {
        _;
    }
}

contract C is A {
    modifier m() override {
        _;
    }

    function f() public A.m returns (uint) {}
}
```

### Version:

0.11.3

### Relevant log output:

```shell
ERROR:ContractSolcParsing:Missing function Variable not found: A.m (context C a.sol#7-13 (1 - 2))
INFO:Slither:a.sol analyzed (2 contracts with 100 detectors), 0 result(s) found
```

cats2101:
## Summary

Fixes #2838

When a modifier is referenced via `IdentifierPath` (e.g. `A.m` in `function f() public A.m`), the compiler resolves the exact declaration id via `referencedDeclaration`. However, `_find_variable_from_ref_declaration` was filtering out matches where `is_shadowed` is true — which incorrectly rejected parent modifiers that are overridden by a child contract.

Since `referencedDeclaration` is the compiler's authoritative resolution, we should trust it without filtering by `is_shadowed`. The shadowed filter is appropriate for name-based lookups (where you want the most-derived version), but not when looking up by compiler-resolved id.

### Before

```
ERROR:ContractSolcParsing:Missing function Variable not found: A.m (context C)
```

### After

Parses correctly, no error.

## Reproducer

```solidity
contract A {
    modifier m() virtual { _; }
}

contract C is A {
    modifier m() override { _; }
    function f() public A.m returns (uint) { return 1; }
}
```

## Test plan

- Verified fix resolves the exact reproducer from #2838
- All 6593 AST parsing tests pass
- All 150 modifier-related parsing tests pass

### DIFF
--- a/slither/solc_parsing/expressions/find_variable.py
+++ b/slither/solc_parsing/expressions/find_variable.py
@@ -55,6 +55,7 @@ def _find_variable_from_ref_declaration(
     all_functions: list["Function"],
     function_parser: Optional["FunctionSolc"],
     contract_declarer: Optional["Contract"],
+    is_identifier_path: bool = False,
 ) -> Contract | Function | None:
     """
     Reference declarations take the highest priority, but they are not available for legacy AST.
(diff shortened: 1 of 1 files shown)


cats2101:

PR #3003, opened 2026-04-12
Fix: filter is_constructor_variables in printers

### LINKED ISSUE #865: Filter is_constructor_variables on printers
Some printers still output info if [`is_constructor_variables`](https://github.com/crytic/slither/blob/e31487455c5cc07de776cf300da4d69f9a0fac5d/slither/core/declarations/function.py#L355-L363) is true.

We should review every printer to see if it makes sense to print them

cats2101:
Closes #865

Printers that iterate over `contract.functions` were including synthetic constructor-variable functions (`slitherConstructorVariables`, `slitherConstructorConstantVariables`) in their output. These are internal bookkeeping functions and shouldn't be displayed.

This adds `is_constructor_variables` filtering to the 9 printers that were missing it, matching the pattern already used in `contract-summary`, `function-id`, and `inheritance-graph`.

**Printers fixed:**
- `cfg` — CFG export
- `dominator` — dominator tree export
- `vars-and-auth` — state variables and authorization
- `modifiers` — modifier calls
- `require` — require/assert calls
- `evm` — EVM instructions
- `slithir` — SlithIR representation
- `slithir-ssa` — SlithIR SSA representation
- `call-graph` — call graph export

### DIFF
--- a/slither/printers/call/call_graph.py
+++ b/slither/printers/call/call_graph.py
@@ -332,7 +332,9 @@ class PrinterCallGraph(AbstractPrinter):
             ]
             all_functions = [item for sublist in all_functionss for item in sublist]
             all_functions_as_dict = {
-                function.canonical_name: function for function in all_functions
+                function.canonical_name: function
+                for function in all_functions
+                if not function.is_constructor_variables
             }
             content = "\n".join(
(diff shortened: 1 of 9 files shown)


cats2101:

PR #3011, opened 2026-04-20
feat: accept slither.conf.json as alternative default config

### LINKED ISSUE #595: Support slither.conf.json and slither.config.json
Right now Slither will look for the config file named `slither.config.json`. We could also look for `slither.conf.json` for convenience

cats2101:
Closes #595

## Summary

When no `--config-file` is provided, Slither will now look for `slither.conf.json` as a fallback if `slither.config.json` is not present. If both files exist, `slither.config.json` takes precedence (preserves current behavior).

This matches the request in the issue: accept `slither.conf.json` for convenience without breaking anyone.

## Changes

- `slither/utils/command_line.py`: introduce `DEFAULT_CONFIG_FILENAMES = ("slither.config.json", "slither.conf.json")` and update `read_config_file` to iterate the tuple.
- `slither/__main__.py` and `slither/tools/slither_format/__main__.py`: update `--config-file` help text. For `slither-format`, the default changes from the hardcoded string `"slither.config.json"` to `None` so the auto-detection chain applies consistently across both entrypoints.
- `docs/src/Usage.md`: mention both filenames.
- `tests/unit/utils/test_command_line.py`: new unit tests covering both filenames, precedence, and the no-default case.

## Test plan

- [x] New unit tests pass (`pytest tests/unit/utils/test_command_line.py -v`)
- [x] `ruff check` passes on modified files
- [x] Behavior check: with only `slither.conf.json` present, it is picked up and its flags are applied
- [x] Behavior check: with both present, `slither.config.json` still wins


cats2101:

PR #3067, opened 2026-07-31
fix(vyper): correct shift and division assignment operator mappings

racequite:
## Description

Fix incorrect Vyper AST operator mappings for shift operations and division assignment.

Previously, Vyper `Shr` and `Shl` were mapped to the opposite binary operators, so `>>` could be parsed as `<<` and vice versa. The augmented assignment mapping had the same issue for `>>=` and `<<=`, and also mapped `Div` to `-=` instead of `/=`.

These operator strings are passed into `BinaryOperationType.get_type()` and `AssignmentOperationType.get_type()`, so the incorrect mapping could affect SlithIR generation and analyses that depend on arithmetic or bitwise semantics.

## Changes

- Map `Shr` to `>>` and `Shl` to `<<`
- Map `Div` augmented assignment to `/=`
- Map `Shr` to `>>=` and `Shl` to `<<=`
- Add unit tests covering the affected Vyper operator mappings

## Testing

```text
python -m pytest tests/unit/vyper_parsing/test_ast.py -q
```

image


racequite:
@smonicas Hi, gentle ping on this Vyper parsing fix.

It corrects the reversed Shl/Shr mappings for binary and augmented-assignment operations, as well as the incorrect Div augmented-assignment mapping. Focused AST unit tests are included for all affected operators.

Would you mind taking a look when you have a chance? Thanks.

PR #3069, opened 2026-08-04
Add RNM mutator (Remove Negation)

sidarth16:
This PR introduces the **RNM (Remove Negation)** mutator to the existing mutator set.

The mutator removes logical negation (`!`) from unary expressions, enabling mutation testing to evaluate whether tests correctly detect inverted boolean logic.

### Examples

`require(!paused)` → `require(paused)`

`assert(!(x > 0))` → `assert((x > 0))`


slither/tools/mutator/mutators/RNM.py
```diff
@@ -0,0 +1,58 @@
...
+
+class RemoveNegationVisitor(ExpressionVisitor):
+    def __init__(self, expression, mutator, result):
+        self._mutator = mutator
+        self._result = result
+
+        # Traverse the expression AST
+        super().__init__(expression)
+
+    def _post_unary_operation(self, expression: UnaryOperation):
```
bohendo: Let's annotate these param types and return values

slither/tools/mutator/mutators/RNM.py
```diff
@@ -0,0 +1,58 @@
...
+
+            for node in function.nodes:
+                if not self.should_mutate_node(node):
+                    continue
+
+                expression = getattr(node, "expression", None)
+                if expression is None:
+                    continue
+
+                RemoveNegationVisitor(expression, self, result)
```
bohendo: Note that for a ternary expression like `!(a ? b : c)`, slither breaks this down into an `IF a` node and two return nodes: `!b` and `!c,` and we're hitting `RemoveNegationVisitor` once for each of the return nodes. Result is two mutations:
- `!(a ? b : c)` -> `b`
- `!(a ? b : c)` -> `c`

We should prob simplify this and just trim the `!` with `source[1:]` after checking `source.startswith("!")` instead of replacing `!<expression>` with `<expression>` bc we're tripping over what's actually in the <expression> here. This will still generate two mutations but they'll be the same & get deduplicated, which is fine.

tests/tools/mutator/test_mutator.py
```diff
@@ -283,3 +284,37 @@ def test_should_mutate_function_includes_modifier(solc_binary_path):
...
+            target_modifiers=None,
+        )
+
+        patches = mutator._mutate()
+
+        assert "patches" in patches
+        assert file_path in patches["patches"]
+
+        assert any(
+            patch["old_string"] == "!(newNumber != 7)" and patch["new_string"] == "(newNumber != 7)"
```
bohendo: Let's add a ternary expression test case here

bohendo (changes requested):
Good work. But ternary expressions are tricky!

PR #3085, opened 2026-08-19
fix(vyper): model loop bounds and range arguments correctly

zloglevel:
## Summary

Fix Vyper loop modeling in the CFG and SlithIR generation.

## Root Cause

Vyper loops were modeled with an inclusive upper bound:

```text
counter_var <= len(iterable)
counter_var <= range_end
```

This caused one extra modeled iteration and could produce an invalid access at `iterable[len(iterable)]`.

For two-argument `range(start, end)`, Slither also initialized the artificial counter to `0` and used the first argument as the endpoint. This did not match Vyper's half-open range semantics.

## Changes

- Use `counter_var < len(iterable)` for array iteration.
- Use `counter_var < end` for `range(end)`.
- Initialize the artificial counter with `start` for `range(start, end)`.
- Use the second positional argument as the endpoint.
- Preserve Vyper's `bound` keyword behavior without treating it as the range endpoint.
- Add regression coverage for dynamic array iteration and two-argument ranges.
- Update affected Vyper CFG/SlithIR snapshots.

## Testing

```text
30 passed
ruff check passed
git diff --check passed
```

### DIFF
--- a/slither/vyper_parsing/declarations/function.py
+++ b/slither/vyper_parsing/declarations/function.py
@@ -300,6 +300,21 @@ class FunctionVyper:
 
                 link_underlying_nodes(curr_node, node_startLoop)
 
+                counter_initial = Int("-1:-1:-1", -1, 0)
+                range_end = None
+                if isinstance(expr.iter, Call):
+                    # Vyper's optional ``bound`` keyword limits compilation work;
+                    # it does not change the half-open range endpoint.
+                    if len(expr.iter.args) == 1:
+                        range_end = expr.iter.args[0]
+                    elif len(expr.iter.args) == 2:
+                        counter_initial = expr.iter.args[0]
+                        range_end = expr.iter.args[1]
+                    else:
+                        raise ParsingError(
+                            "Vyper range loops must have one or two positional arguments"
+                        )
+
                 local_var = LocalVariable()
                 local_var.set_function(self._function)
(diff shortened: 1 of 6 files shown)


glarregay-tob:
Two test cases missing before this lands:

- `range(x, bound=N)` — `bound=` doesn't appear anywhere under `tests/`. The new branch depends on `bound` arriving in `Call.keywords` rather than `args`; nothing pins that.
- `range(x, x + N)` — the added case only covers the literal `range(5, 7)`. The non-literal form is the one master models as `counter_var = 0` / `counter_var <= x`.

Both fit the existing `slither_from_vyper_source` fixture.

Worth pinning as a third: `for i in range(-3, 2)` is valid on 0.3.10 — vyper emits `[repeat, range_ix0, -3, 5, 5, ...]`. Master models it as `counter_var <= -3` against a `uint256`-annotated counter, so `type-based-tautology` fires on correct source. This branch drops that finding.

zloglevel:
@glarregay-tob  @smonicas  Could you please take a look when you have a chance? Thanks!

zloglevel:
@glarregay-tob Thanks for the review. I've added the cases you called out, including the bound= form, runtime/non-literal range bounds, and negative range starts, along with the corresponding expected results.

Would you mind taking another look when convenient? Thanks!

Issue #1775, opened 2023-03-17
[Bug-Candidate]: Missing data dependencies involving assembly with solc <0.6.0 


webthethird:
### Describe the issue:

With solc versions < 0.6.0, assembly is not parsed very well in Slither, which has caused me trouble before. All we get is a single assembly node with `node.inline_asm` containing all of the assembly code as one long string.

My problem right now is that `slither.analyses.data_dependency.get_dependencies()` is missing several dependencies when the data flow involves some assembly and solc is < 0.6.0. In the code example provided, I am trying to trace the `implementation` parameter in `_delegate(address implementation)` back to its source, `IMPLEMENTATION_SLOT`. The data dependency should be as follows:

- `implementation` gets its value from a temporary variable being passed into a call to `_delegate` in the function `_fallback`.
- The temporary variable is the return value from a call to `_implementation()`, i.e. `_delegate(_implementation())`.
- The return variable is declared in the function signature of `_implementation` as `address impl`.
- `impl` is assigned a value _in assembly_: `impl := sloat(slot)`
- `slot` is assigned the value of `IMPLEMENTATION_SLOT` before the assembly.

When I do call `get_dependencies(implementation, contract)`, the only dependency currently being returned is the temporary variable.

The new `test_upgradeability_util.py` included in #1757 fails for this reason [at line 73](https://github.com/crytic/slither/blob/bff30a3481f61babbd55323b3ab6a42eb3be6118/tests/test_upgradeability_util.py#L73).

### Code example to reproduce the issue:

```solidity
pragma solidity ^0.5.0;

contract ZosProxy {
  function () payable external {
    _fallback();
  }

  function _implementation() internal view returns (address);

  function _delegate(address implementation) internal {
    assembly {
      calldatacopy(0, 0, calldatasize)
      let result := delegatecall(gas, implementation, 0, calldatasize, 0, 0)
      returndatacopy(0, 0, returndatasize)
      switch result
      case 0 { revert(0, returndatasize) }
      default { return(0, returndatasize) }
    }
  }

  function _willFallback() internal {
  }

  function _fallback() internal {
    _willFallback();
    _delegate(_implementation());
  }
}

library AddressUtils {
  function isContract(address addr) internal view returns (bool) {
    uint256 size;
    assembly { size := extcodesize(addr) }
    return size > 0;
  }
}

contract UpgradeabilityProxy is ZosProxy {
  event Upgraded(address indexed implementation);

[...]

PR #3087, opened 2026-08-23
fix: detect assembly delegatecall and callcode in unprotected-upgrade

BhariGowda:
## Summary

`unprotected-upgrade` reports an upgradeable logic contract whose `initialize` is callable by anyone and which contains a function that can destroy it. The destruction check in `_can_be_destroyed` had two independent gaps, so a contract with an attacker-reachable `delegatecall` or `callcode` went unreported even when everything else about the finding held.

Both gaps are in the same function and neither was covered by a fixture, so they are fixed together.

## Root cause 1: the low-level call name list never matched `callcode`

`slither/detectors/statements/unprotected_upgradeable.py:16` on master:

```python
if ir.function_name in ["delegatecall", "codecall"]:
```

`codecall` is not an EVM opcode and is not a name SlithIR ever produces. The rest of the codebase spells it `callcode`:

- `slither/slithir/convert.py:1274` and `:1308` build the `LowLevelCall` with `function_name == "callcode"`
- `slither/detectors/attributes/locked_ether.py:73`
- `slither/detectors/statements/controlled_delegatecall.py:17`
- `slither/solc_parsing/yul/parse_yul.py:46`

So that half of the condition was dead. A pre-0.5 contract calling `target.callcode(data)` was never flagged, even though `callcode` runs the callee's code against this contract's storage the same way `delegatecall` does, and a `selfdestruct` in the callee destroys the caller either way. `callcode` was removed in solc 0.5.0, so this half only affects 0.4.x code, but the branch was clearly meant to cover it.

Reproduced with two contracts differing only in the call, compiled with 0.4.25: the `delegatecall` one is reported on master, the `callcode` one is not. Both produce a `LowLevelCall`, with `function_name` `delegatecall` and `callcode` respectively.

## Root cause 2: assembly calls are `SolidityCall`, not `LowLevelCall`

Inline assembly is parsed into Yul nodes, and `call` / `callcode` / `delegatecall` / `staticcall` become `SolidityCall` operations (`slither/solc_parsing/yul/parse_yul.py:46`), not `LowLevelCall`. `_can_be_destroyed` walked `all_low_level_calls()` for delegatecall and `all_solidity_calls()` only for `suicide` / `selfdestruct`, so an assembly `delegatecall` matched neither loop.

The practical consequence is that an upgradeable logic contract with an executor function like this was entirely invisible to the detector:

```solidity
function execute(address target) external {
    require(msg.sender == owner);
    assembly {
        let success := delegatecall(gas(), target, 0, 0, 0, 0)
        if iszero(success) { revert(0, 0) }
    }
}
```

Anyone can call `initialize()` on the logic contract, become `owner`, then point `execute` at a contract that self-destructs, bricking every proxy delegating to it. That is the same scenario the detector's own `WIKI_EXPLOIT_SCENARIO` describes, just written in assembly.

Verified against the IR: for the assembly version `all_low_level_calls()` is empty and `all_solidity_calls()` contains `delegatecall(uint256,uint256,uint256,uint256,uint256,uint256)`, or `callcode(uint256,uint256,uint256,uint256,uint256,uint256,uint256)` for the callcode form.

`slither/detectors/attributes/locked_ether.py:71-86` already handles exactly this split, checking `LowLevelCall.function_name` and `SolidityCall` against the Yul signatures side by side. This PR applies the same pattern. `slither/core/declarations/contract.py:1425` works around the same gap in `is_upgradeable_proxy` with a raw substring match on `node.inline_asm`, which is a separate matter and is left alone here.

## Changes

`slither/detectors/statements/unprotected_upgradeable.py`: pull the two match sets out as module constants, fix `codecall` to `callcode`, and add the two Yul signatures to the solidity-call set alongside `suicide` / `selfdestruct`. No change to the traversal or to the reporting.

## Fixtures, and a limitation worth stating

- `tests/e2e/detectors/test_data/unprotected-upgrade/0.4.25/NativeCallcode.sol` covers root cause 1, using `target.callcode(data)`.
- `tests/e2e/detectors/test_data/unprotected-upgrade/{0.6.11,0.7.6,0.8.15}/AssemblyDestruct.sol` covers root cause 2, with one contract per assembly form so both signatures are exercised.

The assembly fixtures deliberately start at 0.6.11. With solc below 0.6.0 the whole assembly block arrives as a single node holding the source as one opaque string in `node.inline_asm`, so there are no Yul nodes and no `SolidityCall` for the detector to match. I first wrote the assembly fixture for all five versions and the 0.4.25 and 0.5.16 snapshots came back empty, which would have pinned a false negative as expected behaviour, so those two were dropped rather than committed. That parsing gap is tracked in #1775 and is out of scope here.

## Test plan

Run with `uv` per CONTRIBUTING.md.

- Baseline `pytest tests/e2e/detectors/test_detectors.py -k UnprotectedUpgradeable` on master: **25 passed**.
- After the change: **29 passed**. Four new cases, one per new fixture.
- Reverted the detector file and kept the new fixtures: **4 failed, 25 passed**. The four failures are exactly the four new cases, so the fixtures do catch the regression rather than passing vacuously.
- Full `pytest tests/e2e/detectors/test_detectors.py`: **384 passed** on master, **388 passed** with the change. No existing snapshot was modified, only added.
- `ruff check slither tests`: clean. `ruff format --check` on both touched Python files: already formatted.
- Full `pytest tests` on both master and this branch produces an identical set of 6 failures and 10 errors, all of them vyper cases with no vyper compiler installed plus the two `read-storage` tests that need a node. Nothing in that set moves with this change.

Sibling PR: #3088 fixes the same `SolidityCall` versus `LowLevelCall` gap in `controlled-delegatecall`. The two are independent and can be reviewed in either order.


glarregay-tob:
One gap before this lands: assembly `selfdestruct` still isn't matched. Yul builtins register as `name(uint256,...)` (`evm_functions.py:225,253`), so it arrives as `selfdestruct(uint256)` and the list only has the address spellings. Tested it: unprotected `initialize()` plus `assembly { selfdestruct(caller()) }` isn't reported; adding `SolidityFunction("selfdestruct(uint256)")` reports it.

Could you add it with a fixture in `AssemblyDestruct.sol`?

Rest looks good.

BhariGowda:
Good catch, thanks. Pushed as a follow-up commit rather than an amend so the change against your review is easy to see.

`SolidityFunction("selfdestruct(uint256)")` added to `_DESTRUCTIVE_SOLIDITY_CALLS`. I confirmed the signature off the IR rather than just from `evm_functions.py`: dumping `all_solidity_calls()` over the fixture gives `SolidityFunction('selfdestruct(uint256)')`. I also expanded the comment above the list to say why the assembly spellings differ, since the old wording only mentioned delegatecall/callcode.

`suicide` is not a Yul builtin, so there is no assembly counterpart to add for that name. Happy to add it defensively if you would rather have it.

Fixture: `AssemblySelfdestruct` appended to `AssemblyDestruct.sol` on 0.6.11/0.7.6/0.8.15, keeping the one-contract-per-destructive-form layout already there. Zip artifacts and snapshots regenerated.

Verification:

- Your exact case reproduces before the fix: unprotected `initialize()` plus `assembly { selfdestruct(caller()) }` gives 0 findings, and 1 after.
- Reverted only the new signature line and re-ran: exactly the 3 `AssemblyDestruct` cases fail and 26 pass, and the only snapshot difference is the missing `AssemblySelfdestruct` line. Reapplied and all 29 pass. So the fixture pins the behaviour rather than just riding along.
- Full detector suite: 388 passed before, 388 passed after. No snapshot outside those 3 moved.
- `ruff check` and `ruff format --check` clean.

